PR #10, opened 2016-05-30
task(dsl): Replace colons with dashes

jaagr:
Colons break ini syntax highlighters and linters and there isn't really a good reason for using them. Just an unfortunate choice that would be good to fix early on.

How to deal with backward compatibility?


jaagr:
9efaa0e7e904bd582481e9c2191b3305d378cc42 makes it possible to patch old configs.

Do it automagically or notice user?

jaagr:
2c74cb06de7e7605492dbb4d1981b95476587140

PR #26, opened 2016-06-23
Add man pages

ammongit:
In response to #11, I created some man pages for lemonbuddy. While these are by no means perfect or complete, it should be a good starting place. `lemonbuddy_config.5` is especially is far from complete, but it is probably best to not invest too much time into it when the modules it describes are still in early stages of development.

The `Makefile` I added to the man page directory serves three purposes:
* gzip the man pages
* Create a PDF with all the man pages. (Ghostscript and groff must be installed)
* Update the version in the man page header. (This must be explicitly called via `make update-versions`)


jaagr:
Hi.. Thanks alot for contributing! I will get back to you as soon as I get some time this weekend.

jaagr:
Hi @ammongit,

Thanks again for contributing! Been meaning to get some man pages put together for quite some time now but it's been pushed aside by other stuff, so thanks for creating some initial pages.
I'll merge this and try to combine it with the work I've done locally regarding man pages.

PR #38, opened 2016-08-01
[bspwm] Separate focused/unfocused state from occupation state

delehef:
Hello,

I like to keep the same icons whether a workspace is focused or not, but makes its color change; hence the following commit removing the `WORKSPACE_FOCUSED` flag and replacing it with `WORKSPACE_FOCUSED_{URGENT,EMPTY,OCCUPIED}`.

Do you like the idea?

### DIFF
--- a/README.md
+++ b/README.md
@@ -674,11 +674,17 @@ See [the bspwm module](#module-internalbspwm) for details on `label-dimmed`.
   ;   %icon%
   ;   %index%
   ; Default: %icon%  %name%
+  ; Focused workspaces may either all use the same properties...
   label-focused = %icon%
   label-focused-foreground = #ffffff
   label-focused-background = #3f3f3f
   label-focused-underline = #fba922
   label-focused-padding = 4
+  ; ... or you may change the properties depending of their state
+  label-focus-occupied = O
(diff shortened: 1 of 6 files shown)


jaagr:
Hi and thanks for contributing. Yeah I think it allows for more specific configuration but I think we should add a `label-focused` parameter that the user could define to avoid having to set all 3 if they are all the same.

Could you also update the README to reflect the changes?

delehef:
Indeed, that's a nice idea, I'll try to add it this evening

Cheers!

delehef:
Well, the “evening” took a little bit more than I'd have liked it to, but now it should be OK.

delehef:
Sorry to insist, bu t are you still interested in this PR?

jaagr:
Yeah sorry for the delay... stuff keeps popping up. Will have a look once I'm settled with my current list.

delehef:
No problem, I just wanted to know if you were still interested.

jaagr:
Would you mind applying the changes to the new module? Can't merge since the file structure have changed.

delehef:
Sent in a new PR.

jaagr:
#201 has been merged

PR #41, opened 2016-08-14
Fix crash in custom/menu

### LINKED ISSUE #40: Crash with custom/menu
When adding a custom/menu module and pressing the opening label, nothing happens and after a few more clicks the bar crashes with the following message:
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::substr: __pos (which is 18446744073709551615) > this->size() (which is 5)
/usr/bin/lemonbuddy_wrapper: line 41: 30712 Aborted                 (core dumped) $lemonbuddy "$@" -p "$pipe"
```
My module looks like this (but I don't think it matters):
```
[module/menu-apps]
type = custom/menu
label-open = Apps
label-close = x
menu-0-0 = Firefox
menu-0-0-exec = firefox &
menu-0-0-foreground = #fba922
menu-0-1 = Chromium
menu-0-1-exec = chromium &
menu-0-1-foreground = #fba922
```
I have managed to track down the offending code and the program crashes on the following line
https://github.com/jaagr/lemonbuddy/blob/master/src/services/builder.cpp#L398
which is called by
https://github.com/jaagr/lemonbuddy/blob/master/src/modules/menu.cpp#L79

The problem is the `val.substring(val.size() - 6)` part, since `val` has at that point a length of 5 and looks like this `#C8C6`, since on line [393](https://github.com/jaagr/lemonbuddy/blob/master/src/services/builder.cpp#L393) you shorten `val`.

I will try to fix this myself, but I'm having problems understanding your logic behind the `color_alpha` method, so you're probably a lot faster.

patrick96:
Wasn't that bad after all.
I also did some refactoring in the color_alpha class
Fixes #40

### DIFF
--- a/src/services/builder.cpp
+++ b/src/services/builder.cpp
@@ -394,8 +394,7 @@ void Builder::color_alpha(std::string alpha_)
   }
 
   if (alpha.size() == 4) {
-    // If alpha looks like #123 then we treat as a color and ignore the fg color
-    this->color(alpha + alpha.substr(1));
+    this->color(alpha);
     return;
   }
 
--- a/src/services/builder.cpp
+++ b/src/services/builder.cpp
@@ -389,22 +389,21 @@ void Builder::color_alpha(std::string alpha_)
 {
   auto alpha(alpha_);
   std::string val = this->get_opts()->foreground;
-  if (alpha.find("#") == 0) {
-    if (alpha.size() == 3)
-      this->color(alpha.substr(0, 3) + val.substr(val.size() - 6));
-    else if (alpha.size() == 4)
-      this->color(alpha + alpha.substr(1));
+  if (alpha.find("#") == std::string::npos ) {
+    alpha = "#" + alpha;
+  }
+
+  if (alpha.size() == 4) {
+    // If alpha looks like #123 then we treat as a color and ignore the fg color
+    this->color(alpha + alpha.substr(1));
     return;
   }
+
   if (val.size() < 6 && val.size() > 2) {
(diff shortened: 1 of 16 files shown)


src/services/builder.cpp
```diff
@@ -380,22 +380,21 @@ void Builder::color_alpha(std::string alpha_)
...
-      this->color(alpha.substr(0, 3) + val.substr(val.size() - 6));
-    else if (alpha.size() == 4)
-      this->color(alpha + alpha.substr(1));
+  if (alpha.find("#") == std::string::npos ) {
+    alpha = "#" + alpha;
+  }
+
+  if (alpha.size() == 4) {
+    // If alpha looks like #123 then we treat as a color and ignore the fg color
+    this->color(alpha + alpha.substr(1));
```
jaagr: This looks like a bug (I know it wasn't added by you).

`#123 != #123123` but `#112233`

`this->color(alpha);` would be correct.

patrick96: Will update this shortly.

jaagr:
Please rebase onto upstream master

patrick96:
Done

jaagr:
run this:

```sh
git reset HEAD~1
git remote add upstream https://github.com/jaagr/lemonbuddy
git fetch upstream
git rebase upstream/master
```

patrick96:
Duh, I forget to `git fetch` before rebasing. :smile:

jaagr:
Merged 365a9cc

PR #43, opened 2016-08-16
Make interfaces properly detect if they are connected

patrick96:
The old approach would make interfaces that are 'up' but not connected say 
they are connected (and in case of wireless print 'No Network' as essid)


patrick96:
Sorry about that. This still had a commit from another pull request

PR #55, opened 2016-08-31
more robust way to extract relevant data

### LINKED ISSUE #54: bspwm module on multiple monitors
I am using the bspwm module on a triple monitor setup. The second Monitor renders unclickable labels for the third monitor. Let me demonstrate this on a simple example:

Configuring bars for each monitor, only displaying a single bspwm module:

    [settings]
    [bar/left]
    monitor = DP-0
    modules-left = bspwm

    [bar/middle]
    monitor = DP-2
    modules-left = bspwm

    [bar/right]
    monitor = DP-4
    modules-left = bspwm

    [module/bspwm]
    type = internal/bspwm
    label-active = %name%
    label-occupied = %name%
    label-urgent = %name%
    label-empty = %name%

bspwm monitors are:

    bspc monitor DP-0 -d A B
    bspc monitor DP-2 -d C D
    bspc monitor DP-4 -d E F

running all bars:

    lemonbuddy_wrapper left &
    lemonbuddy_wrapper middle &
    lemonbuddy_wrapper right &

DP-0 (left) shows `ABCD` when DP-4 has focus, `AB` when DP-0 has focus
DP-2 (middle) shows `CDEF` no matter which monitor has the focus
DP-4 (right) shows `EF` no matter which monitor has the focus

On DP-2 both B and C are clickable, D and E are not doing anything, but they update whenever D or E is clicked on DP-4.

Do you have any idea why it is behaving like that? I assume its a bug in the bspwm module, but i haven't looked into the source yet.

Would be great if we could solve this together!

PhilippSchaffrath:
fixes bugged extraction of monitor data, fixes #54

### DIFF
--- a/src/modules/bspwm.cpp
+++ b/src/modules/bspwm.cpp
@@ -2,6 +2,7 @@
 #include <vector>
 #include <sstream>
 #include <sys/socket.h>
+#include <regex>
 
 #include "config.hpp"
 #include "lemonbuddy.hpp"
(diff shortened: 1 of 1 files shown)


jaagr:
Nice pattern. Unfortunately it's really slow.

```
1000000 passes
execution speed: 647ms
execution speed: 171217ms
```

PhilippSchaffrath:
You should only "optimize performance" if you have a performance problem. In this case, this pattern will be executed n times per window focus change, where n is the amount of bars displaying the bspwm module. I really don't see this as critical.

PR #57, opened 2016-08-31
fix for "invalid geometry specified"

### LINKED ISSUE #56: "Invalid geometry specified" if locale is set
The lemonbar does not render and prints the error "Invalid geometry specified" once a locale has been defined. It seems like it doesn't matter what locale you set. The error will be printed out as soon as a locale is present in the configuration.

```
[bar/something]
locale = de_DE.UTF-8
```

PhilippSchaffrath:
setting the locale through `std::setlocale` instead of `std::locale::global` gets rid of the error, and applies the locale correctly

fixes #56

### DIFF
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -48,7 +48,7 @@ Bar::Bar() : config_path(config::get_bar_path()), opts(std::make_shared<Options>
 
   try {
     this->opts->locale = config::get<std::string>(this->config_path, "locale");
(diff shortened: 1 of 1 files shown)


PhilippSchaffrath:
I don't know why, but using setlocale instead of locale::global fixes the problem.

jaagr:
Closing this.. lemonbar has been dropped in 2.0

PR #60, opened 2016-08-31
Remove python executable replacment in PKGBUILD

patrick96:
Only for the '-git' package since the other one would also require a
version bump


jaagr:
The `pkgrel` would need to be changed in `contrib/lemonbuddy-git.aur/.SRCINFO` as well

patrick96:
Done

PR #61, opened 2016-09-01
Fix broken travis build

patrick96:
Somehow ubuntu on the travis server does not have
/usr/lib/python2.7/dist-packages in the path and thus the xcbgen module
cannot be found. Adding it to the PYTHONPATH fixes this


jaagr:
nice

PR #92, opened 2016-10-14
feat(i3): Scrollable workspaces and strip-wsnumber option

MoKev:
Hello,

I like to be able to scroll through my workspaces, so I added scroll up and down events which send the command `workspace next` or `workspace prev`.

I also added the option `strip-wsnumbers` which is a feature of i3bar. If set to true the workspace label is split by `:` and only the second part is shown. This makes it possible to use numbered workspaces but only show the name.


jaagr:
Hi! Looks good. I only wonder if it would be better to use `prev_on_output` and `next_on_output` to better handle multi-monitor setups?

MoKev:
Yes, this is better :)

jaagr:
Awesome :-) Merged.

PR #147, opened 2016-11-05
feat(volume): Added volume mapping

NBonaparte:
I'm used to the volume mapping used by `amixer -M` and `alsamixer`, so I thought it would be useful to have an option which displays the volume this way instead of the raw percentage.


jaagr:
Looks good... merged

PR #156, opened 2016-11-11
feat(temperature): Added temp module and examples

NBonaparte:
Adds temperature module. A specific thermal zone `thermal-zone` can be set, and a threshold temperature `temperature-warn` can be specified for displaying a separate 'warning' label `label-warn`.


jaagr:
Nice.. could you also add a page to the wiki on how to configure the module?

NBonaparte:
Unfortunately, I don't have write access to the wiki.

jaagr:
There, now the wiki should be writable by everyone now..

NBonaparte:
Thanks! I've added a page for the temperature module and also a volume mapping example from dce81d4266d970e36d0b1c382bcbf28b3e1c865a.

jaagr:
Great! I've added support for `<ramp>` to the module, to allow icons based on temp value. Will add it to the wiki as well.

PR #163, opened 2016-11-14
feat: add padding for values

NBonaparte:
Added `string_util::format_minwidth` which allows for padding, expanding upon the `udspeed-minwidth` setting. The values which are most susceptible to shifting (volume, battery, cpu percentage, etc.) now use this function, but it might be beneficial to add it to all numerical values.

### DIFF
--- a/src/drawtypes/label.cpp
+++ b/src/drawtypes/label.cpp
@@ -26,7 +26,7 @@ namespace drawtypes {
   }
 
   void label::replace_token(string token, string replacement) {
-    if (m_text.find(token) == string::npos)
+    if (m_text.find(token) == string::npos || m_token_iterator == m_token_bounds.end())
       return;
 
     m_tokenized = string_util::replace_all_bounded(m_tokenized, token, replacement,
@@ -77,12 +77,13 @@ namespace drawtypes {
     name = string_util::ltrim(string_util::rtrim(name, '>'), '<');
 
     string text;
-    string line{text};
 
     if (required)
       text = conf.get<string>(section, name);
     else
       text = conf.get<string>(section, name, def);
+
+    string line{text};
 
     while ((start = line.find('%')) != string::npos && (end = line.find('%', start + 1)) != string::npos) {
       auto token = line.substr(start, end);
--- a/src/drawtypes/label.cpp
+++ b/src/drawtypes/label.cpp
@@ -72,37 +72,55 @@ namespace drawtypes {
    */
   label_t load_label(const config& conf, string section, string name, bool required, string def) {
     vector<struct bounds> bound;
+    size_t start, end, pos;
 
     name = string_util::ltrim(string_util::rtrim(name, '>'), '<');
 
     string text;
+    string line{text};
 
     if (required)
       text = conf.get<string>(section, name);
     else
       text = conf.get<string>(section, name, def);
 
-    vector<string> text_split = string_util::split(text, '%');
+    while ((start = line.find('%')) != string::npos && (end = line.find('%', start + 1)) != string::npos) {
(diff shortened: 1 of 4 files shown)


jaagr:
I think it would be better if we incorporated min/max width in the tokens to avoid new parameters. For example:

~~~
    min: %token:3%
    max: %token:0:5%
min+max: %token:2:5%
~~~

Then we could let the label class read the values the first time they are being replaced.

NBonaparte:
That's a good idea; I'll try working on it later.

NBonaparte:
Hey @jaagr, I finished implementing the feature as you suggested. Could you check it out?

jaagr:
Nice job, will have a look

src/drawtypes/label.cpp
```diff
@@ -73,6 +82,28 @@ namespace drawtypes {
     else
       text = conf.get<string>(section, name, def);
```
jaagr: Noticed some errors in the token parsing block. Here's a proposed patch:

- Skip all non-token values
- Support `%token:<min>%`
- Catch exception thrown on value conversion failure

~~~ cpp
size_t start, end, pos;
string line{text};

while ((start = line.find('%')) != string::npos && (end = line.find('%', start + 1)) != string::npos) {
  auto token = line.substr(start, end);

  line.erase(0, end);

  // ignore false positives (lemonbar-style declarations)
  if (token[1] == '{')
    continue;

  bound.emplace_back(bounds{0, 0});

  // find min delimiter
  if ((pos = token.find(':')) == string::npos)
    continue;

  // strip min/max specifiers from the label string token
  text = string_util::replace(text, token, token.substr(0, pos));

  try {
    bound.back().min = std::stoul(&token[pos + 1], nullptr, 10);
  } catch (const std::invalid_argument& err) {
    continue;
  }

  // find max delimiter
  if ((pos = token.find(':', pos + 1)) == string::npos)
    continue;

  try {
    bound.back().max = std::stoul(&token[pos + 1], nullptr, 10);
  } catch (const std::invalid_argument& err) {
    continue;
  }

  // ignore max lengths less than min
  if (bound.back().max < bound.back().min)
    bound.back().max = 0;
}
~~~

NBonaparte: That's a much neater way to do it. If you don't mind, could you show how `%token:min%` fails? While testing it, I didn't seem to encounter any errors.

src/drawtypes/label.cpp
```diff
@@ -64,15 +71,57 @@ namespace drawtypes {
    * Create a label by loading values from the configuration
    */
   label_t load_label(const config& conf, string section, string name, bool required, string def) {
+    vector<struct bounds> bound;
+    size_t start, end, pos;
+
     name = string_util::ltrim(string_util::rtrim(name, '>'), '<');
 
     string text;
+    string line{text};
```
jaagr: This would need to be moved down a few line until after `text` has been initialized.

NBonaparte: Right... what was I thinking?

src/drawtypes/label.cpp
```diff
@@ -13,19 +13,26 @@ namespace drawtypes {
...
+    m_token_iterator = m_token_bounds.begin();
   }
 
   bool label::has_token(string token) {
     return m_text.find(token) != string::npos;
   }
 
   void label::replace_token(string token, string replacement) {
-    m_tokenized = string_util::replace_all(m_tokenized, token, replacement);
+    if (m_text.find(token) == string::npos)
```
jaagr: add iterator safety-check `... || m_token_iterator == m_token_bounds.end()`

Issue #165, opened 2016-11-14
Add other states for bspwm dimmed desktops


edgard:
Hi Michael, one more idea. :D

On bspwm, you can configure overrides for dimmed desktop, however, it overrides any other state, including occupied, empty, urgent, etc. It would be nice if we could configure overriding dimmed granular to the other states. 

Use case:
On active monitor, I have a bar like [this](https://u.teknik.io/lSbWe.png) where active desktop have a blue overline and occupied ones have a gray overline.

However, if I want to make the dimmed desktops with a gray overline for occupied ones and no overline for empty, it's not possible at the moment (or I haven't figured out how?). If I override the overline color on dimmed, it applied to all desktops on that monitor.

jaagr:
Yeah that's how it works at the moment. That's how I styled my bar when I wrote it so that's why. But yeah, it would be a nice thing to have.

edgard:
ah, so, it wasn't just me goofing. :D nice to know. also, sorry for those issue opening lately. i really enjoy the project and would love to send some pr, but I have absolutely no experience with X/XCB programming. maybe in the future. :D

edgard:
merged #167

PR #167, opened 2016-11-15
feat(bspwm): Additional dimmed states configuration

edgard:
I tried implementing those additional dimmed states as in #165. See if this makes sense to you. It works here but I might be missing something.


jaagr:
Looks good, merging

Issue #84, opened 2016-10-14
List of feature requests


jaagr:
Hi,

While my main focus right now is to fix bugs so that we can release a stable version of 2.0 I'd still like to know what kind of features or changes people would like to add to the project. Now when lemonbar has been dropped and all rendering is done internally it opens up alot of doors.

My list of things to change:
- Drop xft and implement all text handling using pango
- Add include directive to the configuration
- Add inherit directive to the configuration

Quent42340:
Personally, I'd like to see some features like:
- Allowing `click-left`, `click-right`, etc... on `custom/script` to change directly the output on the bar
- The `interval` field for the same module should have an option `never` (ie. only updated by click, so related to the above suggestion)
- I didn't saw any `internal/disk-space` module, that would be nice
- Add a field `format-extended` which would work like the `date-detailed` field but on every module
- Don't know if it's possible but an overriding `--output` option would be great since inheritance isn't already implemented (that would avoid duplicating bar configs when using multi-monitor setup)

And for the repository itself:
- Move the `README.md` documentation into the wiki

jaagr:
New module: keyboard

Show the current keymap and active modifiers (caps, etc)

jaagr:
New module: xwindow

Show details about the focused X window (title, atom values, etc).

jaagr:
Add the ability to pause/resume the running bar process. Useful when launching a game, etc.

ashleymcdonald:
A system tray would be awesome. Would love to ditch stalonetray

jaagr:
@ashleymcdonald It's been implemented.

hthuong09:
I would love to have some more features like below

- Ability to automatically set padding of bspwm depend on bar height setting (and try reset when exiting bar)
-  Custom mouse action for all module (Right/Left/Middle/Scroll Up/Scroll Down)
- Reload when received signal for custom script (Like i3blocks)
- Documentation for modules developing?

ashleymcdonald:
Awesome. Would it be possible to have it work like a module regarding positioning? 

Ie. module-right = label tray clock

jaagr:
It's possible yeah, but it's not a small task and I don't think it's something I'll look into in the near future so don't hold your breath :-) Pull requests are always welcome tho.

mrkgnao:
How do I write my own module?

jaagr:
You have two options @mrkgnao:

- Define a script module that executes an external command (see the wiki)
[...]

PR #216, opened 2016-12-04
feat(tags): Add prefix and suffix tags to all formats

NBonaparte:
Here's my attempt at implementing prefix and suffix tags for all formats, as per my feature request in #84.

Prefixes and suffixes are named similarly to their formats (`format-charging` -> `<prefix-charging>`) if there are multiple formats, or just `<prefix>` when there is only one.

### DIFF
--- a/include/modules/meta/base.hpp
+++ b/include/modules/meta/base.hpp
@@ -9,6 +9,7 @@
 #include "components/config.hpp"
 #include "components/logger.hpp"
 #include "components/types.hpp"
+#include "drawtypes/label.hpp"
 #include "errors.hpp"
 #include "utils/concurrency.hpp"
 #include "utils/functional.hpp"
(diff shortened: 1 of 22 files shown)


jaagr:
Nice, merged with some changes:

- Move prefix/suffix label output to base class 
- Require label when format tag is used

jaagr:
Could you add a section about the tags to the wiki? https://github.com/jaagr/polybar/wiki/Formatting

jaagr:
By the way, wouldn't it be enough if the format itself had the prefix/suffix as params? Just to avoid boilerplate stuff in the config.

~~~ dosini
format-charging = ...
format-charging-prefix = ...
format-charging-suffix = ...
~~~

jaagr:
Like this https://github.com/jaagr/polybar/pull/217

NBonaparte:
It prevents the fringe case where someone would put a ramp/bar/etc. before the prefix, but I'm not sure if anyone would do that. If you're confident that this isn't an issue, I guess that way is better.

jaagr:
in my own config i will replace all prefix icons.. can't find any case where i would put it elsewhere.. but we could do a mix of both

NBonaparte:
I think the new way (#217) is good for the prefixes and suffixes, but we can add a static text tag or something like that later if needed.

PR #219, opened 2016-12-04
feat: Add left and right padding and margins

NBonaparte:
Since the `prefix` and `suffix` strings don't have spaces before/after, having separate left/right padding and margins is the only way to insert a space between them and the rest of the output.

### DIFF
--- a/include/drawtypes/label.hpp
+++ b/include/drawtypes/label.hpp
@@ -2,6 +2,7 @@
 
 #include "common.hpp"
 #include "components/config.hpp"
+#include "components/types.hpp"
 #include "utils/mixins.hpp"
 
 POLYBAR_NS
@@ -16,11 +17,6 @@ namespace drawtypes {
     size_t min;
     size_t max;
     string suffix{""};
-  };
-
-  struct sides {
-    int left;
-    int right;
   };
 
   class label;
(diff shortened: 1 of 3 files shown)


include/drawtypes/label.hpp
```diff
@@ -18,6 +18,11 @@ namespace drawtypes {
     string suffix{""};
   };
 
+  struct sides {
+    int left;
+    int right;
+  };
+
```
jaagr: you could use the type in `components/types.hpp`:

~~~cpp
struct side_values;
~~~

src/drawtypes/label.cpp
```diff
@@ -115,12 +127,33 @@ namespace drawtypes {
...
     if (required) {
       text = conf.get<string>(section, name);
     } else {
       text = conf.get<string>(section, name, move(def));
     }
 
+    const auto get_left_right = [&](string key) {
+      int left = conf.get<int>(section, key + "-left", 0),
+          right = conf.get<int>(section, key + "-right", 0),
+          all = conf.get<int>(section, key, 0);
```
jaagr: Use the `{padding,margin}` parameter as main and `-{left,right}` as override, i.e:

~~~ cpp
auto value = conf.get<int>(section, key, 0);
auto left = conf.get<int>(section, key + "-left", value);
auto right = conf.get<int>(section, key + "-right", value);
return {left, right};
~~~

PR #222, opened 2016-12-04
feat(mpd): Add support for date

mrkgnao:
A trivial addition to the mpd module. :)


mrkgnao:
As an example, here's the relevant snippet from my current `mpd` config:

    label-song = %artist% - %title% (%{T2}%album%%{T1}, %date:4:4%)

jaagr:
looks good to me, could you update the wiki with the new token

PR #223, opened 2016-12-04
fix(script): set default tag to label

NBonaparte:
Shouldn't the default tag be `<label>` instead of `<output>`? Otherwise, it would just complain that it's deprecated even though the user hasn't set anything.


jaagr:
yeah good catch

NBonaparte:
I'll change it in the wiki.

PR #227, opened 2016-12-05
Combine charge levels with two batteries

### LINKED ISSUE #538: Is it possible to show 2 batteries (BAT0, BAT1) in one module?
Hey, 
I am wondering if there is a way I can show both of my batteries in one battery module. 
It's a common setup for example for thinkpads which have often 2 batteries. The only thing I manage to do now is making 2 modules, one or BAT0 and the other for BAT1. It would be very nice if I can calculate them somehow togeher to one %percent% output.
I think the calculation would not be so hard, the most important part of recognizing different batteries works already smooth. Just something like dividing the results by 2: (BAT0 + BAT1)/2 = %percent%. Any hints if and when yes how I could accomplish this with polybar. It's my last missing piece to have the perfect bar :dancer: 
Thanks in advance.

vin-ivar:
So I've got two internal batteries, and I'm not a huge fan of having two separate battery blocks on my bar. This lets you add a second battery using `battery_snd` in your config, then averages out the charge levels for both. If you've left `battery_snd` blank, it just defaults it to your primary battery, then doubles and then halves the battery level.

Resolves #538.


ramblingenzyme:
I'd just like to note that if the batteries are not of the same capacity, simply averaging their charge levels won't be enough to get an accurate idea of what's going on. I've run into this personally, and needed to calculate the combined total charge and the current charge to get an accurate percentage.

patrick96:
Hey @vin-ivar. Do you still have time to work on this? If so, I'd also like to see this merged.
There have been quite a lot of changes to the battery module, so you would have to rewrite a good portion of your changes. From what I can tell ad0af86a7b43397fb62b1e1bd28506c106544623 is where most of the conflicts were introduced. If you do that and address @ramblingenzyme's concerns about batteries with different capacities, I think we would be quite close to merging this. And also use `strtof` instead of `atof`.
Sorry this took so long

IvanMalison:
...or you could do the sane thing and actually rewrite the battery stuff to use something like upower

patrick96:
This would add an additional dependency though. Also we don't really have the capacity to just rewrite a whole module right now.

PR #284, opened 2016-12-30
feat(bspwm): Add support for psuedo tiled nodes

bozbalci:
With this addition, adding

`label-psuedotiled = P`

will display P whenever a bspwm node is set to psuedo-tiled (the default keybind is Super+Shift+T).

Should be a trivial addition, but I compiled & tested on my computer nevertheless.

If approved, I'll edit the wiki to include the option.


bozbalci:
I misspelled "psuedo". Correcting that now. I'll close this one.

Issue #356, opened 2017-01-19
Abort/Segfault on start


Mitame:
Aborting run: https://gist.github.com/Mitame/4ee8c7f7452e33af42c9bfb2e44cc37b
Segfaulting run: https://gist.github.com/Mitame/592f711b819676140469ae30420749bf
config: https://gist.github.com/Mitame/fd805c6df2d8a04ebe33a1d067dd1aee

Polybar fails to start consistently, either segfaulting or aborting. Don't believe I've changed anything to make this happen from when it was working. 

EDIT: The logs are from commit 9e3a7cf06c17175e649692908c957dfdc0a8fab7, though this is also present on the latest release (2.5.2 at time of writing).

jaagr:
Hi, can you post the output of `xrdb -q`

Mitame:
No output from `xrdb -q`

Mitame:
-_- i'm a moron... let me edit config to remove those references

Mitame:
OK, error was having xrdb references in the config without having xrdb configured with those properties. Could do with a clearer error though.

jaagr:
It's supposed to default to an empty value.. I'll have a look

Mitame:
I have the full coredump from `coredumpctl` if that's useful?

jaagr:
Guess it's without debug symbols, but youmight as well add it

Mitame:
Well, it's coredump from git with `-DCMAKE_RELEASE_TYPE=Debug` or whatever the correct one was. Linked anyway
[Abort dump](https://u.nya.is/jdryen.dump)
[Segfault dump](https://u.nya.is/objyoc.dump)

jaagr:
Can't reproduce it.. will try a bit later on my arch machine.

Mitame:
Can't really say exactly how to reproduce it. I deleted my `.Xresources` file because its colours were messing up some other stuff and so when I logged back in, polybar never appeared and this error was occurring. Fixed it by putting it back and loading it back in with `xrdb .Xresources` but able to reproduce by removing the file and logging back in to remove the configurations from xrdb.

Also, running `polybar main --dump=PARAM` with the parameter as something from xrdb with a fall-back causes the error as well, though this always seems to abort rather than segfault.

jaagr:
I think 374ad65 might fix it

Mitame:
[...]

PR #357, opened 2017-01-19
Fix for #356

Mitame:
Makes sure `m_db` is a `nullptr` when it is unable to get the X resource manager.
Refs #356


jaagr:
The Xlib version has been dropped all together in favor of the xcb version.

PR #359, opened 2017-01-20
feat(bspwm,i3): Fuzzy-matching names for icons

opcon:
Thanks for this project, I recently switched from i3bar and am loving it! :)

I was playing around with setting icons for different workspaces, and wanted a way to match the icon id as a substring of the workspace name, since my web browser is not always on the same numbered workspace.

I implemented this along with a configuration switch for both bspwm and i3: `fuzzy-match`. It is false by default, which uses the old matching algorithm.

Let me know if anything is not right, I'm still new to c++. Thanks!

### DIFF
--- a/include/drawtypes/iconset.hpp
+++ b/include/drawtypes/iconset.hpp
@@ -13,7 +13,7 @@ namespace drawtypes {
    public:
     void add(string id, icon_t&& icon);
     bool has(const string& id);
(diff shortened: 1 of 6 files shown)


jaagr:
Hi and thanks for contributing. Looks good to me. Could you also update the wiki?

opcon:
No worries, thanks for merging! I've updated the wiki

PR #368, opened 2017-01-23
res_class==NULL to properly get the resource

quite:
Not sure what "String" was supposed to do, my tests with xcb-xrm gives that
res_class==NULL properly gets the resource value, =="String" does not.

### DIFF
--- a/include/x11/xresources.hpp
+++ b/include/x11/xresources.hpp
@@ -36,7 +36,7 @@ class xresource_manager {
   template <typename T>
   T require(const char* name) const {
     char* result{nullptr};
-    if (xcb_xrm_resource_get_string(m_xrm, string_util::replace(name, "*", ".").c_str(), "String", &result) == -1) {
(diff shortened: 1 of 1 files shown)


jaagr:
It was a quick rewrite so it's probably left from the xlib version.

include/x11/xresources.hpp
```diff
@@ -36,7 +36,7 @@ class xresource_manager {
   template <typename T>
   T require(const char* name) const {
     char* result{nullptr};
-    if (xcb_xrm_resource_get_string(m_xrm, string_util::replace(name, "*", ".").c_str(), "String", &result) == -1) {
+    if (xcb_xrm_resource_get_string(m_xrm, string_util::replace(name, "*", ".").c_str(), NULL, &result) == -1) {
```
jaagr: Use `nullptr` instead of `NULL`

quite:
Corrected.

PR #371, opened 2017-01-24
Ensure reloading when IN_IGNORED fired on config

skystrife:
This fixes a "bug" where polybar wouldn't reload on a configuration file change on some configurations of vim, which don't actually issue any `IN_MODIFY` events because they choose to move the file, replace it with a new one, and then delete the file instead.

To work around this, we now also listen for `IN_IGNORED` which fires when the file we are watching is destroyed. When this happens, we re-attach the configuration file watcher to the new file and reload.


jaagr:
Looks good

Issue #363, opened 2017-01-22
Noto Emoji not displaying?


sphaugh:
I followed the steps to use Noto Emoji as the font for my workspace icons, but they are still not appearing. How can I fix this or figure out what's going wrong? Thanks.

jaagr:
What steps did you follow?

skystrife:
I can replicate this issue. It occurs for me when I attempt to use the following emoji: 🔇 (U+1F507). I've traced the problem down to parsing the configuration file.

Specifically, `parser::text()` refuses to emit any character that is larger than the size of a `uint16_t`. This emoji lies outside of that, so it instead writes `0xfffd` (the replacement character).

Why is `write_text_unicode` a `value_signal<write_text_unicode, uint16_t>`? I feel like you actually need the full `uint32_t` here to support all possible UTF codepoints, right?

jaagr:
@skystrife 4-byte sequences aren't supported (goes beyond the sized integral type).. I'm almost done with the new implementation though, available on the `refactor-cairo` branch

jaagr:
I've pushed the changes to the master branch, so please try it out.

skystrife:
Yes, that seems to fix the issue.

Is it expected, though, that the new renderer makes fonts smaller than they were before? Using the same font settings, for example, produces

![font-sizes](https://cloud.githubusercontent.com/assets/1378336/22263062/c4e9a04e-e238-11e6-88e8-ee64ab091999.png)

Where the top one is the master branch and the bottom one is v2.5.2. (I can create a separate issue for this if you'd like, though, since the emoji now render properly for me.)

jaagr:
No it's not intended but there's currently no scaling of point sizes. Will look into it.

jaagr:
Could you post the output of 

~~~
$ xrdb -q | grep Xft
~~~

skystrife:
No output (empty). This is all that's there:

```
$ xrdb -q
rofi.color-active:	#191a16, #ffb52a, #21211b, #33332b, #ffb52a
rofi.color-enabled:	true
rofi.color-normal:	#191a16, #dbdbd5, #21211b, #33332b, #dbdbd5
rofi.color-urgent:	#191a16, #f9266a, #21211b, #33332b, #f9266a
rofi.color-window:	#191a16, #002b37, #33332b
rofi.font:	Meslo LG M DZ 12
rofi.separator-style:	solid
```

jaagr:
What about 

~~~
[...]

PR #373, opened 2017-01-25
Use proper font scaling based on DPI

skystrife:
(See #363). This tries to mimic the old renderer's behavior as closely as possible. In the absence of any information, DPI is assumed to be 96x96. DPI can be configured on a per-bar basis using the configuration key "dpi".

To use the DPI configuration from Xresources (if built with support), one can specify the following in the bar config:

```
dpi = ${xrdb:Xft.dpi:96}
```

Individual keys `dpi-x` and `dpi-y` are supported. Negative values are taken to mean "automatic" where the DPI will be calculated from screen features (accuracy here will depend on your hardware, YMMV).

### DIFF
--- a/include/cairo/font.hpp
+++ b/include/cairo/font.hpp
@@ -30,7 +30,7 @@ namespace cairo {
     virtual string name() const = 0;
     virtual string file() const = 0;
     virtual double offset() const = 0;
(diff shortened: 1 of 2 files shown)


jaagr:
Nice, looks good.

Issue #393, opened 2017-01-29
Windows Title with #Number makes polybar disappear


qgnox:
Version 3.0.3 , using Chrome or Firefox when it loads a url/page that in title has a #number the status bar disappear when using xwindow module , systray stays. log info shows : 

warn: Dropping unmatched character  (U+0002) 

http://imgur.com/ufyuu88l.png

Url/Page example : github.com/jaagr/polybar/issues/392

jaagr:
It works fine for me. Can you post your config

qgnox:
This is my config, tested with and without compton , and also deactivating i3 "for_window *** floating enable" same behaviour . with 2.5 version same config no problems. 

`;=====================================================
;
;   To learn more about how to configure Polybar
;   go to https://github.com/jaagr/polybar
;
;   The README contains alot of information
;
;=====================================================

[colors]
;background = #222
background = #586e75
background-alt = #444
foreground = #dfdfdf
;foreground-alt = #55
foreground-alt = #dfdfdf
primary = #ffb52a
secondary = #e60053
alert = #bd2c40

[global/wm]
margin-top = 0
margin-bottom = 1

[bar/example]
;monitor = ${env:MONITOR:HDMI-1}
width = 100%
;height = 27
height = 25
offset-x = 0
offset-y = 0

;background = ${xrdb:color9}
background = ${colors.background}
foreground = ${colors.foreground}

overline-size = 2
overline-color = #f00
underline-size = 2
underline-color = #00f

border-bottom-size = 2
border-bottom-color = #333

padding-left = 0
padding-right = 2
module-margin-left = 1
module-margin-right = 2

;font-0 = fixed:pixelsize=10
font-0 = xos4 Terminus:pixelsize=10
font-1 = unifont:fontformat=truetype:size=6:heavy;-2
;font-1 = Terminus:size=6:heavy;-2
font-2 = Wuncon Siji:pixelsize=10
;font-2 = FontAwesome:pixelsize=10 
modules-left = i3 xwindow
modules-center =  
modules-right = mpd backlight-acpi volume filesystem memory cpu wlan eth battery temperature date powermenu

tray-position = right
tray-padding = 2
;tray-transparent = true
;tray-background = #0063ff

;wm-restack = bspwm
;wm-restack = i3

;override-redirect = true

;scroll-up = bspwm-desknext
;scroll-down = bspwm-deskprev

;scroll-up = i3wm-wsnext
;scroll-down = i3wm-wsprev

[...]

PR #434, opened 2017-02-14
fix(label): Truncate label replacements based on codepoint count

skystrife:
This helps ensure that when a string is truncated it is not done in the middle of a utf8 multi-byte sequence.

This doesn't 100% correspond to user-perceived characters, but it should be pretty close in most cases.

I think this might fix issue #393.


jaagr:
Thank you

PR #440, opened 2017-02-16
fix(builder): Use unicode-aware truncation for label-maxlen

skystrife:
This should fix the remaining problems in issue #393, since I missed a spot in the last PR.


jaagr:
Thanks

PR #449, opened 2017-02-21
feat(xworkspaces): add urgent desktop detection

NBonaparte:
Listens for `WM_HINTS` changes in all windows and checks if the urgency bit is set, and sets the corresponding desktop state accordingly.


jaagr:
Thank you

PR #450, opened 2017-02-21
fix(build): POSIX compliant command usage

uggedal:
Shells like dash do not support printing paths of more
than one argument with the builtin `command -vp`.


jaagr:
Thank you

PR #455, opened 2017-02-22
fix(bspwm): Update when focusing urgent desktops

adamransom:
When focusing a desktop with the urgent flag, two events are received
from `bspc` simultaneously, separated by a newline character.

This was not handled correctly and the second event was discarded causing
the urgent style to be removed, but the focused style would remain on the
previously focused desktop.

This fixes the problem by handling any number of events that arrive at the
same time (separated by newlines).


jaagr:
Thank you

PR #456, opened 2017-02-22
WIP: MPRIS module

jereksel:
This pull requests adds MPRIS module suggested here: https://github.com/jaagr/polybar/issues/84#issuecomment-267010859.

Basic stuff (song name/artist, pausing, stopping works). I must find a way to generate .c and .h from .xml in CMake and logic must be also reviewed. I was also thinking about combining mpd and mpris modules, but there seems to be a lot of logic differences.

TODO:
- [x] Fix memory leaks
- [ ] Generate .c with dbus at compile time
- [ ] Implement Shuffle
- [ ] Implement Position


jaagr:
I will check out the pr later today. As for the logic of the mpd module you should probably ignore it as it's rather messy.

fa7ad:
*bump* any update on this?

Lomadriel:
Hello @jereksel, 
Do you still want to work on this feature?
If you don't want, I'll continue to implement that feature.

jereksel:
@Lomadriel 
Feel free to take it, I don't use polybar anymore.

patrick96:
@Lomadriel another idea we had was to implement a playerctl module. It's interface is a lot simpler than implementing the whole of the MPRIS spec

Ref: https://github.com/polybar/polybar/projects/1#card-10511466

fa7ad:
@patrick96 Yeah, that's what I (and most other people I know) use right now. A custom module that runs a playerctl based script periodically.

Lomadriel:
Hum okay, this might be a good idea but my goal was to add dbus into polybar to allow to use other bus interface like for example upower (to fix the battery problems).
I didn't know that playerctl had an api (I always thought it was just a command line program)

patrick96:
@Lomadriel I think you can agree with me that we should postpone this until we are done with some of the major refactoring. I am also wondering if it might not be a little overkill to add DBUS for only the backlight.

Lomadriel:
> @Lomadriel I think you can agree with me that we should postpone this until we are done with some of the major refactoring. I am also wondering if it might not be a little overkill to add DBUS for only the backlight.

Yeah that's what I was planning to do, I think it is better to wait (at least) until we split polybar in modules.

PR #461, opened 2017-02-25
install_manifest has a .txt extension

saboorian:
As you can see [here](https://cmake.org/Wiki/CMake_FAQ#Can_I_do_.22make_uninstall.22_with_CMake.3F), `install_manifest` file reference in this template is missing an extension.


jaagr:
I removed it. The cmake snippet you reference is just an example

saboorian:
I don't understand why the extension was removed. I can't find any reference to extension-less one.
[This](https://cmake.org/gitweb?p=cmake.git;a=blob;f=Source/cmLocalGenerator.cxx;h=07d712c0a3c08589649176a2f991e4bb85b31288;hb=refs/heads/master#l483) is apparently the CMake part responsible for building that file.

jaagr:
I dislike extensions that doesn't add any semantic value to the file. Didn't know it was used by cmake internally tho so I'll merge your pr.

PR #482, opened 2017-03-12
fs: Fix wrong size report

nico202:
Hi, I found that when the available fs free size was under 1Gb, it reported the right MB size but in GBs (like: 943 GB).

This should fix it

### DIFF
--- a/src/utils/string.cpp
+++ b/src/utils/string.cpp
@@ -257,15 +257,16 @@ namespace string_util {
   /**
    * Create a filesize string by converting given bytes to highest unit possible
    */
-  string filesize(unsigned long long kbytes, size_t precision, bool fixed, const string& locale) {
-    vector<string> suffixes{"TB", "GB", "MB"};
-    string suffix{"KB"};
-    double value = kbytes;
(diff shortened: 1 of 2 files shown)


jaagr:
Thanks! Could you also update `tests/unit_tests/utils/string.cpp#101` (should be 3 MB)

nico202:
Sorry not to have run the tests :D
Yes but the problem here is that here there is a discrepancy between your comment:
> Create a filesize string by converting given bytes to highest unit possible
and the variable name (kbytes)

Also, if this is really taking bytes we need to change the sizes to:
>    vector<string> suffixes{"TB", "GB", "MB", "KB"};
>   string suffix{"B"};
I added some new test (B/KB/GB/TB), let me know

jaagr:
Your changes makes sense. Probably got mixed up somewhere down the line. I'll double check my test modules in a few hours.

nico202:
You refactored the things in this commit: c2ac7fde
If I see well it was wrong at least since then, no tests on this existed before

jaagr:
Works fine, thanks

Issue #503, opened 2017-03-28
scrolling through workspaces on multiple monitors


thamma:
I assume this is a bug. I'm running polybar v3.0.4 with the i3 module on i3wm v4.13.

# Reproduction:
Using two monitors, both running one instance of polybar, having their separate configs set up. I scroll the workspaces on the first polybar instance, then move to the other monitor to scroll through the workspaces on the second polybar instance.

# Expected behaviour:
When scrolling on the bar of one monitor, I expect the scrolling to only affect the (associated) workspaces of that monitor. Switching over to the other monitor should yield the same behaviour for that bar.

# Actual behaviour:
Scrolling on the first monitor works as expected. However if I move my cursor to the second monitor, hover the polybar and start scrolling, I'm scrolling through the workspaces of the first monitor. It appears that the second polybar doesn't become focused unless I manually click it. Clicking the bar before scrolling yields the expected behaviour.

Maybe I'm not setting up my two monitor setup correctly and should rather use some settings designed to be used with two monitors. Please hint me on that, if that's the case, as I couldn't find the proper settings using google and the wiki.

Best regards

jaagr:
Hi and thanks for your well written issue. Did you try the module parameter `pin-workspaces = true`?

TonCherAmi:
@jaagr hey, I can confirm this behavior, I also do have `pin-workspaces` set to `true` and that does not seem to resolve the issue.

thamma:
Can confirm, `pin-workspaces` does not change the behaviour. I already used that.

jaagr:
Then I'd also say it's a bug. Will have a look later.

TonCherAmi:
@jaagr hey I almost finished working on the fix. I will open a PR soon I think.

jaagr:
Great, then I'll let you finish that up.

jaagr:
@TonCherAmi, could you also fix the bspwm and xworkspaces modules? I would guess that they have the same problem.

TonCherAmi:
@jaagr I'm not sure to be honest, I've never dealt with bspwm. But I could check it out tomorrow I think.

jaagr:
The scroll code should be pretty similar to the one used in the i3 module.

TonCherAmi:
Yeah, it's just that I'll need to install and configure it in order to check it, might take some time.
Also, the problem in i3 module lies in the ipc command, so I'm not really sure that those modules have the same issue.

jaagr:
[...]

PR #510, opened 2017-03-29
fix(i3): Check whether current ws is focused before issuing scroll commands

TonCherAmi:
Fix for #503 

Any critique is very much welcome!

### DIFF
--- a/src/modules/i3.cpp
+++ b/src/modules/i3.cpp
@@ -198,18 +198,20 @@ namespace modules {
     }
 
     try {
-      using namespace i3_util;
-
-      string scrolldir;
-      const connection_t conn{};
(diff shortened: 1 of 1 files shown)


thamma:
Works perfectly fine to me! Great job and thank you. Now nothing is holding me back from switching over from i3blocks.

> (I fetched and built the PR, ran it, realized the issue was not fixed, questioned the fix, was using `polybar` over `./polybar` all the time, smh)

TonCherAmi:
Duh, I realized that I kind of overengineered this, since current solution will work if we check whether the bar's workspace is focused or not.
I'll add a commit momentarily.

jaagr:
Just tried the patch and it works fine for me as well. Merging.

PR #521, opened 2017-04-06
fix(i3): Use ws name for click handling

TonCherAmi:
Hey,

This fixes clicking on workspaces without an index.


jaagr:
Thanks. Could you verify that workspaces with `:` in the name works (the builder should already handle the escaping).

TonCherAmi:
Sure, as soon as I get home.

TonCherAmi:
Yep, works as intended for me.

Issue #524, opened 2017-04-08
MPD module output on stopped playback


TonCherAmi:
Hey,

I was wondering whether it is possible to customize the output of the MPD module for the case when playback is stopped, since as of now, at least for me, it looks quite ugly:
* Stopped
![stopped](https://cloud.githubusercontent.com/assets/13941584/24831170/5e658d34-1c9d-11e7-9cda-1b2ad00cfe21.png)

* Playing 
![playing](https://cloud.githubusercontent.com/assets/13941584/24831162/30fd953a-1c9d-11e7-976b-d1da8a4c8f1c.png)

As per wiki I couldn't find a straightforward way of doing it, so unless I'm missing something, I thought that perhaps we should address the matter somehow. For example in addition to `format-online` we could have state-specific format parameters.

My MPD config section just in case:
```
[module/mpd]
type = internal/mpd

host = 127.0.0.1
port = 6600

format-online = <icon-prev>  <icon-stop>   <label-song>   <icon-play> <icon-pause>  <icon-next>

icon-prev = 
icon-stop = 
icon-pause = 
icon-play = 
icon-next = 

label-song-maxlen = 50
label-song-ellipsis = true

format-offline = 

format-online-underline = #AA
format-online-padding = 1
```

jaagr:
We could maybe add some more output formats, like `format-playing`, `format-stopped`, etc.

TonCherAmi:
Yeah, that sums up what I had in mind quite well, will look into implementing this.

kamek-pf:
In the same vein, it would be nice to have the ability to display something when the playlist is empty.

PR #567, opened 2017-05-06
feat(mpd): Add state-specific formats

TonCherAmi:
@jaagr hey, 

This implements #524. 

Basically, in `module_formatter::add()` we check whether `fallback` is a format itself and try to get its value from the config if it is. That allows us to specify `format-online` as a fallback format for `format-playing`, `format-paused` and `format-stopped`.

I guess that the biggest downside to this approach is that we lose the ability to specify a default value in the case where no formats are found in the config. We could avoid that by turning  `fallback` into a vector, but I suppose that would affect all modules, so it might be a bit of an overkill. I'd like to hear your thoughts on this.

### DIFF
--- a/include/modules/mpd.hpp
+++ b/include/modules/mpd.hpp
@@ -29,6 +29,9 @@ namespace modules {
 
    private:
     static constexpr const char* FORMAT_ONLINE{"format-online"};
+    static constexpr const char* FORMAT_PLAYING{"format-playing"};
+    static constexpr const char* FORMAT_PAUSED{"format-paused"};
+    static constexpr const char* FORMAT_STOPPED{"format-stopped"};
     static constexpr const char* TAG_BAR_PROGRESS{"<bar-progress>"};
     static constexpr const char* TAG_TOGGLE{"<toggle>"};
     static constexpr const char* TAG_TOGGLE_STOP{"<toggle-stop>"};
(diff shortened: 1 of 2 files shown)


jaagr:
It looks good to me, but I'd recommend getting the `format-online` value before adding the formats to avoid breaking anyone's config (since it previously had a default value). It would also remove the need of 66095e3.

For example:

~~~ diff
diff --git a/src/modules/mpd.cpp b/src/modules/mpd.cpp
index edcea9f..327ee43 100644
--- a/src/modules/mpd.cpp
+++ b/src/modules/mpd.cpp
@@ -20,8 +20,9 @@ namespace modules {
     m_synctime = m_conf.get(name(), "interval", m_synctime);
 
     // Add formats and elements {{{
+    auto format_online = m_conf.get<string>(name(), FORMAT_ONLINE, TAG_LABEL_SONG);
     for (auto&& format : {FORMAT_PLAYING, FORMAT_PAUSED, FORMAT_STOPPED}) {
-      m_formatter->add(format, FORMAT_ONLINE,
+      m_formatter->add(format, format_online,
           {TAG_BAR_PROGRESS, TAG_TOGGLE, TAG_TOGGLE_STOP, TAG_LABEL_SONG, TAG_LABEL_TIME, TAG_ICON_RANDOM,
               TAG_ICON_REPEAT, TAG_ICON_REPEAT_ONE, TAG_ICON_PREV, TAG_ICON_STOP, TAG_ICON_PLAY, TAG_ICON_PAUSE,
               TAG_ICON_NEXT, TAG_ICON_SEEKB, TAG_ICON_SEEKF});

~~~

TonCherAmi:
Yup, it looks much nicer this way. :+1:

jaagr:
Thank you

PR #576, opened 2017-05-10
feat(xwindow): allow pinning window titles to output where the bar is displayed

### LINKED ISSUE #252: feat(xwindow) Monitor locality: show (last) active window title on bar's monitor
This would be useful. But it is also somewhat annoying to have the bars xwindow indicator changing on a monitor that I'm not actually working on. A flicker in the corner of my eye.

Additional, the ability to configure label-dimmed-* like for bspwm module would be great.

raidzero:
Having the active window title displayed on all bars did not make sense to me. With this pull request I added a config value to skip displaying the window title on bars on other monitors.

-boolean config property called "pin-output". defaults to false
-window title is only displayed on bar on currently focused monitor
-window geometry from xcb is used to determine a window's position

Fixes #252.

### DIFF
--- a/include/modules/xwindow.hpp
+++ b/include/modules/xwindow.hpp
@@ -18,6 +18,7 @@ namespace modules {
 
     bool match(const xcb_window_t win) const;
     string title() const;
+    xcb_window_t get_window() const;
 
    private:
     xcb_connection_t* m_connection{nullptr};
@@ -40,10 +41,13 @@ namespace modules {
 
    private:
     static constexpr const char* TAG_LABEL{"<label>"};
+    bool active_window_on_monitor(xcb_window_t window, monitor_t monitor) const;
 
     connection& m_connection;
     unique_ptr<active_window> m_active;
     label_t m_label;
+
+    bool m_pinoutput{false};
   };
 }
 
--- a/src/modules/xwindow.cpp
+++ b/src/modules/xwindow.cpp
@@ -58,11 +58,41 @@ namespace modules {
     }
   }
 
+  xcb_window_t active_window::get_window() const {
+    return m_window;
+  }
+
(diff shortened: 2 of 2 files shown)


jaagr:
Thank you for contributing. I'll check it out once I have some time.

raidzero:
Well, out of nowhere I decided to give bspwm a try. Turns out, this does not work on bspwm. I'll see about fixing that.

raidzero:
This may show just how little I understand X11, but on bspwm using xcb to query window size is returning unexpected values. The positions are being given back doubled. A window is at actual horizontal pixel 6459, but xcb is saying 12918.

raidzero:
I did some testing with other WM's (i3, openbox, fluxbox, awesome, xfce, lxqt, lxde, mate at least). bspwm is the only one that I have seen that exhibits this behavior. The workaround is to not use translated coordinates relative to the root window, since for whatever reason on bspwm the coordinates relative to parent window (given by xcb_get_geometry()) are correct. I cleaned up the active_window_on_monitor() function and put in a check for the bspwm socket. If it is found, do not check the window position using the translated coordinates. I do not know a better way around this bspwm oddity. 

TLDR: the last commit implements a workaround which fixes detection of window positions on bspwm.

betaboon:
I am interested in this feature.
i just tried it out against current master after compensating for bit-rot.
so i am now using this patch: https://gist.github.com/betaboon/5483f73600532c69d2ea5b2713aa162f

betaboon:
@raidzero @jaagr so how to move this forward?

NBonaparte:
@betaboon I have a somewhat stale PR #1122 which is basically a refactor of this. The old method is somewhat hacky so I decided to rewrite it, but it seems I forgot to close this PR. I'll take a look at fixing it up in the next few weeks; you are welcome to help test and/or fix anything.

Issue #411, opened 2017-02-03
Why notify the desktop to be inactivate?


DarwinSenior:
So, here is my question, I noticed the function `rebuild_desktop_states` in `src/modules/xworkspaces.cpp` sets active desktop to empty? Why that is case? Should not the workspace that is active always be active until it is not active?

```cpp
  /**
   * Update active state of current desktops
   */
  void xworkspaces_module::rebuild_desktop_states() {
    for (auto&& v : m_viewports) {
      for (auto&& d : v->desktops) {
        if (d->index == m_current_desktop && d->state != desktop_state::ACTIVE) {
          d->state = desktop_state::ACTIVE;
        } else if (d->state == desktop_state::ACTIVE) {
          d->state = desktop_state::EMPTY; // why deactivate it?
        } else if (d->label) {
          continue;
        }
```

jaagr:
The state might be named badly but active is actually the state for focused workspaces.

DarwinSenior:
Is it possible to have a current workspace? I have just switched to xmonad and I find it most useful.

Like this:

```cpp
  /**
   * Update active state of current desktops
   */
  void xworkspaces_module::rebuild_desktop_states() {
    for (auto&& v : m_viewports) {
      for (auto&& d : v->desktops) {
        if (d->index == m_current_desktop) {
          d->state = desktop_state::ACTIVE;
        } else {
          d->state = desktop_state:EMPTY;
        }
   }
```

NBonaparte:
Closing, as this should be fixed by #587. Not sure why it didn't automatically close.

PR #587, opened 2017-05-17
Fixes #535 and #411.

### LINKED ISSUE #535: xworkspaces module with XMonad updating incorrectly
I'm running XMonad with desktopConfig and the polybar xworkspaces module.

The module displays a list of the names of all my workspaces, and highlights the name of the active workspace. It correctly updates which workspace name is highlighted when I move focus from one workspace to another, but when I move focus between windows within a workspace, it stops highlighting any of the workspace names until I switch to another workspace.

I think this might be related to #411 but I'm not sure.

robertseaton:
The ewmh module doesn't properly set the "active" workspace when using XMonad (and perhaps others) as referenced in #535 . Applying the code from #411 fixes that.


jaagr:
Thanks

PR #631, opened 2017-07-04
Add more information on packages

AMDmi3:
As per discussion in https://github.com/jaagr/polybar/commit/802e69abafa575e3e38d36cca0c1ce07028dfb2e

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,12 +58,20 @@ Some of the services included so far:
 
 ## Getting started
 
+<a href="https://repology.org/metapackage/polybar">
+    <img src="https://repology.org/badge/vertical-allrepos/polybar.svg" alt="Packaging status" align="right">
+</a>
+
 If you are using **Arch Linux**, you can install the AUR package [polybar-git](https://aur.archlinux.org/packages/polybar-git/) to get the latest version, or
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,12 +58,20 @@ Some of the services included so far:
 
 ## Getting started
 
+<a href="https://repology.org/metapackage/polybar">
```
jaagr: Markdown has tags for this, e.g:

~~~ md
[![Packaging status](https://repology.org/badge/vertical-allrepos/polybar.svg)](https://repology.org/metapackage/polybar)
~~~

AMDmi3: There's a purpose for using HTML here: only with HTML it's possible to use ```align="right"``` attribute which makes the badge right-aligned and floating, so it doesn't create a gap in text, [see how it looks](https://github.com/AMDmi3/polybar/tree/more-packages#getting-started), which IMO is the best way of embedding it. If you don't like it I can turn it into markdown (thus make it non floating), a plain text link, or remove it completely.

jaagr: Oh I missed that part. Totally agree with you.

jaagr:
Thanks

PR #632, opened 2017-07-04
Disable -Werror for parentheses-equality warning

AMDmi3:
This warning fires here:

```
/home/travis/build/jaagr/polybar/src/utils/command.cpp:135:46: error: equality comparison with extraneous parentheses [-Werror,-Wparentheses-equality]
    } else if ((((*(int *) &(m_forkstatus))) == 0xffff)) {
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~
```

and as this system macro cannot be changed, let's not turn this into an error


jaagr:
Thanks

Issue #644, opened 2017-07-11
Error 2 when attempting to build cxx object bar.cpp.o


AceOfHurts:
I have been trying to build polybar from the source and keep running into this issue.  I have never filed an issue before so I'm not sure how much I should include.  In the terminal before the error 2 stuff there is a long list of errors.  They are all relating to xcb_randr_get_monitors_monitors and .  I attempted to use the fix in the issue thread #633 but still ran into troubles.  

bin/CMakeFiles/polybar.dir/build.make:182: recipe for target 'bin/CMakeFiles/polybar.dir/components/bar.cpp.o' failed
make[2]: *** [bin/CMakeFiles/polybar.dir/components/bar.cpp.o] Error 1
CMakeFiles/Makefile2:798: recipe for target 'bin/CMakeFiles/polybar.dir/all' failed
make[1]: *** [bin/CMakeFiles/polybar.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2

patrick96:
Can you post the output you got when you ran `cmake ..` and also the whole output of make (in a gist or something like that)

AceOfHurts:
Here is the gist for when I run [make](https://gist.github.comAceOfHurtsd19eed23ac79c30a7058991d8ac1fbff).  Here is the gist for running only [cmake ..](https://gist.github.com/AceOfHurts/2a57fa00cd6cc6a13bfa5687eac0ad4f) and lastly when I run [cmake with clang](https://gist.github.com/AceOfHurts/2624d42db9c586da3e86acdecf49b71e).  Unfortunately running cmake with clang didn't seem to work.

patrick96:
Did you run `make` after you configured cmake with clang? If so can you post the output for that too.

It seems this may be a problem with your `libxcb`, what version do you have installed?
Also what distro are you running?

AceOfHurts:
Yes I did run make after I configured cmake with clang but I still got an Error2 here is the [gist](https://gist.github.com/AceOfHurts/b15b64bfa956514b98624e2ee1aae385).  I am running Ubuntu 17.04 with openbox as my wm.  As far as libxcb goes I am running 1.12.

patrick96:
Alright, that is really strange, it complains about a few functions that are not defined, but they should be part of the `libxcb` version you are running.

Can you make sure that you have installed all of the packages listed in the [Wiki](https://github.com/jaagr/polybar/wiki/Compiling)

[...]

PR #649, opened 2017-07-12
Add textual indicator for module support

patrick96:
If color output was surpressed, there was no way to tell in the cmake                                                              
summary, if a module is actually enabled or not.               
Now each module is prefixed with either a "[X]" to indicate it's enabled                                                           
or a "[ ]" to indicate it's not.

This is something that could have saved a lot of hassle in #644 

EDIT: It now looks like this:
2017-07-13-003303_294x311_scrot


jaagr:
Thank you!

PR #701, opened 2017-08-17
Added Support for Mixed Percent/Pixel Values for Bar Geometry

Luke-Diamond:
This allows for the `width`, `height`, `offset-x`, and `offset-y` configuration options to be set in one of the following formats:

    width = 100%:-10px
    width = 100%:-10
    width = 100%
    width = 100px
    width = 100

This allows for pixel offsets on percent values. For example, with i3-gaps, you can line up the bar so that it is flush with the windows on the screen. This change does not require any configuration changes, but simply adds an optional offset.


lukediamond:
image

I fixed some parsing weirdness, changed all `atoi`'s to `strtof`, moved some stuff around, and tested it. Everything seems to work as expected.

I think that's all, feel free to change/move anything.

lukediamond:
image

The Travis failure appears to be on your end, it's a missing package. All other checks succeeded.

patrick96:
Do you no longer want to work on this?

patrick96:
I have now pulled your changes onto my own branch, added my tests and merged the whole thing. Thanks for the implementation

PR #717, opened 2017-08-30
fix(cmake): Suppress -Wnoexcept-type warning on GCC

skystrife:
There is no way to avoid this warning in the code itself, so we must suppress this warning if we wish to keep -Werror.

This is part of a series of pull requests that restore building of polybar on GCC. Once this is merged along with jaagr/xpp#6 and jaagr/i3ipcpp#2 (and the submodules updated after those merge), the project should build again with GCC 7.1.


jaagr:
Thank you!

Issue #544, opened 2017-04-19
format-NAME-prefix-fg overrides format-NAME-fg


alexciarlillo:
Setting foreground/background colors on the prefix seem to override the "regular" set values. e.g.:
```

format-volume-prefix-background = ${colors.white}
format-volume-prefix-foreground = ${colors.black}
format-volume-background = ${colors.black} // these 
format-volume-foreground = ${colors.white} // dont get applied
```

It seems that the colors are not being reset to the proper values after the prefix is printed. Or am I just doing something wrong?

jaagr:
Could be a bug. Could you post one output line from running polybar with the `-s` flag?

alexciarlillo:
I created a simplified bar to test this. Here is the config and the -s output:

config:
```
[colors]
black = ${xrdb:color0:#282828}
white = ${xrdb:color15:#ebdbb2}

[bar/test]
width = 100%
height = 25
radius = 0
fixed-center = true

background = ${colors.background}
foreground = ${colors.foreground}

font-0 = "Hack:size=10;2"
font-1 = "FontAwesome:size=10;2"
font-2 = "Hack:size=20;3"

modules-left =
modules-center = voltest
modules-right =

[module/voltest]
type = internal/volume

format-volume-padding = 0
format-volume-background = ${colors.white}
format-volume-foreground = ${colors.black}
format-volume-prefix-background = ${colors.black}
format-volume-prefix-foreground = ${colors.white}
format-volume-prefix = " prefix "
format-volume = "<label-volume> <bar-volume> "
label-volume = " "

format-muted = <label-muted>
label-muted = " muted "

bar-volume-width = 8
bar-volume-gradient = false
bar-volume-indicator = 
bar-volume-indicator-font = 2
bar-volume-fill = ─
bar-volume-fill-font = 2
bar-volume-empty = ─
bar-volume-empty-font = 2
bar-volume-empty-foreground = ${colors.white}
```

output:
```
ciarlill@mervyn2 ~ $ polybar test -s
warn: No monitor specified, using "eDP1"
%{c}%{A1:volmute: A4:volup: A5:voldown: B#ebdbb2 F#282828 B#282828 F#ebdbb2} prefix %{B- F-}  %{T2}──────%{T2}%{T- F#ebdbb2 T2}─%{T- F-} %{F- B- A A A}
```

jaagr:
Thanks I'll have a closer look at it after work.

jaagr:
[...]

PR #729, opened 2017-09-07
fix(modules): Separate prefix/suffix tags, revert tag stack

### LINKED ISSUE #639: Color Stack Behaviour
In f754fa27103e0df9a10b0083babd7ad29605e016 you introduced color stacks to fix the `-prefix` color problems described in #544.
With those color stacks the following module will make subsequent modules use `#373B41` as the background color unless they explicitly specify another color.
```dosini
[module/text]
type = custom/text
content = "%{B#373B41}1%{B#DE935F}3%{B-}"
```
Is this now intended behaviour and are format tags now supposed to be used in pairs (like HTML tags) so that after a module is processed the stack is empty again?

If this isn't intended behaviour, then maybe stacks are not the solution to this problem. It would probably be better to inject the color tags for the main section in between the prefix and the main section of the module as well as before the whole module. Or determine the colors for the prefix and main sections seperately and putting the prefix tags only in front of the prefix section and the same for the main section with the prefix section having two fallbacks: the color of the main section and the default color for the bar.

If this is intended behaviour, then there should probably be a way to totally reset the format tags (not just a simple reset that merely pops a value from the stack).

With the latter I may be able to help and submit a PR but for the former I have no idea how to do this, since I haven't figured out where and how the prefixes are handled

EDIT: See #666, #684, #575,  #736  for how this affects progressbars (the progressbar doesn't properly close its color tags)

NBonaparte:
Fixes #639 and #544, as requested by @patrick96. Provides a non-breaking solution to the problem posed in #544.

Stacks are removed (including the font index `%{T}`), but now the tags of prefixes and suffixes are separated from the main output.

There might still be some optimizing to be done, as the reset flag `%{X-}` still sometimes appears when switching between two defined values: `%{B#8d7e45}prefix %{B- B#feffff}main`. Also, module-wide settings (`format-<type>`) are not applied to the prefix and suffix.

### DIFF
--- a/include/components/parser.hpp
+++ b/include/components/parser.hpp
@@ -1,7 +1,5 @@
 #pragma once
 
-#include <stack>
-
 #include "common.hpp"
 #include "errors.hpp"
 
@@ -30,9 +28,8 @@ class parser {
   void codeblock(string&& data, const bar_settings& bar);
   size_t text(string&& data);
 
-  unsigned int parse_color(std::stack<unsigned int>& color_stack, string& value, unsigned int fallback);
-  unsigned int parse_color_string(const string& s, unsigned int fallback = 0);
-  int parse_fontindex(const string& value);
+  unsigned int parse_color(const string& s, unsigned int fallback = 0);
+  int parse_fontindex(const string& s);
   attribute parse_attr(const char attr);
   mousebtn parse_action_btn(const string& data);
   string parse_action_cmd(string&& data);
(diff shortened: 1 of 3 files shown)


patrick96:
To apply module wide settings to the prefix and suffix, could we move the `builder->node(suffix);` before we close all the formatting tags (Line 44)?
And for the prefix, we could duplicate the opening of all the format tags in front of the prefix node.
The prefix and suffix should then overwrite the settings, if they were set. Or is that not how labels and the builder work?

NBonaparte:
We could do that, but it doesn't seem very elegant. Our other option of a shallow reset might be cleaner.

patrick96:
Sorry, for the delay.
While it is not very elegant and duplicates a lot of code, I think having stacks just for that is overkill. And I think that with your latest commit, you agree.
I'll start testing on this now

PR #745, opened 2017-09-12
fix(date): date and time string limited to 63 chars

### LINKED ISSUE #744: Date module is limited to 63 chars
The string `date = AAAAAAAAAABBBBBBBBBBCCCCCCCCCCDDDDDDDDDDEEEEEEEEEEFFFFFFFFFFGGGGGGGGGGHHHHHHHHHHIIIIIIIIII` renders `AAAAAAAAAABBBBBBBBBBCCCCCCCCCCDDDDDDDDDDEEEEEEEEEEFFFFFFFFFFGGG` - this does not seem to be a limitation of `date`, as running `date +AAAAAAAAAABBBBBBBBBBCCCCCCCCCCDDDDDDDDDDEEEEEEEEEEFFFFFFFFFFGGGGGGGGGGHHHHHHHHHHIIIIIIIIII` outputs the exact input string.

This prevents me from using a date string like `%Y%%{F#5c6370}-%%{F-}%m%%{F#5c6370}-%%{F-}%d %H%%{F#5c6370}:%%{F-}%M%%{F#5c6370}:%%{F-}%S`.

patrick96:
Sensible upper limit of 1024 chars since anything of that length will
not be fully rendered on the bar anyways.

fixes #744

### DIFF
--- a/src/modules/date.cpp
+++ b/src/modules/date.cpp
@@ -38,26 +38,36 @@ namespace modules {
     }
   }
 
+  void date_module::set_stream_locale(std::stringstream &stream) {
+    if(!m_bar.locale.empty()) {
+      stream.imbue(std::locale(m_bar.locale.c_str()));
+    }
+  }
+
   bool date_module::update() {
     auto time = std::time(nullptr);
 
     auto date_format = m_toggled ? m_dateformat_alt : m_dateformat;
-    char date_buffer[64]{'\0'};
-    strftime(date_buffer, sizeof(date_buffer), date_format.c_str(), localtime(&time));
+    std::stringstream date_stream; 
+    set_stream_locale(date_stream);
+    date_stream << std::put_time(localtime(&time), date_format.c_str());
+    auto date_string = date_stream.str();
 
     auto time_format = m_toggled ? m_timeformat_alt : m_timeformat;
-    char time_buffer[64]{'\0'};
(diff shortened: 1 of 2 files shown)


patrick96:
Ah crap, I thought boost was already a dependency, but I guess this was back when this was still called lemonbuddy.
I'll fix this in the next few days

src/modules/date.cpp
```diff
@@ -42,22 +42,26 @@ namespace modules {
...
-    strftime(time_buffer, sizeof(time_buffer), time_format.c_str(), localtime(&time));
+    // Clear stream contents
+    datetime_stream.str("");
+    datetime_stream << std::put_time(localtime(&time), time_format.c_str());
+    auto time_string = datetime_stream.str();
 
-    bool date_changed{strncmp(date_buffer, m_date.c_str(), sizeof(date_buffer)) != 0};
-    bool time_changed{strncmp(time_buffer, m_time.c_str(), sizeof(time_buffer)) != 0};
+    bool date_changed{strncmp(date_string.c_str(), m_date.c_str(), sizeof(date_string)) != 0};
+    bool time_changed{strncmp(time_string.c_str(), m_time.c_str(), sizeof(time_string)) != 0};
```
NBonaparte: I think that since we are using strings and not char arrays anymore, we can use the `!=` operator. `sizeof` won't work here because it isn't a char array anyways.

PR #773, opened 2017-09-23
Add .editorconfig

patrick96:
For editors that don't autodetect indent style and size

@NBonaparte can you think of anything else that should be in there?


NBonaparte:
`trim_trailing_whitespace` maybe?

PR #779, opened 2017-09-26
feat(volume): Add pulseaudio backend

NBonaparte:
~~Provides an option to use PulseAudio instead of ALSA for the volume module, which is determined at compile time. If both are available, it will be compiled with PulseAudio by default.~~
Provides a new PulseAudio module, including an adapter.

The only configuration option currently is `sink`, which provides the sink to be used as default (name can be found by running `pacmd list-sinks`). The backend subscribes to any changes in the PulseAudio sinks, so if the sinks change (e.g. headphones are unplugged) and the specified sink disappears, it will automatically revert to the default sink. Similarly, when the specified sink reappears, it will use it instead of the default. It seems like usually there is only one sink, so this option should be optional. The default configuration uses the default sink at all times.

~~Currently, there is no volume cap, so it can go over 100% to an extremely high value, as supported by PulseAudio (not sure whether to limit or leave it to the user).~~
The volume cap is set to `PA_VOLUME_UI_MAX`, which is the recommended maximum value for UIs.

To-Do after merge:
* [x] Add module wiki page
* [ ] Rename `internal/volume` to `internal/alsa`
* [ ] Deprecate `internal/volume`


patrick96:
I have not looked at the code yet, but I think it's great that polybar gets native pulse support. One thing right away: I think it would be better to have a dedicated `pulseaudio` module instead of cramming the functionality into the same module.
This way people can setup either module without having to recompile everything and also the valid syntax of the volume module would not depend on how polybar was compiled. Also it would break all existing configurations that use the built-in volume module on systems that also have pulse installed, once polybar is recompiled.

I will try to take another look at this on the weekend and give some more feedback.
Thanks!

NBonaparte:
Yeah, I think I made a mess of the build configuration there as well. Since the volume module is essentially the alsa module, I think it would be best to rename it to `alsa` and deprecate `volume` (but still link the option `internal/volume` to `internal/alsa` for now).

patrick96:
Yeah, I thought the same thing. And once we push the next major release we can remove the `internal/volume` module.

NBonaparte:
Should I rename the volume module in this PR or do it separately?

patrick96:
I think we leave it like it is in this PR and open a separate one, once this one is merged

NBonaparte:
I've added logging, a volume limit, and cleaned up some redundant volume retrieving.

If an abstract module were to be built, it would only have `get_output()`, `get_format()`, part of `update()`, and part of the constructor. Creating one would probably make things too complicated.

patrick96:
Sorry for the wait on this. This is quite a big PR and I haven't had the time to review it properly. I will try to get in some thourough testing in the next few weeks.

NBonaparte:
Have you had a chance to test this yet?

patrick96:
I'm sorry, I haven't. My exams start in a week and I'm busy studying. I'll try to fit in some testing on my laptop and desktop on the weekend (nothing too thourough though)

patrick96:
@NBonaparte Why do we check for the `pulseaudio` binary in addition to the `libpulse` library?

If we only checked for the library we could change the dependency in the AUR from`pulseaudio` to only `libpulse`.

NBonaparte:
You can have `libpulse` installed without having `pulseaudio`. For instance, chromium only requires the former, and I tried to run pulseaudio a long time ago before realizing I actually didn’t have the binary installed.

patrick96:
Ah, I see. But the binary is not necessary for polybar to run?

NBonaparte:
Yeah I suppose. It would be up to the user to ensure that they actually have the binary installed.

PR #785, opened 2017-09-30
Add minimum width to CPU percentage

roobre:
It's quite obscure where to find that you can set a minimum width for tokens, and CPU usage will oscillate between 0-9 and 10+ quite frequently. I think that adding this to the example config will help users to find out that this feature exists, and avoid this annoying default.

### DIFF
--- a/doc/config.cmake
+++ b/doc/config.cmake
@@ -193,7 +193,7 @@ interval = 2
 format-prefix = " "
 format-prefix-foreground = ${colors.foreground-alt}
 format-underline = #f90000
-label = %percentage%%
+label = %percentage:2%%
 
 [module/memory]
 type = internal/memory


patrick96:
Looks good. Thanks!

NBonaparte:
Wouldn't having 3 as the minimum width be better?
Also, we should apply minimum widths to other relevant tokens (memory %, temperature, etc.) to make everything consistent.

patrick96:
A minimum width of 3 would cover all cases but a cpu usage of 100% doesn't happen that often.
But in the end, this is just an example config and only here for new users to see what's possible, so either way doesn't really have that much of an impact. So, if you'd rather use 3, we'll do that.
And as for the other labels, I think we can do the same, as long as they have an upper limit of digits

roobre:
@NBonaparte Given how CPU usage is calculated, values of 0 and 100 are almost never (and, depending on the implementation, never) reached.

NBonaparte:
It all depends on one's workload and CPU. Personally, I've observed that the bar calculates usage as 0% while idle and 100% when compiling using all threads. IMO it's better to consider all possibilities rather than just the general case.

Issue #596, opened 2017-05-26
Select time zone for date module


nyarly:
It'd be extremely handy to be able to select which time zone a date module displays. Even just "utc = true" would be useful.

jaagr:
See #557

patrick96:
I reopened this because we will now start tracking this as a feature request for a new parameter in the date module and ignore the script workaround

Lomadriel:
I tried something with `setenv` to set the `TZ` environment variable as suggested in #788 by @NBonaparte. But there is a problem.

If we suppose the following config:
```
[bar/example]
…
modules-left = date_france date_local
…
```
The environment modification applies to all threads, so if there is two `internal/date` modules, as in the example above, the timezone may be applied to the two modules rather than just the `date_france` module.

One solution could be to use a "global clock", each module will get the time from a blocking method that will use `setenv`. Since this is a blocking call the request from each module will be sequential and it will resolve the thread concurrency problem.
What do you think of this solution?

patrick96:
This feature is currently blocked because this seems to be impossible to do safely. See [this comment](https://github.com/polybar/polybar/pull/3160#issuecomment-2323348167) for more details.

TL;DR using `setenv` is inherently unsafe and there is currently no other way to format a point in time for a specific timezone.
There is a C++20 feature that may allow us to do this safely, but C++20 is still experimental in GCC

PR #788, opened 2017-10-04
Timezone for Date Module (Issue #596)

braydenjw:
This pull request addresses [issue 596](https://github.com/jaagr/polybar/issues/596).

It allows users to specify a timezone under the date module (e.g. `timezone=America/New_York`).  Using [Google's CCTZ libray](https://github.com/google/cctz), it prints the current time in the specified timezone.  If no timezone is specified, or the specified timezone was not found, the system timezone is assumed.

### DIFF
--- a/include/modules/date.hpp
+++ b/include/modules/date.hpp
@@ -5,7 +5,6 @@
 
 #include <iostream>
 #include <iomanip>
-#include <ctime>
 
 POLYBAR_NS
 
@@ -33,6 +32,8 @@ namespace modules {
     string m_dateformat_alt;
     string m_timeformat;
     string m_timeformat_alt;
+
+    string m_timezone;
 
     string m_date;
     string m_time;
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -7,6 +7,13 @@
 list(APPEND dirs ${CMAKE_CURRENT_LIST_DIR}/concurrentqueue/include)
 
 # }}}
+# Library: cctz {{{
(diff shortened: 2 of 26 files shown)


braydenjw:
This is actually my first time contributing to an Open Source project. Let me know what I can do to move this along.  I'm open to making any changes needed.

NBonaparte:
I think using cctz is overkill, as it adds 5000 lines of code just for timezones. If we were to use it, it would have to be added as a submodule, rather than having all the files in our repo. 

A much simpler option would be to use `setenv()` to set the `TZ` environment variable, and then calling `tzset()` which will then allow `localtime()` to calculate it according to the timezone.

braydenjw:
I can definitely add it as a sub-module instead, my only problem with that is the build could fail if there were breaking changes upstream.  

I considered just setting the environment variable, my worry was what the behaviour would be when adding multiple date modules, each with a different timezone. 

Regardless, I'd like to make a few improvement to what I have, namely moving some stuff out of the upload function and fixing the g++ build (looks like just a cmake fix).

patrick96:
I'm not a fan of adding a new dependency just for the timezone feature.

I don't think setting the environment variable inside the update method is that much of a performance hog.

braydenjw:
Sure! I'll remove it and use setenv instead, probably check for the existence of the timezone in _/usr/share/zoneinfo_ to ensure the entered timezone is valid.  

I might do a performance analysis against using CCTZ to see what we're dealing with.

PR #801, opened 2017-10-12
Add Fedora build helper

nathanchere:
Since polybar still isn't accepted into Fedora repositories, this adds a one-liner for installation on Fedora-based systems.


patrick96:
Thanks for the effort! But I don't think that this needs to be part of the README, this information is already available on the [Compiling Wiki Page](https://github.com/jaagr/polybar/wiki/Compiling).

PR #804, opened 2017-10-13
Optional display of temperature in Fahrenheit units

chrissnell:
This PR creates a new boolean option, `temperature-fahrenheit`, for the `internal/temperature` module.  When set to true, the temperature will be displayed as degrees Fahrenheit.

This is my first time writing C++.  Please let me know if there's a way to make this better.

### DIFF
--- a/doc/config.cmake
+++ b/doc/config.cmake
@@ -316,7 +316,6 @@ animation-charging-framerate = 750
 type = internal/temperature
 thermal-zone = 0
 warn-temperature = 60
-;temperature-farhrenheit = true
 
 format = <ramp> <label>
 format-underline = #f50a4d
--- a/doc/config.cmake
+++ b/doc/config.cmake
@@ -316,6 +316,7 @@ animation-charging-framerate = 750
 type = internal/temperature
 thermal-zone = 0
 warn-temperature = 60
+;temperature-farhrenheit = true
 
 format = <ramp> <label>
 format-underline = #f50a4d
--- a/src/modules/temperature.cpp
+++ b/src/modules/temperature.cpp
@@ -19,7 +19,6 @@ namespace modules {
     m_path = m_conf.get(name(), "hwmon-path", ""s);
     m_tempwarn = m_conf.get(name(), "warn-temperature", 80);
     m_interval = m_conf.get<decltype(m_interval)>(name(), "interval", 1s);
-    m_temperature_fahrenheit = m_conf.get(name(), "temperature-fahrenheit", m_temperature_fahrenheit);
 
     if (m_path.empty()) {
       m_path = string_util::replace(PATH_TEMPERATURE_INFO, "%zone%", to_string(m_zone));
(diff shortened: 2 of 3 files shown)


src/modules/temperature.cpp
```diff
@@ -44,10 +46,15 @@ namespace modules {
   bool temperature_module::update() {
     m_temp = std::atoi(file_util::contents(m_path).c_str()) / 1000.0f + 0.5f;
     m_perc = math_util::cap(math_util::percentage(m_temp, 0, m_tempwarn), 0, 100);
+    int temp_f = floor((((9.0/5.0) * m_temp) + 32) + 0.5);
```
patrick96: Why the `+0.5` here?

patrick96:
Thank you very much for your contribution! Your code looks good.
Fahrenheit is definitely something that the `temperature` module should support. Though it would allow for greater extensibility and would also look cleaner if there were separate tokens for fahrenheit and celsius instead of having a new key in the config.
Is this something you have time and would like to implement?

src/modules/temperature.cpp
NBonaparte: I think it's for rounding purposes. Usually `floor()` isn't used after that though, because the value is already rounded.

chrissnell:
Accidentally screwed up my commit log a little bit on this one but the changes are correct.  Perhaps you can squash the PR that I did within chrissnell/polybar

doc/config.cmake
```diff
@@ -316,6 +316,7 @@ animation-charging-framerate = 750
 type = internal/temperature
 thermal-zone = 0
 warn-temperature = 60
+;temperature-farhrenheit = true
```
patrick96: Can you remove that, since this option doesn't exist anymore

chrissnell:
@patrick96 fixed

src/modules/temperature.cpp
```diff
@@ -43,10 +44,13 @@ namespace modules {
 
   bool temperature_module::update() {
     m_temp = std::atoi(file_util::contents(m_path).c_str()) / 1000.0f + 0.5f;
+    int m_temp_f = floor(((9.0/5.0) * (m_temp + 32)) + 0.5);
```
patrick96: Seems like the formula is a little off, it should be
`1.8 * temp + 32`

chrissnell: Not sure how that's any different.  We do need the floor() and the + 0.5 to round to a whole number.

patrick96: Sorry, I should have been clearer.
You have the `m_temp + 32` in brackets so you multiply that whole thing by 1.8 instead of just `m_temp`.
The rounding part is fine

chrissnell: Doh.  Fixing.

patrick96:
Thanks!

CumpsD:
I was wondering if there is any option to not append the degrees sign? So I can use the Celsius sign from Material Icons :)

patrick96:
@CumpsD set `units = false` in the module.

CumpsD:
@patrick96 thanks :) https://github.com/CumpsD/dotfiles/commit/108c5e61942acfc1674077696852020ed25bb7ad

PR #807, opened 2017-10-15
Fixed a compilation error on Fedora Rawhide

pierg75:
I've built rpm packages for fedora and I run into an error on Fedora Rawhide.

`...use of undeclared identifier 'pow10'...`

Looking around this seems a quick fix with pow(10, ...).
It compiles now with these fixes.


patrick96:
Looks good. Thanks!
Seems like `pow10` is a GNU extension and will not properly compile on all systems. This is probably the easiest fix

Issue #810, opened 2017-10-20
github module without internet connection


Kokan:
If the github module cannot connect to github to fetch the number of notification it disappears from the bar, and do not come back until polybar restarted (or config reload in case of `--reload` flag).

Tested with the current master.

Issue #754, opened 2017-09-15
internal/github module causes crash after some period of time


apetresc:
I have the following module configured in my polybar config:
```
[module/github]
type = internal/github
token = xxxxxxxxxx
empty-notifications = false
format = <label>
label = %{A1:xdg-open https\://github.com/notifications/:}  %notifications% %{A}
format-underline = #3f4448
```
(Yes, my polybar is compiled with `+curl`)

If I add this module to a bar, everything appears to work fine, I can see the entry, it refreshes every so often, etc. However, randomly and at uneven intervals, but always within ~24 hours, polybar will crash without any sort of output spilled to stderr or stdout.

Even stranger, it doesn't even happen at the same time if I have multiple bars (one for each monitor). Maybe after 3 hours one monitor will lose polybar, and then 10 hours later, the other monitor will lose it.

Removing `github` from the bar makes everything return to normal.

Any hints on how I can start diagnosing this issue? Again, I don't see anything on stdout/stderr. Should I try to get a core dump or something? Has anyone encountered this before?

patrick96:
The first step in debugging this would probably be running polybar with the `-l info` and posting the output of that.
Also, when polybar terminates, what is its return code?

apetresc:
Cool, thanks – I will grab both of those pieces of info for you right away. It just might take a few hours for the crash to happen :)

apetresc:
Okay, both have crashed now. In both cases, there was nothing remarkable in the output, even with `l info` (just a bunch of 'redrawing bar window`s, etc), and the exit code in both cases was 141.

patrick96:
Hmm, the return code seems to suggest that polybar terminated due to SIGPIPE.
Can you confirm for me that the github module is actually the culprit by removing all other modules from the `modules-*` lists.

While you're at it, you can start debugging with `gdb`. You'll have to recompile polybar with debug symbols (add `-DCMAKE_BUILD_TYPE=Debug` to cmake), then run the bar and attach to the process using `sudo gdb -p PID -ex cont | tee /path/to/logfile`. Once the crash occurs, gdb should show a message along the lines of
```
Thread 1 "polybar" received signal SIGPIPE, Broken pipe
```
and the bar will stop updating.
Once that happens, run `info stack` in gdb and then post the contents of the log file here.
If you don't know how to do some of those things, please tell me and I'll write up a quick guide

apetresc:
No problem - yes, I know how to do all those things :) I will give it a
[...]

PR #811, opened 2017-10-20
Githubmodule noconnection

Kokan:
This PR aims to fix the #810 and also contains small refactor for the github module.

### DIFF
--- a/src/modules/github.cpp
+++ b/src/modules/github.cpp
@@ -59,7 +59,13 @@ namespace modules {
   }
 
   int github_module::get_number_of_notification() {
-    string content = request();
+    string content;
+    try {
+      content = request();
+    } catch (application_error& e) {
+      m_log.warn("%s: cannot complete the request to github: %s", name(), e.what());
+      return -1;
+    }
 
     size_t pos{0};
     size_t notifications{0};
@@ -72,7 +78,7 @@ namespace modules {
   }
 
   void github_module::update_label(const int notifications) {
-    if (notifications || m_empty_notifications) {
+    if (0 != notifications || m_empty_notifications) {
       m_label->reset_tokens();
       m_label->replace_token("%notifications%", to_string(notifications));
     } else {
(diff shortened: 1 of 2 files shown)


patrick96:
Thank you very much!
I'll have a look at this on the weekend.

Seems like maybe this is the root cause of #754

Kokan:
Sure, I've added a log message with the exception.

Kokan:
@patrick96 could you please check it ? thanks.

patrick96:
Sorry, for the delay. I'll try to have a look tomorrow. If I don't, please remind me again in a few days.

src/modules/github.cpp
```diff
@@ -25,21 +25,25 @@ namespace modules {
     if (m_formatter->has(TAG_LABEL)) {
       m_label = load_optional_label(m_conf, name(), TAG_LABEL, "Notifications: %notifications%");
 
-      if (m_empty_notifications) {
-        m_label->replace_token("%notifications%", "0");
-      } else {
-        m_label->clear();
-      }
+      update_label(-1);
```
patrick96: I think we should keep the behaviour that the module is initialized with 0

Kokan: Actually it is confusing for me to see no notification, when actually no information is present.
So I preferred to display other value then 0. 
I am planning to introduce the `label-offline` which would be the best option so I changed back to 0 for now :)

patrick96:
Thank you very much!
Would really appreciate it, if you worked on the `label-offline` feature.

Btw. how are you committing your changes? The commit list only shows two commits but you have certainly more than that. Do you append old commits and force push?

Kokan:
Yes, I used fixups for the commits, and force push. Github tracks those changes, and it is better for the history specially in case of rebase instead of merge :)

patrick96:
Yes definitely helps keep the commit history clean. But it also makes it harder to see what was actually changed when something new is commited because you have to review the old commit again too.

Issue #824, opened 2017-10-30
Segfault when mousewheeling on i3 module


daf:
Hi, I'm experiencing a segfault in polybar 3.0.5 when using the mouse wheel on top of the i3 module.

Gdb gives the following backtrace:
```
Thread 5 "polybar" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffeccb4700 (LWP 16153)]
std::__shared_ptr<i3ipc::workspace_t, (__gnu_cxx::_Lock_policy)2>::__shared_ptr (this=<synthetic pointer>) at /usr/include/c++/5/bits/shared_ptr_base.h:923
923	      __shared_ptr(const __shared_ptr&) noexcept = default;
(gdb) bt
#0  std::__shared_ptr<i3ipc::workspace_t, (__gnu_cxx::_Lock_policy)2>::__shared_ptr (this=<synthetic pointer>) at /usr/include/c++/5/bits/shared_ptr_base.h:923
#1  std::shared_ptr<i3ipc::workspace_t>::shared_ptr (this=<synthetic pointer>) at /usr/include/c++/5/bits/shared_ptr.h:107
#2  polybar::v3_0_5::modules::i3_module::input(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&&) (this=0x8bb780, 
    cmd=<optimized out>) at /home/daf/ext/polybar/src/modules/i3.cpp:224
#3  0x00000000005cb159 in polybar::v3_0_5::controller::process_inputdata (this=this@entry=0x8b70a0) at /home/daf/ext/polybar/src/components/controller.cpp:410
#4  0x00000000005d09b5 in polybar::v3_0_5::controller::process_eventqueue (this=<optimized out>) at /home/daf/ext/polybar/src/components/controller.cpp:384
#5  0x00007ffff6138c80 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff594b6ba in start_thread (arg=0x7fffeccb4700) at pthread_create.c:333
#7  0x00007ffff56813dd in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
```

Seems something to do with trying to check workspace visibility:
https://github.com/jaagr/polybar/blob/a3e9bb9847271765e6e63418ca5164a576bcb42b/src/modules/i3.cpp#L223-L224

patrick96:
[...]

PR #826, opened 2017-11-01
fix(i3): Check whether current ws is found

TonCherAmi:
Fixes dereference of `end()` iterator in case current workspace is not found (#824).

I'm not sure whether this should be logged or not, so I'd like to hear your opinion on that.


NBonaparte:
IMO it should be logged as a warning so we could keep track of it better.

TonCherAmi:
Hope that's sufficient.

patrick96:
Looks good. Thank you very much!

PR #831, opened 2017-11-05
fix(renderer): Only draw text bg if the color differs from the bar bg color

### LINKED ISSUE #759: Modules gradient issue
My bar have gradient with two colors. When I used version _3.0.5_ bar looked like this:
Old

But after aur-git update to version _3.0.5-44-ga682d2a_ I got this:
New

Modules has no gradients now. I read the wiki, but there is nothing about it and I tried some combinations with _format-background_ etc. Any help, please? Should I use one color for the bar?

P. S. Sorry about my English, it isn't my primary language.

TonCherAmi:
Fixes #759 by only drawing text background when its color is different from the background color of the bar itself. Original check was made invalid by [this change](https://github.com/jaagr/polybar/commit/0bd8f1f69a8dfdb2f2800a6612c8acb7e6c86ed2#diff-543075e5874c51140b59304a24b67c78L208).

There is one small issue though: if the text background color is explicitly set equal to `background-0` it will be ignored.

### DIFF
--- a/src/components/renderer.cpp
+++ b/src/components/renderer.cpp
@@ -602,7 +602,11 @@ void renderer::draw_text(const string& contents) {
   block.y_advance = &m_blocks[m_align].y;
   block.bg_rect = cairo::rect{0.0, 0.0, 0.0, 0.0};
 
-  if (m_bg) {
+  // Only draw text background if the color differs from
+  // the background color of the bar itself
+  // Note: this means that if the user explicitly set text
+  // background color equal to background-0 it will be ignored
+  if (m_bg != m_bar.background) {
     block.bg = m_bg;
     block.bg_operator = m_comp_bg;
     block.bg_rect.x = m_rect.x;


patrick96:
I don't think there is any way to solve this without a lot of control variables or where explicitly setting the background color to the bar background doesn't break the expected behaviour with gradients.
So I think this is the best solution for this problem.
Can you please add a comment explaining the reason for this and its pitfalls in the code.
I will do some more testing on this.

TonCherAmi:
@patrick96 yeah, I figured as much as well. Regarding the comment, is it fine if I simply push to this branch or am I supposed to do a rebase first since it is behind master?

patrick96:
Do a push without rebasing and if this page shows a conflict, do a rebase

patrick96:
Thank you very much!
I have also added a note on the [Wiki](https://github.com/jaagr/polybar/wiki/Known-Issues) about this for future reference

PR #861, opened 2017-11-25
feat(mpd): Add support for icon-consume

Shrvi:
This PR adds support for _icon-consume_ to the mpd module.
It works exactly like _icon-random_ and the alike.


patrick96:
Looks good and works as expected.

Will merge, once #863 is merged

PR #863, opened 2017-11-26
Release 3.1.0

patrick96:
This is a first draft for the polybar 3.1.0 release.

**Things to do before merging:**

- [x] Write changelog and add it to the github release tag
- [x] Check if all new features are documented on the wiki

**Things to do after this is merged:**

- [ ] Ensure AUR packages build properly and have proper version tags
- [ ] Push updated PKGBUILDs (and .SRCINFO)
- [ ] Inform other package maintainers of new version (https://repology.org/metapackage/polybar/versions)
- [ ]  Update Void Linux package
- [ ] Remove jsoncpp issue from "Known Issues" page

Please tell me, if you think there are other things that need to be done for this release.

Closes #856

PR #868, opened 2017-11-29
fix(mpd): get format-online options for format-{playing,paused,stopped}

NBonaparte:
Fixes the broken `format-online-*` options mentioned in #863 by retrieving them individually and applying them to the other formats if they exist.


patrick96:
That was fast! Everything seems to work now.

PR #870, opened 2017-11-30
feat(build): help target to list all targets

Kokan:
Just a simple `help` target for the `Makefile` which list all of the targets with a single line `awk`.
So sometimes I could avoid opening the `Makefile` only to know the targets :smile:


patrick96:
I didn't even know, we had a makefile ^^
Looks good though. Thanks!

PR #875, opened 2017-12-02
I3: reset mode

tux-mind:
Reset `%mode%` token after switching back to default mode.


patrick96:
I don't have that much experience with the i3 module, so I'm not really sure what this is supposed to do. Can you please explain what behaviour this fixes or how it improves the i3 module?
Also, I have noticed that when the bar first starts the `%mode%` token is not properly replaced and is put directly onto the bar.

patrick96:
Closing due to inactivity

PR #877, opened 2017-12-03
fix(AUR): Remove failing call to version.sh

patrick96:
version.sh is supposed to set the `GIT_TAG` macro in `version.hpp` to the
output of `git describe ...`, however it uses relative paths and can't
find `include/version.hpp`

`include/CMakeLists.txt` already has the same functionality so this is
not needed


patrick96:
Updated on the AUR

PR #882, opened 2017-12-03
feat(xworkspaces): List occupied workspaces

### LINKED ISSUE #874: xworkspace Improvements
I recently switched from xmobar/trayer to polybar and so far, I like it. The only thing I am missing is support for "occupied workspaces" in the xworkspaces module. Given the choice between complaining and demanding or rolling up the sleeves and improve things myself, I opt for the later. 

I shortly looked into the existing code and I have a vague feeling that there is no technical blocker preventing this. 

My plan is to go over the complete list of windows and lookup the given workspace it is assigned to. Since this already asks xproperties of every window, it could also lead to a solution for #448. It also seems that I could build upon the work for the urgency-handling (#449). The urgency-handling seems to implement all building blocks I would need.

Before I go ahead, I want to make sure if this plan is sound or if I am missing something.

### LINKED ISSUE #1033: xworkspaces segfault
This segfault is pretty rare, and I'm not entirely sure what triggers it. Managed to catch it with gdb by running polybar under the debugger all day until it crashed:

```
Thread 13 "polybar" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffcbfff700 (LWP 1106)]
polybar::v3_1_0_56_gbd022c9_git::builder::node (this=0x555555a5c8d0,
    label=<error reading variable: Cannot access memory at address 0x7361625f70614d67>, add_space=false)
    at /home/jon/dev/others/polybar/src/components/builder.cpp:244
244       if (label->m_padding.right > 0) {
(gdb) bt
#0  polybar::v3_1_0_56_gbd022c9_git::builder::node (this=0x555555a5c8d0,
    label=<error reading variable: Cannot access memory at address 0x7361625f70614d67>, add_space=false)
    at /home/jon/dev/others/polybar/src/components/builder.cpp:244
#1  0x00005555557227ce in polybar::v3_1_0_56_gbd022c9_git::modules::xworkspaces_module::build (this=0x555555a5c490,
    builder=0x555555a5c8d0, tag=...) at /home/jon/dev/others/polybar/src/modules/xworkspaces.cpp:300
#2  0x000055555572b7b9 in polybar::v3_1_0_56_gbd022c9_git::modules::module<polybar::v3_1_0_56_gbd022c9_git::modules::xworkspaces_module>::get_output[abi:cxx11]() (this=0x555555a5c490)
    at /home/jon/dev/others/polybar/include/modules/meta/base.inl:159
#3  0x00005555557222bd in polybar::v3_1_0_56_gbd022c9_git::modules::xworkspaces_module::get_output[abi:cxx11]() (
    this=0x555555a5c490) at /home/jon/dev/others/polybar/src/modules/xworkspaces.cpp:264
#4  0x000055555572b1d0 in polybar::v3_1_0_56_gbd022c9_git::modules::module<polybar::v3_1_0_56_gbd022c9_git::modules::xworkspaces_module>::contents[abi:cxx11]() (this=0x555555a5c490)
    at /home/jon/dev/others/polybar/include/modules/meta/base.inl:81
#5  0x000055555567cc76 in polybar::v3_1_0_56_gbd022c9_git::controller::process_update (this=<optimized out>,
    force=false) at /home/jon/dev/others/polybar/src/components/controller.cpp:465
#6  0x0000555555679c1a in polybar::v3_1_0_56_gbd022c9_git::controller::process_eventqueue (this=0x555555a54500)
    at /home/jon/dev/others/polybar/src/components/controller.cpp:357
#7  0x00007ffff5985b9f in std::execute_native_thread_routine (__p=0x555555a5df90)
    at /build/gcc/src/gcc/libstdc++-v3/src/c++11/thread.cc:83
#8  0x00007ffff515208c in start_thread () from /usr/lib/libpthread.so.0
#9  0x00007ffff4e89e7f in clone () from /usr/lib/libc.so.6
```

### LINKED ISSUE #1444: segfault when building the xworkspace-nodes from desktop-labels
I sometimes run into a segfault during my work on #882 .

**Expected behavior:**

I expect no segfault :-)

**Actual behavior:**

When running the bar under gdb until the segfault, I get the following backtrace:

```
#0  std::__shared_ptr<polybar::v3_2_0_52_g1550152_git::drawtypes::label, (__gnu_cxx::_Lock_policy)2>::get (this=0x10)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/bits/shared_ptr_base.h:1060
#1  <function called from gdb>
#2  std::__shared_ptr<polybar::v3_2_0_52_g1550152_git::drawtypes::label, (__gnu_cxx::_Lock_policy)2>::get (this=0x10)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/bits/shared_ptr_base.h:1060
Python Exception <class 'gdb.error'> There is no member named _M_dataplus.: 
#3  0x00000000009114a7 in polybar::v3_2_0_52_g1550152_git::modules::xworkspaces_module::build (this=0xcf5760, builder=0xcf5bf0, tag=)
    at /home/mviehweger/Code/opensource/polybar/src/modules/xworkspaces.cpp:310
#4  0x000000000091a1f5 in polybar::v3_2_0_52_g1550152_git::modules::module<polybar::v3_2_0_52_g1550152_git::modules::xworkspaces_module>::get_output[abi:cxx11]() (this=0xcf5760) at /home/mviehweger/Code/opensource/polybar/include/modules/meta/base.inl:160
#5  0x0000000000910e95 in polybar::v3_2_0_52_g1550152_git::modules::xworkspaces_module::get_output[abi:cxx11]() (this=0xcf5760)
    at /home/mviehweger/Code/opensource/polybar/src/modules/xworkspaces.cpp:280
#6  0x0000000000919a0d in polybar::v3_2_0_52_g1550152_git::modules::module<polybar::v3_2_0_52_g1550152_git::modules::xworkspaces_module>::contents[abi:cxx11]() (this=0xcf5760) at /home/mviehweger/Code/opensource/polybar/include/modules/meta/base.inl:82
#7  0x00000000007bf711 in polybar::v3_2_0_52_g1550152_git::controller::process_update (this=0xcf4b70, force=false)
    at /home/mviehweger/Code/opensource/polybar/src/components/controller.cpp:481
#8  0x00000000007bd22b in polybar::v3_2_0_52_g1550152_git::controller::process_eventqueue (this=0xcf4b70)
    at /home/mviehweger/Code/opensource/polybar/src/components/controller.cpp:394
#9  0x00000000007d36c9 in std::_Mem_fn_base<void (polybar::v3_2_0_52_g1550152_git::controller::*)(), true>::operator()<, void>(polybar::v3_2_0_52_g1550152_git::controller*) const (this=0xd30ca0, __object=0xcf4b70) at /usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/functional:600
#10 0x00000000007d3632 in std::_Bind_simple<std::_Mem_fn<void (polybar::v3_2_0_52_g1550152_git::controller::*)()> (polybar::v3_2_0_52_g1550152_git::controller*)>::_M_invoke<0ul>(std::_Index_tuple<0ul>) (this=0xd30c98) at /usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/functional:1530
#11 0x00000000007d35e5 in std::_Bind_simple<std::_Mem_fn<void (polybar::v3_2_0_52_g1550152_git::controller::*)()> (polybar::v3_2_0_52_g1550152_git::controller*)>::operator()() (this=0xd30c98) at /usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/functional:1520
#12 0x00000000007d33a9 in std::thread::_Impl<std::_Bind_simple<std::_Mem_fn<void (polybar::v3_2_0_52_g1550152_git::controller::*)()> (polybar::v3_2_0_52_g1550152_git::controller*)> >::_M_run() (this=0xd30c80) at /usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/thread:115
#13 0x00007ffff54b5c80 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#14 0x00007ffff4cc86ba in start_thread (arg=0x7fffe17fa700) at pthread_create.c:333
#15 0x00007ffff49fe41d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
```

## Environment:
* WM: xmonad
* Output of `polybar -vvv`:

polybar 3.2.0-52-g1550152-git

Features: +alsa +curl -i3 +mpd +network(wireless-tools) +pulseaudio +xkeyboard

X extensions: +randr (-monitors) -render -damage -sync -composite +xkb +xrm -xcursor

Build type: Debug
Compiler: /usr/bin/clang++
Compiler flags:  -Wall -Wextra -Werror -O2 -pedantic -pedantic-errors -Wno-error=parentheses-equality -Wno-zero-length-array
Linker flags:

The master of polybar that I am building on top of is https://github.com/jaagr/polybar/commit/2fba443.

## Attempted debugging

PR #882 does not change the source-code from which the segfault originates, but it shifts its location a few lines down. See https://github.com/jaagr/polybar/pull/882/files#diff-e89fe726ad41f4ae273dacfca026a7ddR310 for the code mentioned here.

From the above backtrace, I poked around a little to know more. In order to establish the context in which this on segfaults, I did the following:

```
(gdb) f 3
Python Exception <class 'gdb.error'> There is no member named _M_dataplus.: 
#3  0x00000000009114a7 in polybar::v3_2_0_52_g1550152_git::modules::xworkspaces_module::build (this=0xcf5760, builder=0xcf5bf0, tag=)
    at /home/mviehweger/Code/opensource/polybar/src/modules/xworkspaces.cpp:310
310	        if (desktop->label.get()) {
(gdb) p m_index 
$2 = 0
(gdb) p m_viewports 
$3 = std::vector of length 1, capacity 1 = {std::unique_ptr<polybar::v3_2_0_52_g1550152_git::modules::viewport> containing 0xcef500}
(gdb) p m_viewports[0]->desktops 
$4 = std::vector of length 9, capacity 16 = {std::unique_ptr<polybar::v3_2_0_52_g1550152_git::modules::desktop> containing 0xd69690, 
  std::unique_ptr<polybar::v3_2_0_52_g1550152_git::modules::desktop> containing 0xd03380, 
  std::unique_ptr<polybar::v3_2_0_52_g1550152_git::modules::desktop> containing 0xd41e50, 
  std::unique_ptr<polybar::v3_2_0_52_g1550152_git::modules::desktop> containing 0xd37830, 
  std::unique_ptr<polybar::v3_2_0_52_g1550152_git::modules::desktop> containing 0xd3ac80, 
  std::unique_ptr<polybar::v3_2_0_52_g1550152_git::modules::desktop> containing 0xd03a30, 
  std::unique_ptr<polybar::v3_2_0_52_g1550152_git::modules::desktop> containing 0xd3ad70, 
  std::unique_ptr<polybar::v3_2_0_52_g1550152_git::modules::desktop> containing 0xd38460, 
  std::unique_ptr<polybar::v3_2_0_52_g1550152_git::modules::desktop> containing 0xd4b2b0}
(gdb) p desktop
$5 = std::unique_ptr<polybar::v3_2_0_52_g1550152_git::modules::desktop> containing 0x0
```
## Related Problems

This time, the segfault originates from `desktop->label.get()`, other times, it manages to get a few lines further until `desktop->label` is called/passed to the builder. 

I assume by now, that xmonad changes the desktop-properties and that this somehow invalidates/corrupts the data-structure. My proposed solution is to add another check against invalid desktop-pointers.

Keep in mind, I have little knowledge about X11-properties, C++(-datastructures) and the interaction between the moving parts here.

kronn:
My plan is grand and all-encompassing: list all occupied workspaces.

So far, I list all workspaces not currently visible on another workspace. I would be very grateful for input on how to resolve this to a usable state... 

This is supposed to fix #874 

TEAMEDIT:
Fixes #1444 
Fixes #1033

### DIFF
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <bitset>
+#include <mutex>
 
 #include "components/config.hpp"
 #include "components/types.hpp"
@@ -108,6 +109,7 @@ namespace modules {
     size_t m_index{0};
 
     event_timer m_timer{0L, 25L};
+    std::mutex m_mutex;
   };
 }
 
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -67,6 +67,7 @@ namespace modules {
     void rebuild_desktops();
     void rebuild_desktop_states();
     void set_desktop_urgent(xcb_window_t window);
+    void set_desktop_occupied(xcb_window_t window);
 
     bool input(string&& cmd);
     vector<string> get_desktop_names();
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -107,7 +107,7 @@ namespace modules {
     size_t m_index{0};
 
     event_timer m_timer{0L, 25L};
-    std::mutex m_mutex;
+    mutable std::mutex m_mutex;
   };
 }
 
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -65,11 +65,9 @@ namespace modules {
     void handle(const evt::property_notify& evt);
 
     void rebuild_clientlist();
-    void recount_clients_on_desktops();
     void rebuild_desktops();
     void rebuild_desktop_states();
     void set_desktop_urgent(xcb_window_t window);
-    void set_desktop_occupied(xcb_window_t window);
 
     bool input(string&& cmd);
     vector<string> get_desktop_names();
@@ -96,9 +94,9 @@ namespace modules {
     vector<string> m_desktop_names;
     unsigned int m_current_desktop;
     string m_current_desktop_name;
-    vector<unsigned int> m_desktop_client_count;
+    vector<bool> m_desktop_occupied;
 
-    vector<pair<xcb_window_t, unsigned int>> m_clientlist;
+    vector<xcb_window_t> m_clientlist;
     vector<unique_ptr<viewport>> m_viewports;
     map<desktop_state, label_t> m_labels;
     label_t m_monitorlabel;
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -67,7 +67,6 @@ namespace modules {
     void rebuild_desktops();
     void rebuild_desktop_states();
     void set_desktop_urgent(xcb_window_t window);
-    void set_desktop_occupied(xcb_window_t window);
 
     bool input(string&& cmd);
     vector<string> get_desktop_names();
@@ -94,6 +93,7 @@ namespace modules {
     vector<string> m_desktop_names;
     unsigned int m_current_desktop;
     string m_current_desktop_name;
+    vector<int> m_occupied_desktops;
 
     vector<xcb_window_t> m_clientlist;
     vector<unique_ptr<viewport>> m_viewports;
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -94,7 +94,7 @@ namespace modules {
     vector<string> m_desktop_names;
     unsigned int m_current_desktop;
     string m_current_desktop_name;
-    vector<unsigned int> m_occupied_desktops;
+    vector<unsigned int> m_desktop_client_count;
 
     vector<pair<xcb_window_t, unsigned int>> m_clientlist;
     vector<unique_ptr<viewport>> m_viewports;
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -107,7 +107,7 @@ namespace modules {
     size_t m_index{0};
 
     event_timer m_timer{0L, 25L};
-    mutable std::mutex m_mutex;
+    mutable std::mutex m_event_mutex; // protect m_viewports while building the output
   };
 }
 
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -97,6 +97,9 @@ namespace modules {
     unsigned int m_current_desktop;
     string m_current_desktop_name;
 
+    /**
+     * Maps an xcb window to its desktop number
+     */
     map<xcb_window_t, unsigned int> m_clients;
     vector<unique_ptr<viewport>> m_viewports;
     map<desktop_state, label_t> m_labels;
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -2,6 +2,7 @@
 
 #include <bitset>
 #include <mutex>
+#include <set>
 
 #include "components/config.hpp"
 #include "components/types.hpp"
@@ -70,9 +71,10 @@ namespace modules {
     void set_desktop_urgent(xcb_window_t window);
 
     bool input(string&& cmd);
-    vector<string> get_desktop_names();
 
    private:
+    static vector<string> get_desktop_names();
+
     static constexpr const char* DEFAULT_ICON{"icon-default"};
     static constexpr const char* DEFAULT_LABEL_STATE{"%icon% %name%"};
     static constexpr const char* DEFAULT_LABEL_MONITOR{"%name%"};
@@ -94,9 +96,8 @@ namespace modules {
     vector<string> m_desktop_names;
     unsigned int m_current_desktop;
     string m_current_desktop_name;
-    vector<bool> m_desktop_occupied;
 
-    vector<xcb_window_t> m_clientlist;
+    map<xcb_window_t, unsigned int> m_clients;
     vector<unique_ptr<viewport>> m_viewports;
     map<desktop_state, label_t> m_labels;
     label_t m_monitorlabel;
@@ -106,9 +107,12 @@ namespace modules {
     bool m_scroll{true};
     size_t m_index{0};
 
+    // The following mutex is here to protect the data of this modules.
+    // This can't be achieved using m_buildlock since we "CRTP override" get_output().
+    mutable mutex m_workspace_mutex;
+
     event_timer m_timer{0L, 25L};
-    mutable std::mutex m_event_mutex; // protect m_viewports while building the output
   };
-}
+}  // namespace modules
 
 POLYBAR_NS_END
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -64,6 +64,7 @@ namespace modules {
     void handle(const evt::property_notify& evt);
 
     void rebuild_clientlist();
+    void recount_clients_on_desktops();
     void rebuild_desktops();
     void rebuild_desktop_states();
     void set_desktop_urgent(xcb_window_t window);
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -95,7 +95,7 @@ namespace modules {
     string m_current_desktop_name;
     vector<int> m_occupied_desktops;
 
-    vector<xcb_window_t> m_clientlist;
+    vector<pair<xcb_window_t, unsigned int>> m_clientlist;
     vector<unique_ptr<viewport>> m_viewports;
     map<desktop_state, label_t> m_labels;
     label_t m_monitorlabel;
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -93,7 +93,7 @@ namespace modules {
     vector<string> m_desktop_names;
     unsigned int m_current_desktop;
     string m_current_desktop_name;
-    vector<int> m_occupied_desktops;
+    vector<unsigned int> m_occupied_desktops;
 
     vector<pair<xcb_window_t, unsigned int>> m_clientlist;
     vector<unique_ptr<viewport>> m_viewports;
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -129,7 +129,7 @@ namespace modules {
     vector<pair<xcb_window_t, unsigned int>> clients;
     vector<pair<xcb_window_t, unsigned int>> diff;
 
-    m_occupied_desktops.erase(m_occupied_desktops.begin(), m_occupied_desktops.end());
+    m_occupied_desktops.clear();
 
     for(xcb_window_t win : ewmh_util::get_client_list()) {
       pair<xcb_window_t, unsigned int> win_desktop(win, ewmh_util::get_desktop_from_window(win));
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -131,6 +131,8 @@ namespace modules {
     vector<pair<xcb_window_t, unsigned int>> clients;
     vector<pair<xcb_window_t, unsigned int>> diff;
 
+    m_log.info("%s: Tally has %u Desktops at start of rebuild_clientlist", name(), m_desktop_client_count.size());
+
     for(xcb_window_t win : ewmh_util::get_client_list()) {
       pair<xcb_window_t, unsigned int> win_desktop(win, ewmh_util::get_desktop_from_window(win));
       clients.emplace_back(win_desktop);
@@ -153,7 +155,10 @@ namespace modules {
       m_clientlist.emplace_back(win_desktop);
       m_desktop_client_count[win_desktop.second] = (m_desktop_client_count[win_desktop.second] - 1);
     }
+
+    m_log.info("%s: Tally has %u Desktops at the end of rebuild_clientlist", name(), m_desktop_client_count.size());
   }
+
 
   /**
    * Rebuild count of clients on each desktop
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -97,7 +97,7 @@ namespace modules {
    * Handler for XCB_PROPERTY_NOTIFY events
    */
   void xworkspaces_module::handle(const evt::property_notify& evt) {
-    std::lock_guard<std::mutex> guard(m_mutex, std::adopt_lock);
+    std::lock_guard<std::mutex> lock(m_mutex);
 
     if (evt->atom == m_ewmh->_NET_CLIENT_LIST) {
       rebuild_clientlist();
@@ -317,7 +317,7 @@ namespace modules {
    * Output content as defined in the config
    */
   bool xworkspaces_module::build(builder* builder, const string& tag) const {
-    std::lock_guard<std::mutex> guard(m_mutex, std::adopt_lock);
+    std::lock_guard<std::mutex> lock(m_mutex);
 
     if (tag == TAG_LABEL_MONITOR) {
       if (m_viewports[m_index]->state != viewport_state::NONE) {
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -133,24 +133,24 @@ namespace modules {
     vector<xcb_window_t> clients = ewmh_util::get_client_list();
     vector<xcb_window_t> new_windows;
 
-    std::sort(std::begin(clients), std::end(clients));
-    std::sort(std::begin(m_clientlist), std::end(m_clientlist));
-    std::fill(std::begin(m_desktop_occupied), std::end(m_desktop_occupied), false);
+    std::sort(clients.begin(), clients.end());
+    std::sort(m_clientlist.begin(), m_clientlist.end());
+    std::fill(m_desktop_occupied.begin(), m_desktop_occupied.end(), false);
 
-    auto it = std::begin(clients);
-    auto m_it = std::begin(m_clientlist);
-    while (it != std::end(clients) || m_it != std::end(m_clientlist)) {
-      if (it != std::end(clients)) {
+    auto it = clients.begin();
+    auto m_it = m_clientlist.begin();
+    while (it != clients.end() || m_it != m_clientlist.end()) {
+      if (it != clients.end()) {
         auto desktop = ewmh_util::get_desktop_from_window(*it);
         m_desktop_occupied[desktop] = true;
       }
-      if (m_it == std::end(m_clientlist) || (it != std::end(clients) && *it < *m_it)) {
+      if (m_it == m_clientlist.end() || (it != clients.end() && *it < *m_it)) {
         // listen for wm_hint (urgency) changes
         m_connection.ensure_event_mask(*it, XCB_EVENT_MASK_PROPERTY_CHANGE);
         // track window
         new_windows.emplace_back(*it);
         ++it;
-      } else if (it == std::end(clients) || (m_it != std::end(m_clientlist) && *it > *m_it)) {
+      } else if (it == clients.end() || (m_it != m_clientlist.end() && *it > *m_it)) {
         // untrack window
         m_it = m_clientlist.erase(m_it);
       } else {
@@ -158,7 +158,7 @@ namespace modules {
         ++m_it;
       }
     }
-    std::copy(std::begin(new_windows), std::end(new_windows), std::back_inserter(m_clientlist));
+    std::copy(new_windows.begin(), new_windows.end(), std::back_inserter(m_clientlist));
   }
 
   /**
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -97,7 +97,7 @@ namespace modules {
    * Handler for XCB_PROPERTY_NOTIFY events
    */
   void xworkspaces_module::handle(const evt::property_notify& evt) {
-    std::lock_guard<std::mutex> lock(m_mutex);
+    std::lock_guard<std::mutex> lock(m_event_mutex);
 
     if (evt->atom == m_ewmh->_NET_CLIENT_LIST || evt->atom == m_ewmh->_NET_WM_DESKTOP) {
       rebuild_clientlist();
@@ -317,7 +317,7 @@ namespace modules {
    * Output content as defined in the config
    */
   bool xworkspaces_module::build(builder* builder, const string& tag) const {
-    std::lock_guard<std::mutex> lock(m_mutex);
+    std::lock_guard<std::mutex> lock(m_event_mutex);
 
     if (tag == TAG_LABEL_MONITOR) {
       if (m_viewports[m_index]->state != viewport_state::NONE) {
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -99,7 +99,7 @@ namespace modules {
   void xworkspaces_module::handle(const evt::property_notify& evt) {
     std::lock_guard<std::mutex> lock(m_mutex);
 
-    if (evt->atom == m_ewmh->_NET_CLIENT_LIST) {
+    if (evt->atom == m_ewmh->_NET_CLIENT_LIST || evt->atom == m_ewmh->_NET_WM_DESKTOP) {
       rebuild_clientlist();
       rebuild_desktop_states();
     } else if (evt->atom == m_ewmh->_NET_DESKTOP_NAMES || evt->atom == m_ewmh->_NET_NUMBER_OF_DESKTOPS) {
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -98,6 +98,7 @@ namespace modules {
   void xworkspaces_module::handle(const evt::property_notify& evt) {
     if (evt->atom == m_ewmh->_NET_CLIENT_LIST) {
       rebuild_clientlist();
+      rebuild_desktop_states();
     } else if (evt->atom == m_ewmh->_NET_DESKTOP_NAMES || evt->atom == m_ewmh->_NET_NUMBER_OF_DESKTOPS) {
       m_desktop_names = get_desktop_names();
       rebuild_desktops();
@@ -110,8 +111,8 @@ namespace modules {
       if (icccm_util::get_wm_urgency(m_connection, evt->window)) {
         set_desktop_urgent(evt->window);
       }
-    } else if (evt->atom == m_ewmh->_NET_WM_DESKTOP) {
-      set_desktop_occupied(evt->window);
+    // } else if (evt->atom == m_ewmh->_NET_WM_DESKTOP) {
+    //   set_desktop_occupied(evt->window);
     } else {
       return;
     }
@@ -145,6 +146,8 @@ namespace modules {
         m_connection.ensure_event_mask(win, XCB_EVENT_MASK_PROPERTY_CHANGE);
         // track window
         m_clientlist.emplace_back(win);
+        m_occupied_desktops.emplace_back(ewmh_util::get_desktop_from_window(win));
+
       }
     }
   }
@@ -213,8 +216,12 @@ namespace modules {
   void xworkspaces_module::rebuild_desktop_states() {
     for (auto&& v : m_viewports) {
       for (auto&& d : v->desktops) {
+        auto desktop_occupied = std::find(std::begin(m_occupied_desktops), std::end(m_occupied_desktops), d->index);
+
         if (m_desktop_names[d->index] == m_current_desktop_name) {
           d->state = desktop_state::ACTIVE;
+        } else if (desktop_occupied != std::end(m_occupied_desktops)) {
+          d->state = desktop_state::OCCUPIED;
         } else {
           d->state = desktop_state::EMPTY;
         } 
@@ -256,27 +263,6 @@ namespace modules {
       for (auto&& d : v->desktops) {
         if (d->index == desk && d->state != desktop_state::URGENT) {
           d->state = desktop_state::URGENT;
-
-          d->label = m_labels.at(d->state)->clone();
-          d->label->reset_tokens();
-          d->label->replace_token("%index%", to_string(d->index - d->offset + 1));
-          d->label->replace_token("%name%", m_desktop_names[d->index]);
-          d->label->replace_token("%icon%", m_icons->get(m_desktop_names[d->index], DEFAULT_ICON)->get());
-          return;
-        }
-      }
-    }
-  }
-
-  /**
-   * Find window and set corresponding desktop to occupied
-   */
-  void xworkspaces_module::set_desktop_occupied(xcb_window_t window) {
-    auto desk = ewmh_util::get_desktop_from_window(window);
-    for (auto&& v : m_viewports) {
-      for (auto&& d : v->desktops) {
-        if (d->index == desk && d->state != desktop_state::OCCUPIED) {
-          d->state = desktop_state::OCCUPIED;
 
           d->label = m_labels.at(d->state)->clone();
           d->label->reset_tokens();
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -86,11 +86,11 @@ namespace modules {
     m_current_desktop_name = m_desktop_names[m_current_desktop];
 
     rebuild_desktops();
-    recount_clients_on_desktops();
-    rebuild_desktop_states();
 
     // Get _NET_CLIENT_LIST
     rebuild_clientlist();
+
+    rebuild_desktop_states();
   }
 
   /**
@@ -105,7 +105,7 @@ namespace modules {
     } else if (evt->atom == m_ewmh->_NET_DESKTOP_NAMES || evt->atom == m_ewmh->_NET_NUMBER_OF_DESKTOPS) {
       m_desktop_names = get_desktop_names();
       rebuild_desktops();
-      recount_clients_on_desktops();
+      rebuild_clientlist();
       rebuild_desktop_states();
     } else if (evt->atom == m_ewmh->_NET_CURRENT_DESKTOP) {
       m_current_desktop = ewmh_util::get_current_desktop();
@@ -130,55 +130,35 @@ namespace modules {
    * Rebuild the list of managed clients
    */
   void xworkspaces_module::rebuild_clientlist() {
-    vector<pair<xcb_window_t, unsigned int>> clients;
-    vector<pair<xcb_window_t, unsigned int>> diff;
+    vector<xcb_window_t> clients = ewmh_util::get_client_list();
+    vector<xcb_window_t> new_windows;
 
-    m_log.info("%s: Tally has %u Desktops at start of rebuild_clientlist", name(), m_desktop_client_count.size());
+    std::sort(std::begin(clients), std::end(clients));
+    std::sort(std::begin(m_clientlist), std::end(m_clientlist));
+    std::fill(std::begin(m_desktop_occupied), std::end(m_desktop_occupied), false);
 
-    for(xcb_window_t win : ewmh_util::get_client_list()) {
-      pair<xcb_window_t, unsigned int> win_desktop(win, ewmh_util::get_desktop_from_window(win));
-      clients.emplace_back(win_desktop);
-    }
-
-    std::set_difference(
-        m_clientlist.begin(), m_clientlist.end(), clients.begin(), clients.end(), back_inserter(diff));
-    for (auto&& win_desktop : diff) {
-      // untrack window
-      m_clientlist.erase(std::remove(m_clientlist.begin(), m_clientlist.end(), win_desktop), m_clientlist.end());
-      m_desktop_client_count[win_desktop.second] = (m_desktop_client_count[win_desktop.second] + 1);
-    }
-
-    std::set_difference(
-        clients.begin(), clients.end(), m_clientlist.begin(), m_clientlist.end(), back_inserter(diff));
-    for (auto&& win_desktop : diff) {
-      // listen for wm_hint (urgency) changes
-      m_connection.ensure_event_mask(win_desktop.first, XCB_EVENT_MASK_PROPERTY_CHANGE);
-      // track window
-      m_clientlist.emplace_back(win_desktop);
-      m_desktop_client_count[win_desktop.second] = (m_desktop_client_count[win_desktop.second] - 1);
-    }
-
-    m_log.info("%s: Tally has %u Desktops at the end of rebuild_clientlist", name(), m_desktop_client_count.size());
-  }
-
-
-  /**
-   * Rebuild count of clients on each desktop
-   */
-  void xworkspaces_module::recount_clients_on_desktops() {
-    m_log.info("%s: Tally has %u Desktops, EWMH has %u", name(), m_desktop_client_count.size(), m_desktop_names.size());
-    if (m_desktop_client_count.size() != m_desktop_names.size()) {
-      m_log.info("%s: Resizing and Recounting Tally", name());
-      m_desktop_client_count.resize(m_desktop_names.size(), 0);
-
-      m_desktop_client_count.clear();
-
-      for(xcb_window_t win : ewmh_util::get_client_list()) {
-        auto desk = ewmh_util::get_desktop_from_window(win);
-        m_desktop_client_count[desk] = (m_desktop_client_count[desk] + 1);
-        m_log.info("%s: Desktop %u has now %u windows", name(), (desk + 1), m_desktop_client_count[desk]);
+    auto it = std::begin(clients);
+    auto m_it = std::begin(m_clientlist);
+    while (it != std::end(clients) || m_it != std::end(m_clientlist)) {
+      if (it != std::end(clients)) {
+        auto desktop = ewmh_util::get_desktop_from_window(*it);
+        m_desktop_occupied[desktop] = true;
+      }
+      if (m_it == std::end(m_clientlist) || *it < *m_it) {
+        // listen for wm_hint (urgency) changes
+        m_connection.ensure_event_mask(*it, XCB_EVENT_MASK_PROPERTY_CHANGE);
+        // track window
+        new_windows.emplace_back(*it);
+        ++it;
+      } else if (it == std::end(clients) || *it > *m_it) {
+        // untrack window
+        m_it = m_clientlist.erase(m_it);
+      } else {
+        ++it;
+        ++m_it;
       }
     }
+    std::copy(std::begin(new_windows), std::end(new_windows), std::back_inserter(m_clientlist));
   }
 
   /**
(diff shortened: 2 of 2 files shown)


NBonaparte:
So far, the best way I can think of to solve most of the issues is by linking each window with its desktop. This would involve changing `m_clientlist` to `vector<pair<xcb_window_t, unsigned int>>`. Now, removing and tracking windows should be easier, as we know which desktop it was last in and remove the corresponding value from `m_occupied_desktops`.
Speaking of which, it might be easier to have `m_occupied_desktops` as a tally system for each desktop, with the index corresponding to the desktop number and the value to the number of windows within that desktop.

kronn:
I am very sorry and feel like wasting your time. Full disclosure at this point: I know how to program, but my C++ is limited to the two commits you already saw. Mostly, I am missing a way to output the state of certain datastructures, so I can see what exactly is known to polybar. Without this, I have the nagging feeling to miss something obvious.

The type-change you proposed surpasses my skill-level. The compiler output tells me that I hurt it badly, but I cannot make sense of this so far. 

I feel like there is something simple missing, since I see (in my bar) that only the desktops of other visible, but unfocussed viewports are not occupied. Since I can see them, I known there are windows present.

A short test seems to prove my assumption: If I only have one viewport/screen, then I get the list of occupied desktops shown correctly. Just switching to one big-ass monitor might be a solution, but it is both too expensive and unsatisfying :grin: One thing that does not work correctly yet is the removal of the "occupied"-state, but this is something that I expect from the current version. 

Also, just to clarify the terminology: with a tally system, you mean a datastructure to keep track of the windows/clients of each desktop. Something like (in JSON-notation)

```
{
  web: ['chrome'],
  work: ['vim', 'emacs', 'notepad'],
  chat: ['pidgin', 'skype']
}
```

where all the strings are the internal ids of the corresponding windows-titles and desktop-labels?

Tomorrow, I will try to get the removal working with the current datastructures. Then I have at least a working version for a simgle-monitor setup. This won't be enough for the complete feature I have in mind, but I need small steps I can actually reach...

kronn:
After reading and experimenting a bit, I was able to change the type rather easily. While doing so, I notice how right this feels. Thanks for the advice, @NBonaparte. Also, listening to the compiler is useful during such refactoring, it turns out. :wink: The current state (esp. https://github.com/jaagr/polybar/pull/882/commits/ed6db8cffe4e2cadf48ea952edeeb046ed3c4547 and https://github.com/jaagr/polybar/pull/882/commits/2483681f2c5afb61b91236f917cd807a27d4ad49) should keep track of windows in their desktops/workspaces.

For my local testing, I also readded the "sending windows to certain desktops"-function. I have the feeling that it will be not needed anymore, but it does not seem to hurt for now.

At this point I also feel like I might have introduced a race condition, if two bars are using the xworkspaces module. Since my knowledge of the internal structure of polybar is limited: could the current clear/rebuild/sort/unique-dance I do with the `m_occupied_workspaces` be a victim to reentrancy? Is the datastructure shared between instances of polybar? I would say "no", but I really want to make sure, just in case.

Lastly, I noticed an odd behaviour: I have windows on desktops 5, 6, 7 and 8. When I go through them one by one, the desktop before the previous one (depending on the order I jump), is marked unoocupied, although there is still a window present. If you can spot the error right away, I would be grateful for pointers. Otherwise, I will research that in the coming days.

NBonaparte:
There should be no race condition, as all the data structures for each module are unique.

I cannot seem to reproduce your issue. Is there a specific procedure to produce this?

kronn:
Race Conditions: okay. I thought and hoped so, just wanted to make sure.

Odd Behaviour: Maybe I looked at a incomplete build... I cannot reproduce it either, now.

In my own personal agenda, I have two open topics:

1. Showing occupied workspaces on other visible workspaces as occupied (or maybe as "on visible, yet not active")
    * showing them as occupied instead of empty seens way easier than a new state "visible". 
    * therefore I will only show them as occupied, which is at least correct.
    * after a bit of research, "just collecting the occupied" desktop seems more complicated than I thought. Either the underlying X11-libs or my WM (xmonad) seem to be clever and only return the windows that are hidden. The others are somehow not in the `_NET_CLIENT_LIST`. 
2. I rebuild the `m_occupied_desktops` a lot. I leave the structure, but clear and rebuild every time. 
    * this seems less than ideal, mostly a waste of CPU-cycles
    * OTOH, maybe is fast enough and handling the life-cycle of each window is more complicated to maintain (and create in the first place) than just rebuilding this `vector<unsigned int>` of mostly up to 10 elements.

While 1. is a personal must for this feature to be complete, 2. is more a question that a task. Which route would you choose?

NBonaparte:
For 2:

The tally/counting system IMO should solve this in a simple manner. Just to clarify my idea, the vector would have an element per desktop, with the number of windows in that specific desktop as the value. (Maybe change the name to `m_desktop_client_count` or something else that makes more sense.)

For example, let's say we have 3 windows on desktop 1, 2 on desktop 2, and no other windows. The vector would be `[3, 2, 0, 0, 0]` (for 5 desktops). Now we create another window on desktop 1. When we track the window (line 157), we would also increment the first element (corresponding to the first desktop) by 1. If we close any window, we should also decrement the element by 1 (after 148), and since we know which desktop it was in (`win_desktop.second`) this shouldn't be too hard.

If a window moves to a different desktop, we should be able to track that by listening for `_NET_WM_DESKTOP` in the handler, since we already requested for property changes for it. Then we should be able to change the window count appropriately, and update the desktop number in `m_clientlist`.

kronn:
Well, my personal 1. is a bug in xmonad and therefore outside of the scope of this PR (see linked issue).

As for 2: Thanks for the clarification. I think I misunderstood what you had in mind. 

The pure counter sounds good at first, but - coming from a web-app background - I fear maintaining a counter for a long time. Maybe it's premature optimization or being burnt by a web-environment, but I want to recalculate the whole `m_desktop_client_count` if we ever reach -1 somewhere. This forces the recalculation into its own method. I would move the current recalc-part there and output an info-logging if that happens. This way, we can keep an eye on it. What do you think of this?

Tomorrow, I will work on moving `m_occupied_desktops` to your proposed `m_desktop_client_count`. Since git does not forget old code, we can decide about handling the aforementioned edge-case later.

kronn:
By now, I distrust XMonads ability to be a good EWMH-citizen. With the pure counter-implementation I now pushed, it is far easier to mark a workspace as occupied. Keeping track of the window-count seems sadly a lost cause. XMonad advertises changes in the desktop-names which might trigger a recount. Also, the Tally seems to be empty every time a recount is in order. Since I only empty it before the recount, I am a bit confused. 

I pushed the version with debugging info-log statements, so you can see where I wanted to know more. Judging from my local output, I see several (4) entries to the recount-function. The xworkspaces-modules is only active in one bar.

The fresh count seems to always be accurate (within the limits of Xmonads current implementation). Keeping a continuous count seems to not work. Or maybe it does, but the frequent changes to the `_NET_DESKTOP_NAMES` sabotage that.

The current version needs less logging in order to be mergeable. This information is not even useful at a trace-level in normal operations. Other than that, it seems to work okay for me. The frequent desktop-name changes trigger enough complete recounts to keep the score correct.

Therefore I would be happy if anyone with different WM could test this. Who could do that?

NBonaparte:
I use bspwm and @patrick96 uses hlwm (I think), so we could test it out on those WMs, as well as i3 and maybe openbox. Others are also welcome to test this.

kronn:
I think that the bar works for me since XMonad issues a metric ****ton of changes to the rather static `_NET_DESKTOP_NAMES`. Those trigger a complete recount of clients/desktop.

I consider this buggy behaviour by the WM, although the compiled polybar-binary does not even care, it just performs. I seriously wonder how other WMs handle this. I will try out another WM (openbox, most likely) tomorrow to get a feeling for this. Since bspwm and i3 have their own modules, they are not as affected as those relying on EWMH... 

hlwm is not in my Ubuntu sources. While this could be remedied, I prefer to go one step at a time :-)

Then again, maybe it's perfectly normal to change the desktop-count and their names all the time. If that would be the case across WMs I would maybe rethink `xworkspaces_module::rebuild_desktops()`. Well, maybe... I'm missing data here.

mbrgm:
Any progress on this?

kronn:
@mbrgm I had very little time lately to work on this. From trying my local version I see that the count is sometimes off or even completely missing (all WS appear empty). When I then "switch" to the same WS again, everything is updated. I think that I handle the X events a little wrong. This may be rooted in my patched version of xmonad. Or something else...

Either way, debugging this was not a priority lately, due to work and such. I will see if I can get to it in the coming week. At the very least I will rebase my local branch onto current master.

Would you be able and willing to compile and test this version yourself?

kronn:
Oh, and it segfaults all the time. I really need to take the time and fix that. Right now, I have a shortcut to restart polybar because of the many crashes. While convenient for me, I would not impose that on everyone.

mbrgm:
> Would you be able and willing to compile and test this version yourself?

Yup, I'm fine with that. Also no problem about segfaulting. Could you push your branch and post a link?

kronn:
@mbrgm I rebased my branch onto the current master and pushed it to https://github.com/kronn/polybar/tree/list-occupied-workspaces

Currently, I have the following problems:
- sometimes crashing, seems to coincide with changes in the "title" of a window.
- active WS is sometimes only updated after second switch to a WS and then switching back to a previous state
- I feel like the counting is done wrong, but I might just lack understanding

The current build emits a metric ton of logging-output on info-level, whenever some counting or checking is done. Although this clearly is debugging-information, I made these info-level to have them better readable for now. In a final version, I would remove them entirely or reduce them to very few debug-level loggings.

I would like to get some information and feedback from running this under another EWMH-WM than a patched xmonad.

My setup is:

- 1 laptop monitor and two external monitors via docking station
- polybar head + my extensions
- xmonad 0.13 with some changes to the EWMH-adapter, to habe windows always report the correct WS. Without that, xmonad always pretends that all visible windows are on the current WS to help gnome-panel. See referenced PR above (2017-12-17).

jonhoo:
I wonder if the crash you're seeing is related to #1033...

mbrgm:
To be honest, as it's been silent in here and I needed a working solution quickly, I used the weekend to switch to writing polybar-formatted output to a fifo using xmonad's event log hook, which works just fine and seems stable. If someone is interested in the solution, I can post a gist.

kronn:
@mbrgm I will continue with an EWMH-solution, but you should contribute your solution to https://github.com/x70b1/polybar-scripts

if you don't want to contribute there, a gist would be very welcome.

taschenb:
@kronn Do you still work on a solution for that? I prototyped the discussion from this pull request and currently running it in herbstluftwm without any issues. [feat/occupied-workspaces](https://github.com/taschenb/polybar/tree/feat/occupied-workspaces). I hope you find it helpful and continue your work.

kronn:
@taschenb In general, I think I am done. I get segfaults a lot, but those can be rooted in the interaction between xmonad and polybar.

I will take a look at your solution this week and try to finish this PR up. All I wanted was solution that works not only for me but for at least one other WM.

swillner:
There were some issues in the code (e.g. using `std::set_difference` on unsorted vectors), so I took the liberty to fix these and open a new PR (based on @kronn's changes for the credit): #1429 ;)
In there, I switched to a `std::vector<bool>` to just track if a workspace is occupied rather than counting clients as `ewmh_util::get_desktop_from_window` has to be called for each client in `rebuild_clientlist()` anyway...
I have only tested it for xmonad, for which it works fine.

jonhoo:
This is a _little_ orthogonal, but it would be *awesome* if this feature could also be extended to support just highlighting occupied workspaces differently, as opposed to _only_ listing occupied workspaces.

swillner:
hmm, that would easily be done. how would the additional information (number of clients per workspace) be used?

jonhoo:
Well, specifically, I'd like to be able to have workspaces that have something on them styled differently than those that are empty. Basically, I want `label-occupied` to work :p

swillner:
ah, well that is the whole purpose of this PR ;)
I thought, you wanted them styled depending on the exact number of clients rather than just if they are occupied or not (which is already done by this PR)

jonhoo:
Ah, no, just a binary yes/no is perfectly fine. Awesome, then I'm happy! Thanks for taking this on! :D

kronn:
@swillner I cherry-picked your change from your PR to have all in one place. This solves indeed the problems I had with the module before. Therefore, I removed the debug-logging I added to understand the event-timing and data flow.

Have you had a chance to test this with a multi-monitor setup? xmonad had some issues with that for me. I solved it by changing the EWMH-module of xmonad, which might have broken things for me.

NBonaparte:
Is the segfault issue fixed? Let us know when this is ready to be reviewed.

swillner:
I think it should be fixed. At least I don't get any more segfaults using xmonad.

kronn:
I did not get any segfault since applying the change, but I had whole days without them before. However, since I am not the only one running this under xmonad, I am quite optimistic that the issue is fixed.

swillner:
I found a problem when all clients are removed, which is fixed in cc3e2f0. @kronn Can you cherry-pick that, please?

kronn:
Sorry for the delay in cherry-picking. I really welcome your fixes to this feature.

src/modules/xworkspaces.cpp
```diff
@@ -122,27 +127,34 @@ namespace modules {
...
-
-    if (m_clientlist.size() > clients.size()) {
-      std::set_difference(
-          m_clientlist.begin(), m_clientlist.end(), clients.begin(), clients.end(), back_inserter(diff));
-      for (auto&& win : diff) {
-        // untrack window
-        m_clientlist.erase(std::remove(m_clientlist.begin(), m_clientlist.end(), win), m_clientlist.end());
+    vector<xcb_window_t> new_windows;
+
+    std::sort(std::begin(clients), std::end(clients));
```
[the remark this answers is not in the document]
kronn: Is there a code-style/linter for this project? I whole-heartedly agree with the preference, I just did not see it directly.

taschenb:
Really awesome that this gets continued.

At least for herbstluftwm, the current pull request does not yet work when moving a window to another workspace. The problem is that herbstluftwm only fires a single `_NET_WM_DESKTOP` event in that case. xmonad on the other hand fires everything, as far as I can remember. Including the used `_NET_DESKTOP_NAMES` as well as one `_NET_WM_DESKTOP` for each window in any workspace.

kronn:
@taschenb thanks, I was hoping for information like that. I changed the handling to also listen to `_NET_WM_DESKTOP`-changes. Reading through the EWMH-spec and the code for herbstluftwm, I think this is a the correct way to do things anyway.

kronn:
@taschenb thanks, I was hoping for information like that. I changed the handling to also listen to `_NET_WM_DESKTOP`-changes. Reading through the EWMH-spec and the code for herbstluftwm, I think this is a the correct way to do things anyway.

kronn:
@taschenb thanks, I was hoping for information like that. I changed the handling to also listen to `_NET_WM_DESKTOP`-changes. Reading through the EWMH-spec and the code for herbstluftwm, I think this is a the correct way to do things anyway.

kronn:
@taschenb can you confirm or refute the change as beneficial for herbstluftwm?

(bump because github had issues when posting this the first time)

willprice:
@kronn,
This is brilliant, no segfaults any more and occupied workspaces are working great :+1:. Thanks for your work on this issue. You've made me a very happy polybar user.

dkasak:
I'm confirming that I haven't seen any more segfaults either.

willprice:
@kronn I'm happy to test features under xmonad if necessary

kronn:
@patrick96 yes, this PR should resolve those two issues as well.

zamlz:
I'm sad that this pull request is not ready yet. I desperately need this feature.

include/modules/xworkspaces.hpp
```diff
@@ -103,6 +105,7 @@ namespace modules {
     size_t m_index{0};
 
     event_timer m_timer{0L, 25L};
+    mutable std::mutex m_mutex;
```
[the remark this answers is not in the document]
kronn: I renamed the mutex and added a comment. However, this one mutex is doing two things:
- serialize event-handling
- protect m_viewports

Is this double-usage of a mutex okay? The "mutable"-flag/modifier/info is only needed to protect the m_viewports, since the event-handler is const. Am I overthinking this?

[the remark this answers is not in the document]
kronn: These 2 commits contain a lot of explanation on this:
* https://github.com/jaagr/polybar/commit/6c3881b18c01c820a13e3f92ed1121775d5dc1b0
* https://github.com/jaagr/polybar/commit/f4fab2ded14c2067c67e5f22cab1c05fe6b4f4a0

kronn:
@patrick96 I am very happy that you reported this working for both i3 and herbstluft. I left one of your comments unresolved, because I have some follow-up questions.

xynxynxyn:
I think there's an off by one error having to do with the workspace names.
I tested it on 2bwm and the number of workspaces displayed by xworkspaces is always one more than the actual number, causing 2bwm to crash when the last workspace is clicked on.
I assume this is because 2bwm does not set the workspace names.

https://github.com/kronn/polybar/blob/7639f3b2debd73ad93b1289c07fb0fd6998ec6cb/src/modules/xworkspaces.cpp#L255

Removing the + 1 fixed it for 2bwm but I did not test it for other wm's.

xynxynxyn:
The current version segfaults on 2bwm if a window is fixed.

This is because a fixed window doesn't have a desktop associated with it.
It appears that `auto desktop = ewmh_util::get_desktop_from_window(*it_old)` returns `0xffffffff` if the workspace is not defined.
A simple if check would fix this issue.

https://github.com/xynxynxyn/polybar/blob/9d125cbc2fa64b0635f11e40c97839541f74ac6c/src/modules/xworkspaces.cpp#L146-L148

```
Thread 1 "polybar" received signal SIGSEGV, Segmentation fault.
-----------------------------------------------------------------------------------------------------------------------[regs]
  RAX: 0x0000555575C25B08  RBX: 0x0000555555C26090  RBP: 0x00007FFFFFFFB240  RSP: 0x00007FFFFFFFB240  o d I t s z a p c 
  RDI: 0x00007FFFFFFFB280  RSI: 0x0000000000000001  RDX: 0x8000000000000000  RCX: 0x0000555575C25B08  RIP: 0x0000555555A539B4
  R8 : 0x0000000000000004  R9 : 0x0000000000000003  R10: 0x0000000000000000  R11: 0x0000000000000000  R12: 0x0000000000000000
  R13: 0x0000555555C1A060  R14: 0x000000000000000B  R15: 0x0000000000000000
  CS: 0033  DS: 0000  ES: 0000  FS: 0000  GS: 0000  SS: 002B				
-----------------------------------------------------------------------------------------------------------------------[code]
=> 0x555555a539b4 <std::_Bit_reference::operator=(bool)+26>:	mov    rcx,QWORD PTR [rax]
   0x555555a539b7 <std::_Bit_reference::operator=(bool)+29>:	mov    rax,QWORD PTR [rbp-0x8]
   0x555555a539bb <std::_Bit_reference::operator=(bool)+33>:	mov    rdx,QWORD PTR [rax+0x8]
   0x555555a539bf <std::_Bit_reference::operator=(bool)+37>:	mov    rax,QWORD PTR [rbp-0x8]
   0x555555a539c3 <std::_Bit_reference::operator=(bool)+41>:	mov    rax,QWORD PTR [rax]
   0x555555a539c6 <std::_Bit_reference::operator=(bool)+44>:	or     rdx,rcx
   0x555555a539c9 <std::_Bit_reference::operator=(bool)+47>:	mov    QWORD PTR [rax],rdx
   0x555555a539cc <std::_Bit_reference::operator=(bool)+50>:	jmp    0x555555a539f3 <std::_Bit_reference::operator=(bool)+89>
-----------------------------------------------------------------------------------------------------------------------------
std::_Bit_reference::operator= (this=0x7fffffffb280, __x=0x1) at /usr/include/c++/8.2.1/bits/stl_bvector.h:89
89		*_M_p |= _M_mask;
(gdb) bt
#0  std::_Bit_reference::operator= (this=0x7fffffffb280, __x=0x1) at /usr/include/c++/8.2.1/bits/stl_bvector.h:89
#1  0x0000555555a50d39 in polybar::v3_1_0_346_g9573e83_dev::modules::xworkspaces_module::rebuild_clientlist (this=0x555555c24160) at /home/nyx/Documents/polybar/src/modules/xworkspaces.cpp:147
#2  0x0000555555a50852 in polybar::v3_1_0_346_g9573e83_dev::modules::xworkspaces_module::handle (this=0x555555c24160, evt=...) at /home/nyx/Documents/polybar/src/modules/xworkspaces.cpp:103
#3  0x0000555555a7e3c0 in xpp::event::detail::dispatcher::dispatch<xpp::x::event::property_notify<polybar::v3_1_0_346_g9573e83_dev::connection&> > (this=0x555555c24490, e=...) at /home/nyx/Documents/polybar/lib/xpp/include/xpp/event.hpp:206
#4  0x0000555555a7bd00 in xpp::event::registry<polybar::v3_1_0_346_g9573e83_dev::connection&, xpp::randr::extension, xpp::composite::extension, xpp::xkb::extension>::handle<xpp::x::event::property_notify<polybar::v3_1_0_346_g9573e83_dev::connection&> > (this=0x555555c1db40, event=...) at /home/nyx/Documents/polybar/lib/xpp/include/xpp/event.hpp:108
#5  0x0000555555a78b4a in xpp::event::registry<polybar::v3_1_0_346_g9573e83_dev::connection&, xpp::randr::extension, xpp::composite::extension, xpp::xkb::extension>::handler::operator()<xpp::x::event::property_notify<polybar::v3_1_0_346_g9573e83_dev::connection&> > (this=0x7fffffffb430, event=...) at /home/nyx/Documents/polybar/lib/xpp/include/xpp/event.hpp:124
#6  0x0000555555a769c2 in xpp::x::event::dispatcher<polybar::v3_1_0_346_g9573e83_dev::connection&>::operator()<xpp::event::registry<polybar::v3_1_0_346_g9573e83_dev::connection&, xpp::randr::extension, xpp::composite::extension, xpp::xkb::extension>::handler> (this=0x555555c1db40, handler=..., event=std::shared_ptr<xcb_generic_event_t> (use count 2, weak count 0) = {...}) at /home/nyx/Documents/polybar/lib/xpp/include/xpp/proto/x.hpp:11497
#7  0x0000555555a75d4e in xpp::event::registry<polybar::v3_1_0_346_g9573e83_dev::connection&, xpp::randr::extension, xpp::composite::extension, xpp::xkb::extension>::dispatch<xpp::x::extension> (this=0x555555c1db40, event=std::shared_ptr<xcb_generic_event_t> (use count 2, weak count 0) = {...}) at /home/nyx/Documents/polybar/lib/xpp/include/xpp/event.hpp:133
#8  0x0000555555a75639 in xpp::event::registry<polybar::v3_1_0_346_g9573e83_dev::connection&, xpp::randr::extension, xpp::composite::extension, xpp::xkb::extension>::dispatch<xpp::x::extension, xpp::randr::extension, xpp::composite::extension, xpp::xkb::extension> (this=0x555555c1db40, event=std::shared_ptr<xcb_generic_event_t> (use count 2, weak count 0) = {...}) at /home/nyx/Documents/polybar/lib/xpp/include/xpp/event.hpp:140
#9  0x0000555555a74df5 in xpp::event::registry<polybar::v3_1_0_346_g9573e83_dev::connection&, xpp::randr::extension, xpp::composite::extension, xpp::xkb::extension>::dispatch (this=0x555555c1db40, event=std::shared_ptr<xcb_generic_event_t> (use count 2, weak count 0) = {...}) at /home/nyx/Documents/polybar/lib/xpp/include/xpp/event.hpp:61
#10 0x0000555555a73e38 in polybar::v3_1_0_346_g9573e83_dev::connection::dispatch_event (this=0x555555c1dac0, evt=std::shared_ptr<xcb_generic_event_t> (use count 2, weak count 0) = {...}) at /home/nyx/Documents/polybar/src/x11/connection.cpp:252
#11 0x000055555595205b in polybar::v3_1_0_346_g9573e83_dev::controller::read_events (this=0x555555c007f0) at /home/nyx/Documents/polybar/src/components/controller.cpp:338
#12 0x0000555555950f92 in polybar::v3_1_0_346_g9573e83_dev::controller::run (this=0x555555c007f0, writeback=0x0, snapshot_dst="") at /home/nyx/Documents/polybar/src/components/controller.cpp:204
#13 0x00005555558f62aa in main (argc=0x2, argv=0x7fffffffe788) at /home/nyx/Documents/polybar/src/main.cpp:144
```

Lomadriel:
> It appears that `auto desktop = ewmh_util::get_desktop_from_window(*it_old)` returns `0xffffffff` if the workspace is not defined.

Same problem with sticky window + scratchpad in i3wm.

> 0xFFFFFFFF indicates that the window SHOULD appear on all desktops.
https://standards.freedesktop.org/wm-spec/wm-spec-1.3.html#idm140130317612768)

src/modules/xworkspaces.cpp
```diff
@@ -124,27 +129,38 @@ namespace modules {
...
+    auto it_old = clients.begin();
+    auto it_new = m_clientlist.begin();
+
+    // Traverse both the stored and the up-to-date clientlist to update our
+    // stored list with as little changes as possible (add windows that are
+    // only in the new list, remove windows that are no longer in the list)
+    while (it_old != clients.end() || it_new != m_clientlist.end()) {
+      if (it_old != clients.end()) {
+        auto desktop = ewmh_util::get_desktop_from_window(*it_old);
+        m_desktop_occupied[desktop] = true;
```
Lomadriel:
```suggestion
        if (desktop != 0xffffffff) {
          m_desktop_occupied[desktop] = true;
        }
```
Fixes a crash with windows with no desktop. (like i3 sticky windows, …)
See https://github.com/polybar/polybar/pull/882#issuecomment-483458242 and https://github.com/polybar/polybar/pull/882#issuecomment-483488566

src/modules/xworkspaces.cpp
```diff
@@ -86,21 +86,26 @@ namespace modules {
...
+
+    rebuild_desktop_states();
   }
 
   /**
    * Handler for XCB_PROPERTY_NOTIFY events
    */
   void xworkspaces_module::handle(const evt::property_notify& evt) {
-    if (evt->atom == m_ewmh->_NET_CLIENT_LIST) {
+    std::lock_guard<std::mutex> lock(m_event_mutex);
```
Lomadriel:
```suggestion
    std::lock_guard<std::mutex> lock(m_buildlock);
```
The `m_buildlock` mutex is used to protect the `build` phase and is locked before the call of `build`, so this mutex can be removed.
https://github.com/polybar/polybar/blob/cb75857ff66543e9aedac3f87df130ba03132471/include/modules/meta/base.inl#L124-L176
The lock is at line 125 and the call to `build` is at line 160

dkasak:
@kronn, are you still planning to work on this?

kronn:
@dkasak yes, I plan to come back to this in the coming week. I had an accident with my thumb which took me away from the keyboard for a while.

It bugs me a great deal to see this at the 99%-stage.

Lomadriel:
> @dkasak yes, I plan to come back to this in the coming week. I had an accident with my thumb which took me away from the keyboard for a while.
> 
> It bugs me a great deal to see this at the 99%-stage.

If you don't have the time to work on this, I can push the fix on your branch directly.
Don't feel obligated to accept, if you want to do it yourself. It just I'm on holidays for now so I can freely work on polybar until september.
If you do it by yourself, I'll review that asap.

In every case, the polybar team (except me) is quite busy for now so if you want to take your time, you can :wink:

kronn:
@Lomadriel I guess I should just let you do it. My personal life is a little hectic right now. My plan was to take your first suggestions as is (since I do not have 2bwm or i3 installed/configured/running) and also rebase it on top of the current release. I think there may be some changes needed, because there always are. ;-) 

Regarding the Mutex/Lock-Naming, please see https://github.com/polybar/polybar/pull/882#discussion_r274813999 (and the referenced commits there). The mutex did (last time I checked, might have changed) protect two separate locations. Then again, I am not a seasoned C++-Programmer, so you might know better how to deal with those locks.

src/modules/xworkspaces.cpp
kronn: This would fix the crash, but would it handle `0xFFFFFFFF` correctly? If it should appear on all desktops, I would somehow expect all desktops to be occupied. Or maybe none, because it is so omnipresent that it does not need special highlighting. :confused:

Lomadriel: `0xFFFFFFFF` isn't a valid id for a desktop.
For example `polybar` is a window that should appears on every desktop.
I think we shouldn't consider that all desktop are occupied.
In i3wm, you have the concept of sticky windows that are above the desktop, so they aren't on a desktop, there are really above them.

Lomadriel:
Could I ask you to check the option "Allow edits from maintainers.", so I can push directly on your branch? 

image

> Regarding the Mutex/Lock-Naming, please see [#882 (comment)](https://github.com/polybar/polybar/pull/882#discussion_r274813999) (and the referenced commits there). The mutex did (last time I checked, might have changed) protect two separate locations. Then again, I am not a seasoned C++-Programmer, so you might know better how to deal with those locks.

I tried to figure out what needed to be protected and when, for now polybar is using two mutexes: `m_buildlock` and `m_updatelock`, however we should use only one mutex (see #1790).
For now, I decided to use `m_buildlock` to protect the data, there are a lot of variables that aren't protected in this module. I tried to fix that and I think it's okay now. I can't check using my usual tools so I'm not 100% sure that I fixed everything since they are a lot of concurrency problems that come from other component of polybar that are interacting with the modules.

The data need to be protected in almost every function at the exception of the constructor.
In `build`, the `m_buildlock` is already locked so I haven't added a lock here. `input` also needed to be protected by a lock.

kronn:
The checkbox you mentioned was already set to allow edits from maintainers. Paying my respects to the powers that be, I did try to turn if off and on again. I hope that this fixes things and allow you to push as well. If that is the case, I have a guess about the implementation. :wink: If that does not work, I could change my fork so you can push there.

Regarding the mutexes: I remember that sometimes the state of a desktop (occupied or free) would be wrong and then be right again after I did something that fired an event. Also, xmonad fires about every event possible (read: a metric ton of them) if anything changes. In order to cope with that firehose of event, I first tried to add the lock in the handle-function. The problems only disappeared after the buildlock was also introduced and I frankly didn't check again if the first mutex was still needed. 

I can however run your version for a week and see if it crashes or behaves wrongly. Would that be okay?

Lomadriel:
> I can however run your version for a week and see if it crashes or behaves wrongly. Would that be okay?

That seems a good idea to me.

Lomadriel:
> The checkbox you mentioned was already set to allow edits from maintainers. Paying my respects to the powers that be, I did try to turn if off and on again. I hope that this fixes things and allow you to push as well. If that is the case, I have a guess about the implementation. wink If that does not work, I could change my fork so you can push there.

I still can push on your branch but no problem, I'll work on my fork.
For now I found a bug that was already here before your PR :/
I think I'll fix this bug before continuing working on this since I can't really test correctly this feature, see #1849.

kronn:
I added you to my repo as collaborator, so you can push to my fork.

Lomadriel:
> I added you to my repo as collaborator, so you can push to my fork.

Thanks :wink:

dnnr:
I spent the last couple of hours implementing basically exactly what this PR aims to achieve. It was only after I was done, that it occurred to me to should check if someone else is already working on this. Finding this quite old but also very recently active PR was a real bittersweet surprise.

So, in an attempt to feel like I have not totally wasted my time, let me just say that this PR works perfectly for me (using herbstluftwm), including the weird corner cases that I ran into on the way (just like you did).

I'd very much like to see this merged. :+1:

If anyone's curious, this is what I came up with: https://github.com/dnnr/polybar/commit/45c3849b8499da960593a7411f06e62b3b0001bf
(I didn't get to test it in a multi-monitor setup, so I might have missed something.) I figured that keeping track of the occupied desktops was not really worth the effort and opted for cross-referencing the clients and desktops whenever rebuilding the desktop states. I also dropped the diffing of the client list, assuming that the benefit is negligible.

kronn:
@dnnr I like how you simplified the diffing by dropping it. As far as I remember, there was nothing special about multi-monitor (apart from #1849, but that was present before). Problems arose from a change in the number of desktops, as some WMs change those during runtime. 

I think we could take your dropped client-list diffing :wink: as it simplifies the code, but that decision should be made by maintainers who know the project longer. 

I am very curious how @Lomadriel reacts to this development (:smile: pun intended).

Lomadriel:
I confronted the two versions (which are both good), the implementation from @dnnr seems quite simpler than the one from this PR.

> As far as I remember, there was nothing special about multi-monitor (apart from #1849, but that was present before).

Currently, the multi-monitor is unusable (at least for me) but indeed it doesn't impact that feature.

> I think we could take your dropped client-list diffing wink as it simplifies the code, but that decision should be made by maintainers who know the project longer.

In my opinion, it's better that way. I'll add suggestion to your commit @dnnr to remove duplicate code, … but I think it could be a good idea to continue on your version but I would like the opinion on another person (maybe @NBonaparte since he started to review this PR) to confirm that we don't need to maintain a list of occupied desktop (for another future PR).

Thanks for you implication in polybar :smile:

kronn:
With regards to multimonitor: xmonad does not assign multiple workspaces to a monitor, therefore I did not encounter this bug. 

In ~soviet russia~ _xmonad_, each monitor can show one workspace, but the concrete assignment is flexible/dynamic and determined during runtime by user-interaction.

FWIW, as much as I would have like to have my name on this feature, I actually favour @dnnr 's version. It may need the `build_mutex` to be a complete replacement for this PR, but otherwise I am happy to yield.

Lomadriel:
> In ~soviet russia~ _xmonad_, each monitor can show one workspace, but the concrete assignment is flexible/dynamic and determined during runtime by user-interaction.

Good to know how it works in ~soviet russia~ _xmonad_. I don't really know how I'll fix that bug and it really bothers me :/

> It may need the build_mutex to be a complete replacement for this PR, but otherwise I am happy to yield.

Yeah there is that too. We really need to fix these concurrent access to avoid inconsistencies/crash, etc...

dnnr:
Thanks for the feedback. Unless there's no other way, I'd rather avoid complicating the discussion by opening a *third* PR addressing this topic. So if you don't mind, please feel free to just replace this PR's content with my version as you see fit. No need to credit me or anything.

Lomadriel:
> No need to credit me or anything.

I just pushed your commit, I credited you as the author since I don't want to steal your code ;)

I added some modifications to remove code duplicate and fix some concurrent access but again thanks for your contribution.

kronn:
Running the current version of the branch, I get a "laggy" behaviour. Sometimes when I switch the workspace, polybar does not reflect the current workspace correctly. When I switch the next workspace (or the same one again), polybar is updated again correctly. I vaguely remember that I had this before, but I did not search my previous commits (those may contain a reference to this if I fixed that explicitely). 

This does not seem to affect the occupied stage, as far as I can tell. So one might argue that it is a separate bug and would not block this one from being merged.

Lomadriel:
> Running the current version of the branch, I get a "laggy" behaviour. Sometimes when I switch the workspace, polybar does not reflect the current workspace correctly. When I switch the next workspace (or the same one again), polybar is updated again correctly. I vaguely remember that I had this before, but I did not search my previous commits (those may contain a reference to this if I fixed that explicitely).

Damn, I'll try to reproduce and fix that.

kronn:
> > I get a "laggy" behaviour. 

> Damn, I'll try to reproduce and fix that.

This might be a xmonad-issue. Also, I only had this before lunch, afterwards, it was fine. I did not, however, restart polybar or change the monitor-setup. Right now, it happened again. This time, I switched from docking-station to laptop-mode and therefore also restarted polybar.

Lomadriel:
> This might be a xmonad-issue. Also, I only had this before lunch, afterwards, it was fine. I did not, however, restart polybar or change the monitor-setup. Right now, it happened again. This time, I switched from docking-station to laptop-mode and therefore also restarted polybar.

I'll try under xmonad since I can't reproduce that issue under i3wm.

Lomadriel:
I'm unable to reproduce your issue. I think this issue might be linked a problem of concurrent access. :/

I may have a solution to debug that but I'll need your help.

When you have time, can you try to reproduce this issue while running polybar in [rr](https://github.com/mozilla/rr).
Use the following command: `rr record --chaos [polybar cmd]`.
Then run `rr pack` before modifying anything (like the config, the executable, …).
Zip the folder (it should be printed in your terminal when you run pack) and send it to me via a mail or whatever you want.

It may be possible that you can't reproduce this problem while running polybar into `rr` since it can't simulate multithread but that it's worth a try.

kronn:
I'll give it a try later today.

If the problem is not reproducible with rr, then I can just run it in rr all the time and solve it this way :joy:

kronn:
I have not encountered the laggy behaviour again, even without using rr.

kronn:
I have used several different monitor setups and not encountered the laggy behaviour anymore. If it comes up again, I would file a separate bug. From my point of view, this PR is therefore ready for a final review. 

Do you see #1849 as a hard prerequisite or is there anything else blocking this?

Lomadriel:
No I don't see #1849 as a prerequisite.
I'll review that asap.

dkasak:
Just wanted to say I've also been using the newest patch set for this in the last few weeks and things are working smoothly.

patrick96:
Awesome!

After two years we finally did it! :smile: 

Thank you very much @kronn, @dnnr and @Lomadriel for putting in time and effort and working together on this. We're very excited to have this in the repo.

:heart:

kronn:
You cannot imagine the joy this merge brings to me. In the end, very little of my code remains (so it feels), but I have the feeling that I helped with this features. Also, I can now switch back to master and get other features. :-)

jonhoo:
Excellent work everyone!

Is there documentation anywhere for how this feature can be used in styling?

patrick96:
The documentation was actually already there *before* the feature was implemented 😉 https://github.com/polybar/polybar/wiki/Module:-xworkspaces

jonhoo:
After building `polybar` with this merged, it looks like my list of workspaces is now always empty at startup time (i.e., when `polybar` is started by my WM). If I then manually restart it a bit later, the workspaces appear as usual. Is that expected behavior?

patrick96:
@jonhoo could you please open a new issue with this. This often happens when polybar is started too early, I don't think this PR causes it, but let's see.

jonhoo:
@patrick96 it's hard to reproduce reliably, but files https://github.com/polybar/polybar/issues/1915

PR #883, opened 2017-12-03
fix(aur): update to 3.1.0

NBonaparte:
Updates the polybar PKGBUILD to 3.1.0. Also adds `xcb-util-cursor` since this is a new release (I already pushed it to the AUR).

I think we should have a better way of updating the AUR package or the PKGBUILDs in this repo, so we only have to edit one of them and the other will be automatically updated. Should we have a bot or a script to do this for us?

### DIFF
--- a/contrib/polybar.aur/PKGBUILD
+++ b/contrib/polybar.aur/PKGBUILD
@@ -2,7 +2,7 @@
 # Contributor: Michael Carlberg <c@rlberg.se>
 pkgname=polybar
 pkgver=3.1.0
-pkgrel=1
+pkgrel=2
 pkgdesc="A fast and easy-to-use status bar"
 arch=("i686" "x86_64")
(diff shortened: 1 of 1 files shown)


patrick96:
Oh yeah, totally forgot about the new dependency.

It's kind of annoying to have to maintain two copies of the PKGBUILDs but I don't know if setting up a bot or script would actually take more time than it actually saves us.

contrib/polybar.aur/PKGBUILD
```diff
@@ -1,13 +1,13 @@
 # Maintainer: Michael Carlberg <c@rlberg.se>
 # Contributor: Michael Carlberg <c@rlberg.se>
 pkgname=polybar
-pkgver=3.0.5
-pkgrel=4
+pkgver=3.1.0
+pkgrel=1
```
patrick96: We should probably update the pkgrel

PR #894, opened 2017-12-07
fix(xbps): Update to 3.1.0

jaagr:


contrib/polybar.xbps/template
```diff
@@ -1,17 +1,16 @@
...
+_i3ipcpp_version=0.7.0
+_xpp_version=1.4.0
 build_style=cmake
 configure_args="
  -DENABLE_ALSA=$(vopt_if alsa ON OFF)
  -DENABLE_CURL=$(vopt_if curl ON OFF)
  -DENABLE_I3=$(vopt_if i3 ON OFF)
  -DENABLE_MPD=$(vopt_if mpd ON OFF)
  -DENABLE_NETWORK=$(vopt_if network ON OFF)"
-depends="cairo xcb-util-wm xcb-util-image xcb-util-xrm"
```
patrick96: Does this really work w/o the dependency key?

contrib/polybar.xbps/template
```diff
@@ -1,17 +1,16 @@
 # Template file for 'polybar'
 pkgname=polybar
-version=3.0.4
+version=3.1.0
 revision=1
-_i3ipcpp_version=0.6.2
-_xpp_version=1.3.6
+_i3ipcpp_version=0.7.0
```
patrick96: The 0.7.0 tag on i3ipcpp does not contain the fixes for the jsoncpp and the gcc with jsoncpp < 1.7.7 issues. I will shortly push the 0.7.1 tag so we can fix that

contrib/polybar.xbps/template
```diff
@@ -28,15 +27,25 @@ distfiles="
...
 
-if [ "$build_option_network" -a "$XBPS_TARGET_MACHINE" = "aarch64-musl" ]; then
-	CXXFLAGS+=" -D_LINUX_IF_ETHER_H"
+desc_option_alsa="Enable support for ALSA"
+desc_option_curl="Enable support for cURL"
+desc_option_i3="Enable support for i3"
+desc_option_mpd="Enable support for MPD"
+desc_option_network="Enable support for network"
+
+CXXFLAGS="-Wno-error=deprecated-declarations"
```
patrick96: Isn't this already in https://github.com/jaagr/i3ipcpp/commit/d54efb9b76bd0abeb7ac99db7e2d95909306e899?

contrib/polybar.xbps/template
jaagr: I just pointed to the versions that's specified in the `3.1.0` release. This is the template that is merged into the official void repository.

contrib/polybar.xbps/template
jaagr: Yes, because in void the `-devel` packages will include the ones removed here.

contrib/polybar.xbps/template
jaagr: I was not the one who added this to the xbps template so I don't know the reason for it. However, it's part of the merged template.

contrib/polybar.xbps/template
patrick96: Where did we specify a i3ipcpp version for the 3.1.0 release? I don't remember doing that.

I have now pushed the v0.7.1 tag pointing to the newest commit.

contrib/polybar.xbps/template
patrick96: Okay great. Do they also include the new `xcb-util-cursor` dependency? I have forgotten to add this one to the readme and the wiki, I will open a PR for it shortly.

contrib/polybar.xbps/template
jaagr: The submodule points to the `v0.7.0` tag in the i3ipcpp repository.

`git checkout 3.1.0 && cat .gitmodules`

contrib/polybar.xbps/template
jaagr: No. That is not included.

patrick96:
If I understand it correctly the [void-packages](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/polybar/template) repo is similar to the AUR. Since the main voidlinux template file is actually located over there, keeping the two copies in sync would be very time consuming, since, unlike with the AUR PKGBUILDs, anyone can submit new template files to it (without any of us doing anything). So we would have to manually update the template file everytime someone changed it over in the `void-packages` repo.

I suggest totally removing the template file from this repository and letting https://github.com/voidlinux/void-packages/blob/master/srcpkgs/polybar/template be the only template file. What do you think @jaagr

jaagr:
Yes, I agree.

PR #895, opened 2017-12-07
docs: Add xcb cursor dependency to README

patrick96:


NBonaparte:
IIRC I made `xcb-util-cursor` an optional dependency; should it be changed to a required dependency?

patrick96:
Technically it is optional because the functionality is turned off, if the macro is not set when building. But for example in the `PKGBUILD` `xcb-util-cursor` is a regular dependency. Do you think I should specifically mention that it is optional in the README?

NBonaparte:
It probably should be listed as optional to reflect the code.

patrick96:
I will add an optional dependencies section

patrick96:
I did a force push. Have a look at it and merge, if you think everything is in order.

PR #897, opened 2017-12-10
feat(temp): Add temperature tokens without unit

patrick96:
Adds two new tokens, `%temperature-c-n%` and`%temperature-f-n%`, to display the temperatures in those respectives scales without the appended unit.
This also deprecates the `%temperature%` token in favor of the scale specific tokens, `%temperature-c%` and `%temperature-f%`

### DIFF
--- a/doc/config.cmake
+++ b/doc/config.cmake
@@ -322,8 +322,8 @@ format-underline = #f50a4d
 format-warn = <ramp> <label-warn>
 format-warn-underline = ${self.format-underline}
 
-label = %temperature%
-label-warn = %temperature%
+label = %temperature-c%
+label-warn = %temperature-c%
 label-warn-foreground = ${colors.secondary}
 
 ramp-0 = 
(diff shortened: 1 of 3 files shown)


NBonaparte:
To be honest, this is starting to look a little messy. Would it be better if we had just a module setting like `units = celsius/fahrenheit/kelvin` and two tokens (`%temperature%` and `%temperature-n%`)?

patrick96:
Fair point. That would make it a lot cleaner. This would however make it impossible to have two different temperature scales in the same module, but I don't know how many people would see this as a valid use-case. Though I could imagine someone wanting to have celsius and fahrenheit side by side.

I'm not quite sure how to handle this, your suggestion would be a breaking change, so we would need to keep the old functionality plus the code to check for deprecated tokens until the next major release, which would be 4.0. So either way, the code will look messy.

I think we can afford the messiness for the added flexibility and even if we added tokens for the kelvin scale, it would only add another two tokens and that would be it (unless someone really wants their exotic temperature scale). 

What do you think?

NBonaparte:
How about a boolean setting `units = true/false` so we'll have half the number of tokens? I'm pretty sure that no one is going to try to have both versions at the same time.

patrick96:
That's a great idea, this wouldn't break anything and still be less messy. Also, if people really want both, they can add the unit to the label manually.

Issue #248, opened 2016-12-16
Provide default desktop names


Kintaro:
In ```get_desktop_names``` in the ```ewmh_util```, it would be better to provide some default names and use ```_NET_NUMBER_OF_DESKTOPS``` as count.
According to https://specifications.freedesktop.org/wm-spec/latest/ar01s03.html, the number of given names can be less than the actual number of desktops.

PR #904, opened 2017-12-15
xworkspaces: Use _NET_NUMBER_OF_DESKTOPS if _NET_DESKTOP_NAMES is not avaliable

topisani:
[windowchef](https://github.com/tudurom/windowchef) is at least one window manager that doesnt support `_NET_DESKTOP_NAMES`. This patch simply falls back to `_NET_NUMBER_OF_DESKTOPS`, and generates names from those numbers.

### DIFF
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -66,6 +66,7 @@ namespace modules {
     void rebuild_clientlist();
     void rebuild_desktops();
     void rebuild_desktop_states();
+    void rebuild_desktop_names();
     void set_desktop_urgent(xcb_window_t window);
 
     bool input(string&& cmd);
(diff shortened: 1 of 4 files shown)


patrick96:
Idea and code look solid. Will have to do some testing, might be a few weeks though.
Just out of curiosity, what is currently (not in this PR but in master) happening, if the WM does not support `_NET_DESKTOP_NAMES`?

NBonaparte:
We should be using `_NET_NUMBER_OF_DESKTOPS` anyways if we have less names than numbers (see #248). Feel free to implement that as well if you are willing.

topisani:
status on this?

patrick96:
Don't really have that much time on my hand, don't know about the other guys. For me right now, anything that is not a bugfix or a trivial PR has a quite low priority. We will get to it though :)

Also, sorry for the failed build, I wanted to try something out and because we substantially changed the travis config and the scripts it depends on, the old travis config will not run in the current repo, any sort of push will fix this (amend without changing anthing and force push, for example).

PR #910, opened 2017-12-17
feat(github): format-offline label

Kokan:
A new offline label feature for github module.
If the request toward github is not successfull,
now it is possible to change the displayed value via
format-offline which works as the same way as format.

### DIFF
--- a/include/modules/github.hpp
+++ b/include/modules/github.hpp
@@ -16,17 +16,22 @@ namespace modules {
 
     bool update();
     bool build(builder* builder, const string& tag) const;
+    string get_format() const;
 
    private:
     void update_label(const int);
     int get_number_of_notification();
     string request();
     static constexpr auto TAG_LABEL = "<label>";
+    static constexpr auto TAG_LABEL_OFFLINE = "<label-offline>";
+    static constexpr auto FORMAT_OFFLINE = "format-offline";
 
     label_t m_label{};
+    label_t m_label_offline{};
     string m_accesstoken{};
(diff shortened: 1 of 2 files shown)


patrick96:
Looks good. Will do some testing in the next few days, but I think this is ready to merge

Kokan:
I am using it since and I had no issue with it (This patch also applied: https://github.com/jaagr/polybar/pull/911).
@patrick96  Did you have time to check it ?

patrick96:
Sorry, I didn't have time yet. I'll try to get to it soon

Kokan:
Thanks for the feedback, I'll try to debug it later this week.

patrick96:
Hey @Kokan do you still plan on working on this? If not, we can add it to our own backlog and fix the last issue ourselves, I don't think it would be too bad.

Kokan:
I do not think I am going to fix it soon, as I am not using currently polybar.

patrick96:
Alright. We'll try to finish this when we get time for it.

kronn:
@patrick96 I tried to debug this and cannot reproduce the segfault anymore **after rebasing onto current master**. 

If `format-offline` is not specified at all, it defaults to an empty string (or so), but creates the module.
If `format-offline` does not contain `<label-offline>` but `<label>` it errors out of the module-creation with `error: Disabling module "github" (reason: <label> is not a valid format tag for "format-offline")`. 
A string without tokens is (as expected) taken as is without complaint.

I for one think this is sane behaviour and in working condition after a rebase.

Issue #720, opened 2017-09-02
Script click and scroll output


Eyenseo:
I had a look at the issues but didn't find any that matched my issue, if this is already possible or was discussed before it would be nice if you could give me a link.

It would be very nice if the output of the click and scroll actions would be used by the label of the script module. With that the update interval could be higher while the click would feel more responsive.

NBonaparte:
If I understood you correctly, this should be possible through the [ipc module](https://github.com/jaagr/polybar/wiki/Module:-ipc).

Eyenseo:
As far as I can see the ipc module does not have the interval option, which I would still want. I suspect that going over ipc would be over complicated if the trigger is polybar itself.

qqldd:
I vote for this. And this issue also requested for this:
https://github.com/jaagr/polybar/issues/84

NBonaparte:
This looks to be the same issue as #786. Closing in favor of that issue, as its title is more descriptive. Please reopen if I misunderstood this issue.

patrick96:
#786 wants the `exec` command to be rerun and this wants to display the output of the `click-*` commands inside the module.
I'm going to reopen this and add it to our TODO list

patrick96:
Thinking about this a little more, the main motivation for this issue was to have scripts be more responsive, maybe because the click actions produce the same output as `exec` itself. If it's anything else I really don't see a use case for it other than visual feedback. But then there would be the issue of when does that output get removed from the bar.

Updating the script module on click will be implemented as requested in #786 any other use case will not be implemented

PR #931, opened 2018-01-01
Added FreeBSD Support

kylecook80:
Added support for FreeBSD. Uses libinotify wrapper for inotify calls. I also had to disable some compiler warnings as libinotify uses c99 extensions.

Fixes #239.

PR #934, opened 2018-01-02
feat(script): %pid% token for tail commands

patrick96:
This can be used to send signals to the script running (e.g. SIGUSR1, etc) which
makes it easier to have an updatable state inside the script. With this, it is
for example possible to have a two-state date script:
```dosini
[module/date]
type = custom/script
exec = /path/to/date.sh
tail = true
click-left = bash -c "kill -SIGUSR1 %pid%"
```

`date.sh`:
```bash

t=0

toggle() {
    t=$(((t + 1) % 2))
}


trap "toggle" USR1

while true; do
    if [ $t -eq 0 ]; then
        date
    else
        date --rfc-3339=seconds
    fi
    sleep 1 &
    wait
done
```
Todo:
- [ ]  Add to wiki

### DIFF
--- a/include/modules/script.hpp
+++ b/include/modules/script.hpp
@@ -29,6 +29,8 @@ namespace modules {
 
     unique_ptr<command> m_command;
 
+    bool m_tail;
+
     string m_exec;
     string m_exec_if;
 
--- a/src/modules/script.cpp
+++ b/src/modules/script.cpp
@@ -174,39 +174,20 @@ namespace modules {
     string cnt{to_string(m_counter)};
     string output{module::get_output()};
 
-    if (!m_actions[mousebtn::LEFT].empty()) {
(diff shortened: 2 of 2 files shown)


NBonaparte:
Not exactly relevant to this PR specifically but is Travis having some issues? The clang build failed here as well as for #931 and I had to restart the builds for it to pass.

patrick96:
I've had it once before where the travis build would suddenly fail for no reason. I think it's just a minor hickup.

NBonaparte:
Does this fix #720?

patrick96:
> Does this fix #720?

No, it only adds a new token to the script module click commands nothing else. The output of a click command can still not be used inside the bar. However this makes it easier to notify a tail command about a click event, whereas before we would have to write to a certain file to share information with a tailed script.

NBonaparte:
I put the example in the wiki as part of the examples section.

PR #938, opened 2018-01-03
feat(i3): Add workspace separator

patrick96:
Puts a label-separator node between workspaces on the bar. Since the
separator uses a label it can be configured like any other label

To-Do after merge:

- [ ] Add to wiki

### DIFF
--- a/doc/config.cmake
+++ b/doc/config.cmake
@@ -152,6 +152,10 @@ label-urgent = %index%
 label-urgent-background = ${module/bspwm.label-urgent-background}
 label-urgent-padding = ${module/bspwm.label-urgent-padding}
 
+; Separator in between workspaces
+; label-separator = |
+
+
 [module/mpd]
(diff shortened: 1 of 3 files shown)


patrick96:
@NBonaparte seems the travis environment was updated recently ([blog](https://blog.travis-ci.com/2017-12-12-new-trusty-images-q4-launch)). The new image has clang 5 and cmake 3.9, I suspect one of those two are the culprit. If we can fix the build issue, we can remove all the code we use to setup cmake 3.8, otherwise we can revert to the old image with the `group: deprecated-2017Q4` setting.

NBonaparte:
We should probably add this feature to the bspwm module and/or the xworkspaces module as well.

patrick96:
I'm working on the bspwm version, gonna push tomorrow and I'll do the one for xworkspaces after that

PR #962, opened 2018-01-18
Fix #960

### LINKED ISSUE #960: defaults listed in build.sh are not accurate
When running build.sh, the script asks if you want to build various modules, for example, it will print  
`** Include support for "internal/i3" (requires i3) ------------------- [Y/n]:`

For each module, the default is said to be Y (denoted by the capital Y), but the module is only included if the user enters a 'y' and not included if the user presses enter (which should accept the default).

nwam:
This fixes #960 by making Y the default value. Modules will now only be excluded when the user answers N or n to prompts.


NBonaparte:
Why not make N/n the default instead so the functionality stays the same?

nwam:
I have made N be the default now, and be listed as the default.

PR #972, opened 2018-01-21
feat(config): Support fractional size and offset

### LINKED ISSUE #953: Support fractional height/width/offset values in bar configuration
Bar height, width, and offset-x/y are all currently configured either "as pixel value (e.g. 35) or percentage (e.g. 50%)". However, if a percentage value is given, only *integer* percentages can be used due to the use of `int` here

https://github.com/jaagr/polybar/blob/0a5ad9c76fe3f966b7f9f2fabfabea1a8eba05d5/src/components/bar.cpp#L251-L262

Unless there is a particularly good reason for this, I suggest this code be changed to allow fractional percentages to enable allows finer control. This is valuable in settings where an absolute value is not desired (e.g., when working with different DPIs; see #951).

jonhoo:
Fixes #953.


NBonaparte:
Sorry it took so long to merge!

PR #974, opened 2018-01-23
fix(config): Use proper token object

### LINKED ISSUE #971: Maxlen ignored - mpd
I tried setting both label-maxlen and specific label lengths but they are ignored..

ncmpcpp 0.8.1
mpd 0.20.15

image

patrick96:
When using the same token multiple times in the same label with
different length properties, polybar would always use the settings for
the first token. This now replaces the tokens one by one, so that a new
token object is used for the length settings each time a token is used
again. Because the token objects are added in order, the replacement
always uses the right token object for the token that is replaced.

Fixes #971

### DIFF
--- a/src/drawtypes/label.cpp
+++ b/src/drawtypes/label.cpp
@@ -47,14 +47,18 @@ namespace drawtypes {
     }
 
     for (auto&& tok : m_tokens) {
+      string repl{replacement};
       if (token == tok.token) {
-        if (tok.max != 0_z && string_util::char_len(replacement) > tok.max) {
-          replacement = string_util::utf8_truncate(std::move(replacement), tok.max) + tok.suffix;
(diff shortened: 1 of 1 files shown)


src/drawtypes/label.cpp
```diff
@@ -47,14 +47,18 @@ namespace drawtypes {
     }
 
     for (auto&& tok : m_tokens) {
+      string repl = replacement;
```
NBonaparte: Brace initialization could be used here (`string repl{replacement}`).

patrick96: Sure thing! What would be the advantage of brace initialization over the equal sign in this case?

NBonaparte: Mainly for consistency, as it is generally used in the rest of the code.

PR #976, opened 2018-01-27
fix(alsa): use correct name for deprecation

NBonaparte:
Not sure why I changed the module name... it broke all module options because it tried to find `internal/alsa` when `internal/volume` was specified. Now the options should be properly detected.


NBonaparte:
Also, what should we do about the wiki page for the volume module? Should we move everything to a new alsa page and link the volume page there?

patrick96:
I think we should rename everything to alsa, once the renaming makes it into a release, so probably when 3.2.0 is released.

PR #986, opened 2018-02-01
fix(bspwm): remove monitor focus command

### LINKED ISSUE #981: bspwm scrolling through desktops on multiple monitors
Does anyone else have an issue with scrolling through desktops on multiple monitors for the bspwm module? It always gets stuck on the boundaries between monitors, ending up alternating between the first of one and the last of another. 

Reverting d6b0c5f or commenting out https://github.com/jaagr/polybar/blob/5c7ff09ed64ab0f7e3dc5c2688642622430b7de2/src/modules/bspwm.cpp#L498-L503
seems to do the trick, but since this fixes #503, I guess the issue underlying it was fixed upstream. Before I fix this, could anyone else confirm/deny this?

NBonaparte:
I *think* #981 is an issue, but I need someone else to confirm it (both the problem and the fix).

Fixes #981.

### DIFF
--- a/src/modules/bspwm.cpp
+++ b/src/modules/bspwm.cpp
@@ -493,14 +493,14 @@ namespace modules {
 
     if (m_pinworkspaces) {
       modifier = ".local";
-    }
-
-    for (const auto& mon : m_monitors) {
-      if (m_bar.monitor->match(mon->name, false) && !mon->focused) {
-        send_command("monitor -f " + mon->name, "Sending monitor focus command to ipc handler");
-        break;
+      for (const auto& mon : m_monitors) {
+        if (m_bar.monitor->match(mon->name, false) && !mon->focused) {
+          send_command("monitor -f " + mon->name, "Sending monitor focus command to ipc handler");
+          break;
+        }
       }
     }
+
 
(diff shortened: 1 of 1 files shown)


NBonaparte:
bump

patrick96:
I'll try to reproduce this issue on my end, once I get some time. May not be this week, as I'm going on vacation from wednesday through sunday

patrick96:
Got some time to try to reproduce this. Here is what I got:

First of all, I assumed you had `pin-workspaces = false`, when you experienced the bug.

With polybar 3.1.0, the following happens: Scrolling through the workspaces, only scrolls through the ones on the monitor which the bar runs on and when it gets to a boundary, it focuses the workspace on the other monitor.
The question here is, what should be the expected behaviour? Should it only scroll through the workspaces on the monitor the bar is on, or should it scroll through all the workspaces in order?

I think the behaviour should be that it scrolls through all the workspaces it displays, this is what your patch does and it almost works properly for me.
If `pin-workspaces` is `true`, it scrolls through the workspaces on the selected monitor and not the workspaces on the monitor the bar runs on. This is quite unintuitive, I think also here it should scroll through the desktops on the bar monitor.

NBonaparte:
Probably.

PR #993, opened 2018-02-02
fix(battery): check without using path_battery

### LINKED ISSUE #985: battery module crashing
using version 3.1.0-38-g2a056d5, the battery module fails to load, and the stderr output is:
 
error: module/battery: std::bad_alloc
warn: Stopping 'module/battery'...

this is my config for the module: 

[module/battery]
type=internal/battery
label-discharging =  %percentage%%
full-at=99
battery=BAT0

this issue does not arise when using the binary shipped by my distro, (void linux), however that version does not seem to be built with pulseaudio support. 

has anyone else experienced this issue? any ideas on a solution?

NBonaparte:
Fixes #985 by not passing `path_battery` at all, which is the source of the error.


patrick96:
Seems like @krx beat you to it (#992)

NBonaparte:
I figured that passing `path_battery` is basically useless, so I removed it completely. #992 still passes it, albeit by copying.

PR #1009, opened 2018-02-09
feat(net):  Add nl80211 support

### LINKED ISSUE #277: Wireless-tools has been deprecated
Modern (newer) wifi drivers do not work with network tools, is there anything that prevents us from using nl80211 instead?

https://wireless.wiki.kernel.org/en/users/documentation/iw

taschenb:
This replaces the deprecated Wireless Extensions with [nl80211](https://wireless.wiki.kernel.org/en/developers/Documentation/nl80211).

Fixes #277


src/adapters/net.cpp
```diff
@@ -324,68 +350,131 @@ namespace net {
...
+  void wireless_network::parse_signal(struct nlattr** bss) {
+    if (bss[NL80211_BSS_SIGNAL_MBM] != nullptr) {
+      // signalstrength in dBm
+      int signalstrength = static_cast<int>(nla_get_u32(bss[NL80211_BSS_SIGNAL_MBM])) / 100;
+
+      // WiFi-hardware usually operates in the range -100dBm to -20dBm.
+      // For a more intuitive result, clamp the result to a
+      // signal range, where 100% are reachable without directly
+      // sitting on top of the AP
+      const int hardware_max = -40;
```
NBonaparte: IMO if we modify the values too much it might confuse users if they compare it to other sources.
From `nl80211_xbm_to_percent` it seems like -20 and -90 dBm are the commonly used values, so it might be best to use those instead.

NBonaparte:
@taschenb Do you have any interest in continuing this PR?

patrick96:
I think we should, for the time being, have both wireless_tools and nl80211 available side by side and have a config option to enable nl80211. This way we don't leave users with a broken bar, if something doesn't work.

src/adapters/net.cpp
```diff
@@ -270,24 +277,43 @@ namespace net {
...
-    if (iw_get_ext(*socket_fd, m_interface.c_str(), SIOCGIWMODE, &req) == -1) {
+    int driver_id = genl_ctrl_resolve(sk, "nl80211");
+    if (driver_id < 0) {
+      nl_socket_free(sk);
       return false;
     }
 
-    // Ignore interfaces in ad-hoc mode
-    if (req.u.mode == IW_MODE_ADHOC) {
+    nl_socket_modify_cb(sk, NL_CB_VALID, NL_CB_CUSTOM, scan_cb, this);
```
su8:
```c
if (0 != (nl_socket_modify_cb ...) {
handle_error
}
```

src/adapters/net.cpp
```diff
@@ -324,68 +350,131 @@ namespace net {
...
+
+      while (ies_len > hdr_len && ies[0] != WLAN_EID_SSID) {
+        ies_len -= ies[1] + hdr_len;
+        ies += ies[1] + hdr_len;
+      }
+
+      if (ies_len < hdr_len || ies_len < ies[1] + hdr_len) {
+        m_essid.clear();
+      } else {
+        m_essid = string(reinterpret_cast<char*>(ies + hdr_len));
```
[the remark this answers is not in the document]
su8: On top of that, ies can contain non-printable characters. So `isprint` should be used to test **ies**.

src/adapters/net.cpp
```diff
@@ -324,68 +350,131 @@ namespace net {
...
-      return;
-    }
+    auto status = nla_get_u32(bss[NL80211_BSS_STATUS]);
 
-    // Check if the link quality has been uodated
-    if (stats.qual.updated & IW_QUAL_QUAL_UPDATED) {
-      m_linkquality.val = stats.qual.qual;
-      m_linkquality.max = range.max_qual.qual;
-    }
+    return status == NL80211_BSS_STATUS_ASSOCIATED || status == NL80211_BSS_STATUS_IBSS_JOINED;
```
su8: `NL80211_BSS_STATUS_AUTHENTICATED` should be tested too.

su8: Also a **switch()** or **&** operator will be better to test against multiple cases.

taschenb:
I am deeply sorry for the late response.
I addressed the feedback from the code review. Thank you for that so far.

As for the coexistence with Wext, what approach would you suggest to select the interface to use? Would a simple CMake option or fallback-search for libnl in case libiw isn't found be sufficient for the time being? Or should we cleanup the network interface and expose a option in the network module?

NBonaparte:
A compile option should suffice. We should probably have libnl as the default and use libiw as a fallback in case the former isn't found.

taschenb:
In the latest update, nl80211 is now the preferred interface. In case libnl-genl-3.0 is not preset or -DWITH_LIBNL=OFF, it will fall back to Wext.

taschenb:
You are right, totally forgot about readding it. Is now in there as well.

NBonaparte:
A little nitpicky, but could you extend the other prompts to be the same length by adding more `-`s?

taschenb:
Good catch and not nitpicky. Everything should be and look consistent.
But the length of the prompt is still OK in general or should we try to limit this?

NBonaparte:
We could probably shorten it to `Include support for "internal/network" (requires libnl/libiw) -- [y/N]: ` unless there's an even shorter way.

src/adapters/net.cpp
patrick96: This just came up in #1834. Though [`nl80211_xbm_to_percent`](https://github.com/NetworkManager/NetworkManager/blob/master/src/platform/wifi/nm-wifi-utils-nl80211.c#L372) in the NetworkManager code says it goes from `-20` to `-90`, the way the percentage is calculated, it actually goes from `-120dBm` to `-20dBm`. `-90dBm` corresponds to 30%.

But I think the -20 to -90 range is good.

Issue #939, opened 2018-01-04
polybar hangs before displaying anything


rjmorris:
Polybar looks great from what I've seen, and I'm trying to set it up for the first time. This is on a 32-bit Debian machine running openbox. I compiled and installed successfully (as far as I know), but nothing happens when I run it. I eventually kill it with Ctrl-C. 

Output of `polybar -vvv`:

```
polybar 3.1.0

Features: +alsa +curl -i3 -mpd -network

X extensions: +randr (+monitors) -render -damage -sync -composite +xkb -xrm -xcursor

Build type: RelWithDebInfo
Compiler: /usr/bin/c++
Compiler flags:  -Wall -Wextra -Werror -Wno-noexcept-type -O2 -pedantic -pedantic-errors
Linker flags: 
```

Output of `polybar -l info example`:

```
* Loaded ~/.config/polybar/config
warn: No monitor specified, using "DVI-I-1"
* Loaded monitor DVI-I-1 (2560x1080+0+0)
* Bar geometry: 2560x35+0+0
* Starting application
* Starting module/date
* Entering event loop (thread-id=1)
* Eventqueue worker (thread-id=3)
```

At this point it seems to just hang and do nothing until I kill it. (For testing purposes, I removed all the modules except `date` from the config, so that's why it's the only module listed in the output.)

Any troubleshooting suggestions?

rjmorris:
Maybe something about it being a 32-bit system is the problem. I compiled on a 64-bit debian machine, and it runs fine there.

patrick96:
Sorry for the late response.
It's possible that this is indeed an issue with 32-bit. You could try to run polybar with `-s` (this does not display the bar on screen but prints its output to the terminal) and see, if it hangs there as well (and if it outputs anything).

Something else you can try is to debug with gdb, for that run polybar like this:
```bash
gdb --args CMD
```
Where you replace `CMD` with the command that you use to start polybar.
Once the gdb interface opens, type `run`. Once polybar hangs, kill it with Ctrl-C, you should now still be inside the gdb interface, type `bt full` and post the whole output here. This will help us determine where exactly polybar hangs.

patrick96:
After a lot of digging on this, I suspect it has something to do with the connection to the X server.
I don't think the backtrace you posted actually shows the problem, I think the event loop is actually supposed to stall. The problem is most likely in some other thread.

For this I need some more information, can you post the backtrace of all the threads with `thread apply all bt full` and run polybar with `-l trace`.

[...]

PR #1020, opened 2018-02-13
[WIP] fix(controller): Abort select call after 5 seconds

patrick96:
Without a timeout, the select call can possibly wait for an arbitrary
length of time. This helps detect these stalls (as reported in #939)

EDIT: Don't merge yet, I just noticed that a five second period without any event can also occur normally


patrick96:
Closing this, as we can't expect regular events from the event loop. I'll open a new PR with only the refactor commit

PR #1024, opened 2018-02-13
fix(mpd): Clear mpd error after getting error info

patrick96:
Before, polybar would crash, trying to throw a server error because
mpd_connection_get_server_error asserts that the error is of type server
error, but it isn't because it was cleared

### DIFF
--- a/src/adapters/mpd.cpp
+++ b/src/adapters/mpd.cpp
@@ -29,15 +29,26 @@ namespace mpd {
 
   void check_errors(mpd_connection* conn) {
     check_connection(conn);
+
+    string err_msg;
+
     switch (mpd_connection_get_error(conn)) {
       case MPD_ERROR_SUCCESS:
         return;
       case MPD_ERROR_SERVER:
-        mpd_connection_clear_error(conn);
-        throw server_error(mpd_connection_get_error_message(conn), mpd_connection_get_server_error(conn));
+        {
+          err_msg = mpd_connection_get_error_message(conn);
+          enum mpd_server_error server_err = mpd_connection_get_server_error(conn);
+          mpd_connection_clear_error(conn);
+          throw server_error(err_msg, server_err);
+        }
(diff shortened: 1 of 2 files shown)


src/adapters/mpd.cpp
```diff
@@ -29,15 +30,26 @@ namespace mpd {
...
+          err_msg = mpd_connection_get_error_message(conn);
+          enum mpd_server_error server_err = mpd_connection_get_server_error(conn);
+          mpd_connection_clear_error(conn);
+          throw server_error(err_msg, server_err);
+        }
       default:
-        mpd_connection_clear_error(conn);
-        throw client_error(mpd_connection_get_error_message(conn), mpd_connection_get_error(conn));
+        {
+          err_msg = mpd_connection_get_error_message(conn);
```
NBonaparte: Not used (called again for `client_error`)

src/adapters/mpd.cpp
```diff
@@ -29,15 +30,26 @@ namespace mpd {
 
   void check_errors(mpd_connection* conn) {
     check_connection(conn);
+
+    string err_msg;
```
NBonaparte: I don't think we need to create `err_msg`, just call `throw foo_error(mpd_connection_get_error_message(conn), err)` like in the default case.

src/adapters/mpd.cpp
```diff
@@ -1,4 +1,5 @@
 #include <cassert>
+#include <iostream>
```
NBonaparte: Unused

src/adapters/mpd.cpp
patrick96: I actually forgot to use `err_msg` in the default case. We need to temporarily store the message because `mpd_connection_clear_error` clears all error information.

patrick96:
I updated the commits, can't believe I missed so many things

PR #1048, opened 2018-02-23
xkeyboard: add icons for layouts and indicators

ozelis:
Added icon support for layouts and indicators in a similar way as it is done for i3 workspaces.

First of all a new `%icon%` token is available for the `<label-layout>` tag. Default icon (though not very useful) is specified with:

`layout-icon-default = some-icon`

and layout icons with  a `';'`-separated list of pairs

`layout-icon-0 = us;some-icon-1`
`layout-icon-1 = lt;some-icon-2`

The value on the lhs of `';'` will match `%layout%` token for a particular layout, eg. this list will work for layouts set up with

`$ setxkbmap -layout "us,lt" -option "grp:alt_shift_toggle"`

For indicators, `<label-indicator>` also gets `%icon%` token and two states: `'on'` and `'off'`. Default icons for indicators are specified with:

`indicator-icon-default = some-icon-1;some-icon-2`

where the first value is for the `'off'` state and the second is for the `'on'` state. Individual indicator icons can be specified with a list of `';'`-separated triples:

`indicator-icon-0 = caps lock;some-icon-1;some-icon-2`
`indicator-icon-1 = num lock;some-icon-3;some-icon-4`
`indicator-icon-2 = scroll lock;some-icon-5;some-icon-6`

Again, the first icon is for `'off'` state and the second one is for the `'on'` state.

`<label-indicator>` states can now be configured individually like this:

`label-indicator-on = %icon%`
`label-indicator-on-foreground = #0f0`
`...`

`label-indicator-off = %icon%`
`label-indicator-off-foreground = #f00`
`...`

If these states are not explicitly specified, `label-indicator-on` will inherit values from `label-indicator` and `label-indicator-off` will be empty so that existing user configs are not broken (who expect nothing to be shown when indicator is off). Otherwise values set for `label-indicator-*` are ignored (though the tag is itself is still named `<label-indicator>`).

Also `scroll lock` indicator was missing from the implementation even though it was mentioned in the wiki, so I added that. 

I've tested this with my old configs and a dozen or so other dotfiles i found on /r/unixporn and nothing breaks except maybe `scroll lock` indicator would now pop up if it was not blacklisted before (though every config I've seen have indicators disabled anyway).

Going forward I would also like to use a whitelist mode instead of a blacklist mode for two reasons:
1) New indicators (if they are added in the future) would not unexpectedly popup for users who don't have them blacklisted explicitly.
2) The order in which indicators are rendered depend on the order they are defined in `keyboard::indicator::type` `enum`. With a whitelist, we could render them in the order they appear in this list.

In case this gets merged there's an updated wiki page in my repo:
https://github.com/ozelis/polybar/wiki/xkeyboard

### DIFF
--- a/include/modules/xkeyboard.hpp
+++ b/include/modules/xkeyboard.hpp
@@ -21,6 +21,19 @@ namespace modules {
       : public static_module<xkeyboard_module>,
         public event_handler<evt::xkb_new_keyboard_notify, evt::xkb_state_notify, evt::xkb_indicator_state_notify>,
         public input_handler {
+   
+   enum class indicator_state {
+     NONE,
+     /**
+      * @brief Indicator is off
+      */
+     OFF,
+     /**
+      * @brief Indicator is on
+      */
+     ON
+   };
+
    public:
     explicit xkeyboard_module(const bar_settings& bar, string name_);
 
@@ -43,10 +56,10 @@ namespace modules {
     static constexpr const char* TAG_LABEL_INDICATOR{"<label-indicator>"};
     static constexpr const char* FORMAT_DEFAULT{"<label-layout> <label-indicator>"};
     static constexpr const char* DEFAULT_LAYOUT_ICON{"layout-icon-default"};
+    static constexpr const char* DEFAULT_INDICATOR_ICON{"indicator-icon-default"};
     
     static constexpr const char* EVENT_SWITCH{"xkeyboard/switch"};
-    
-    
+
     connection& m_connection;
     event_timer m_xkb_newkb_notify{};
     event_timer m_xkb_state_notify{};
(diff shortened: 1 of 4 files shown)


patrick96:
Great work, thank you!
Will need to find some time to review this.

sviatco:
Any chance for this being reviewed anytime soon? This is great stuff, essentially allowing to have a Mac-like behavior where languages are associated with country flags and displayed as such.

patrick96:
So sorry, this one must have slipped through the cracks.

@ozelis Would you still be around to maybe make changes after I reviewed this?

patrick96:
@TheFelixR work is being done over at #1559 since the author doesn't seem to be around anymore.

PR #1051, opened 2018-02-24
Toggleable labels

### LINKED ISSUE #781: Switch format in network module like the date module?
In the date module, you can specify a `date-alt` or `time-alt` when the date is clicked to switch between format. I want to do something similar, but with the network module so I'm able to click the module and get my IP. My current label is label-connected = %essid% and I would like something like label-connected-alt = %local_ip%
Thanks.

panurg:
I like feature of date module when you can switch two different formats of displayed information. It is very useful when you don't want to see some info all the time, e.g. date module may display current time usually and show date additionally by click.

There are two other modules which don't have input handlers by default: network and battery. They can implement same behavior, e.g. battery usually shows current percentage or nice ramp icon, but when you click on it then time to charge is shown additionally.

Fixes #781.

### DIFF
--- a/include/modules/battery.hpp
+++ b/include/modules/battery.hpp
@@ -2,11 +2,13 @@
 
 #include "common.hpp"
 #include "modules/meta/inotify_module.hpp"
+#include "modules/meta/input_handler.hpp"
 
 POLYBAR_NS
 
 namespace modules {
-  class battery_module : public inotify_module<battery_module> {
+  class battery_module : public inotify_module<battery_module>,
+                         public input_handler {
    public:
     enum class state {
       NONE = 0,
@@ -53,6 +55,7 @@ namespace modules {
     bool on_event(inotify_event* event);
     string get_format() const;
     bool build(builder* builder, const string& tag) const;
+    string get_output();
 
    protected:
     state current_state();
@@ -60,6 +63,7 @@ namespace modules {
     string current_time();
     string current_consumption();
     void subthread();
+    bool input(string&& cmd);
 
    private:
     static constexpr const char* FORMAT_CHARGING{"format-charging"};
@@ -70,8 +74,12 @@ namespace modules {
     static constexpr const char* TAG_BAR_CAPACITY{"<bar-capacity>"};
     static constexpr const char* TAG_RAMP_CAPACITY{"<ramp-capacity>"};
     static constexpr const char* TAG_LABEL_CHARGING{"<label-charging>"};
+    static constexpr const char* TAG_LABEL_ALT_CHARGING{"<label-alt-charging>"};
     static constexpr const char* TAG_LABEL_DISCHARGING{"<label-discharging>"};
+    static constexpr const char* TAG_LABEL_ALT_DISCHARGING{"<label-alt-discharging>"};
     static constexpr const char* TAG_LABEL_FULL{"<label-full>"};
+    static constexpr const char* TAG_LABEL_ALT_FULL{"<label-alt-full>"};
+    static constexpr auto EVENT_TOGGLE = "batterytoggle";
 
     static const size_t SKIP_N_UNCHANGED{3_z};
 
@@ -81,8 +89,11 @@ namespace modules {
     unique_ptr<consumption_reader> m_consumption_reader;
 
     label_t m_label_charging;
+    label_t m_label_alt_charging;
     label_t m_label_discharging;
+    label_t m_label_alt_discharging;
     label_t m_label_full;
+    label_t m_label_alt_full;
     animation_t m_animation_charging;
     progressbar_t m_bar_capacity;
     ramp_t m_ramp_capacity;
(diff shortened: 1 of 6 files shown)


NBonaparte:
Could you please fix the error in the Travis build?

panurg:
Done.

BTW, here is a little demonstration how this feature works on my laptop:
myimage

AichP:
Thats awesome supporting the merge!

rien333:
When can this be merged?

patrick96:
Sorry this is so late.

I quite like the idea of having toggleable labels everywhere (and many users do too). But the way we have to do it (and the way you do it as well), is to basically copy-paste the same code into every module to handle the click events, to add the action tags and so on. And this is just so much code reuse that I don't want to have it in the repo like this.

I think we would need a more generall approach to address this. Maybe something where that "alternative" functionality is directly built into the label drawtype or the formatter. What do you think @NBonaparte?

Another issue here is that the event names are only module type specific but if you have multiple modules of the same type, the click events always go to the same module (See [Known-Issues](https://github.com/jaagr/polybar/wiki/Known-Issues)), this would only increase the number of people that run into that issue. So I think we should only start to work on this, once this other issue has been fixed because it may need some deep cutting changes in how events are handled.

Thank you very much for contributing, I'm sorry this hasn't panned out yet.

patrick96:
toggleable modules would be an interesting feature. The issue I see here is that click sections inside those modules would override the click section that is used for toggling around the whole module.

dannycolin:
@patrick96 What about assigning (configurable by the user ofc) different click actions like double-click vs. single-click or right/left mouse click?

patrick96:
Hmm, so like double-clicking toggles the different modules and single-click does some internal action on the module (like in the date module)?

This would work, if the defaults are intuitive. But where would we put the configuration for toggling whole modules? This would then also hinge on us first fixing #1172

dannycolin:
> Hmm, so like double-clicking toggles the different modules and single-click does some internal action on the module (like in the date module)? 

Yes and those click actions could be configurable by the user. 

> This would work, if the defaults are intuitive. But where would we put the configuration for toggling whole modules?

To toggle the whole module, the settings should be in the bar settings section imho. We could also have general settings for the click actions. something like (don't pay to much attention to the naming i'm just brainstorming here):

```
whole-module-click-action = [single | double] [left | middle | right]
inside-module-click-action = [single | double] [left | middle | right]
```

> This would then also hinge on us first fixing #1172

Would it be possible to assign an ID to each instance of a module? So when you click on it polybar will know which module was triggered. Of course, that means more code will need to be changed.

patrick96:
> Would it be possible to assign an ID to each instance of a module?

Well, we will likely restrict it so that you can only have a single module with the same name. So you can't have two modules named `abc`, you can still have multiple modules of the same type though. The ID will likely be the module name in the config.

> To toggle the whole module, the settings should be in the bar settings section imho.

Yes, putting that inside a module wouldn't make much sense because there are always at least two modules involved.

I don't understand what does general settings for click actions are supposed to do, could you elaborate on that?

dannycolin:
> I don't understand what does general settings for click actions are supposed to do, could you elaborate on that?

Instead of hardcoding it like double-clicking toggles different modules and single-click toggles internal action on the module those settings will let the user set the binding they prefer for each action.

An example would be:
```
# To toggle between whole module 
whole-module-click-action = single left

# To toggle internal action on a module 
inside-module-click-action = double right
```

patrick96:
Oh, I see.

The `inside-module-click-action` would be a bit conflicting then. Since many modules already support specifying click actions for different mouse buttons. I think we could just leave out `inside-module-click-action` and just have `whole-module-click-action` which defaults to `double right`.

dannycolin:
> The inside-module-click-action would be a bit conflicting then. Since many modules already support specifying click actions for different mouse buttons.

That's true. Still it would be nice to consider making it configurable by the user in the future. :)

> Another idea would be to have a new module that can group modules together and always only display a single one.

Oh, I really like that idea. It looks clean and is similar to other options like `ramp`. I think we should go with that syntax.

patrick96:
> That's true. Still it would be nice to consider making it configurable by the user in the future. :)

It is already configurable by the user, they can set different click actions for each mouse button. Or am I misunderstanding you?

I have added a card to our [Feature Request Project](https://github.com/jaagr/polybar/projects/1#card-14839673)

dannycolin:
> It is already configurable by the user, they can set different click actions for each mouse button. Or am I misunderstanding you?

Oh nvm I misread your previous comment ;).

> I have added a card to our Feature Request Project

Cool! Thanks.

GreenGas48:
Guys is this PR still relevant? I would love this feature but it seems like this discussion died.

patrick96:
@GreenGas48 the PR itself isn't relevant anymore. We want to find a way where we don't have to duplicate the same code over and over for each module. Maybe once #1524 is implemented something like this could become easier to implement but don't hold your breath.

panurg:
Closed as per previous comments.

PR #1059, opened 2018-02-27
Changed loglevel to info

fbrinker:
See this as a suggestion, i didn't want to create an issue for such a small fix.
> Changed loglevel for "No matching input area found" from warn to info

In my opinion it is irritating when polybar emits a warning when i click with a not defined button on an element. I didn't define that click handler, so everything should be fine


patrick96:
Good idea

PR #1065, opened 2018-03-02
Some issues with syntax tags

ozelis:
## Problem

There is a serious issue with the way syntax tags (i.e %{..}) are parsed. Say you have `xwindow` module enabled that shows the title of currently focused window and that title contains some syntax tags. Eg. if a user navigates to some webpage that has the following contents:

`<html>`
`<head>`
`  <title>%{F#f00}red title%{F-}</title>`
`</head>`
`<body>`
`lorem...`
`</body>`

then her browser will set contents inside `<title>..</title>` as window title (+ name of a browser) and `xwindow` module will use that string as contents for a label. This will then go through `builder` where these syntax tags will be parsed and you will see a red "red title" in your bar. This is innocent enough but consider action tags, eg. if contents inside `<title>` are

`<title>%{A1:some_command:}click me%{A}</title>`

you will only see "click me" in your bar and upon left-clicking on it, a value in between `:`s will end up in `controller::process_inputdata()` where it will iterate through modules looking for the ones that implement `input_handler` class and then call their `::input()` method with `"some_command"` as argument. If no `input_handler`s will handle this command it will then execute `"some_command"` directly in a shell! This is *very* bad! Consider something like:

`<title>%{A1:touch ~/pwnd:}polybar warning: click here to dismiss%{A}</title>`

you'll find a file name `pwnd` in your home directory upon left-clicking on it.

Acctually I just noticed that as I was writing this PR. Originally I found a bug in `i3` module (which implements `input_handler` class) where its `::input()` method would look at the command and see if it starts with `"i3wm-wsfocus-"`. If so then it would strip away that part, prepend `"workspace "` and send it over IPC to i3. But i3 IPC accepts `';'`-seperated list of commands so if the command was something like `"i3wm-wsfocus-1;exec touch pwnd"` it would strip away `"i3wm-wsfocus-"` and send `"workspace 1; exec touch pwnd"` to i3. That was bad enough but this is way, way worse.

## Solution

The solution involves allowing `label`s to ignore syntax tags in their contents. From a user perspective you can now set "-notags" option on your label, eg:

`somelabel            = %{T1}some text%{T-} %some_token%`
`somelabel-foreground = #f00`
`somelabel-notags     = true`

so you can still style your labels with `-foreground`, `-background` etc. but in this case `%{T1}` and `%{T-}` would not be parsed and simply displayed as-is.

Implementing this turned out to be not so trivial because of the way all contents of modules in a bar are dumped into one big string and then parsed all at once, but I can try to walk through the changes I made so it is easier to review this.

So first stop is `drawtypes/label.hpp|cpp`. `label` class gets a new member `m_ignore_syntax_tags` which is set to `false` by default and it's constructor and `clone()` method  are modified accordingly. Also in `load_label()` function it reads "-notags" value from config file (again, defaulting to `false` if it's not set explicitly, so it's buisness as usual)

Next is `components/builder.hpp|cpp` and `components/types.hpp`. `enum class syntaxtag` gets a new tag, `syntaxtag::X`. The idea is to insert this tag just before the text that needs its syntax tags ignored. So eg. in

`%{X11}%{T1}a%{T-} rest of the text`

the next 11 characters after the closing brace of the `X` tag, ie. `%{T1}a%{T-}` would be parsed as-is.

Ok so `builder::node(const label_t& label, ...)` method now checks if label wants its syntax tags ignored and if so adds `%{X<skip_count>}` (where `skip_count` is length of labels text) just before calling `builder::node(string str, ...)` method. This is done in `builder::tag_open(tag, value)` just like for other tags. Now `builder::node(string str, ...)` checks if `builder` is in `m_ignore_syntax_tags` state and simply appends `str` to its output without going through all the shebang of parsing tags inside it.

Finally in `components/parser.cpp` `parser::parse(...)` method first looks for our special `%{X<skip_count>}` tag, parses the number inside it and sends `skip_count` characters after the closing brace to `parser::text(..)` method instead of `parser::codeblock(..)`.

Thats it, except not quite so. There are a few things that mess this all up. 

At the end of `builder::flush()` all `"%__"` in its output are replaced by `" "` so if there were `"%__"`s in a text we want to interpret as-is, they would be replaced by single character `' '` and now `%{X<skip_count>}` would also include text outside its intended range. So I moved the line that does it to the end of `builder::node(string str, ...)` instead. Logic behind it is that

`output = replace_all(a + b + c, "%__", " ");`

is essentially the same as 

`a = replace_all(a, "%__", " ");`
`b = replace_all(b, "%__", " ");`
`c = replace_all(c, "%__", " ");`
`output = a + b + c;`

except maybe if the last two characters of `a` were `"%_"` and the first character of `b` was `'_'` or for some other such contrived edge case.

Yeah but now `modules/text.cpp` breaks a little (?) bit because it circumvents `builder::node()` construction and does it's own thing, eventually expecting `builder::flush()` to replace `"%__"` by `" "`. So I help it a little by doing it at the end of `text_module::get_output()`. There are no `label`s in this module so `m_ignore_syntax_tags` is not in play anyway, but I still feel a little bit uneasy about this if some other class inherits this and then uses `label`s, etc.

Finally the last issue was in `components/controller.cpp` where in `controller::process_update()` method it concatenates all the outputs from all the modules in the bar and then replaces things like `"T-}{T1"` with just `"T1"` etc. and joins adjacent tags under same scope. That is

`%{F#0f0}%{B#00f}text%{B-}%{F-}`

becomes

`%{F#0f0 B#00f}text%{B- F-}`

thus again potentially messing up any text that follows our new `X` tag. 

Ok so semanticaly these operations don't change anything and the parser will handle either form the same way so I simply commented these lines out. I was worried a bit that things would get slower but I measured and its same same (in my case all these `replace_all` functions took 0.2ms compared to ~2ms it takes to render the whole bar!). Turns out `string_util::replace_all` is a bit inefficient if a thing you are replacing with is of different length than the thing you are replacing.

Finally in `modules/xwindow.cpp` I force the label to ignore syntax tags and we're done. My rice is safe!

### DIFF
--- a/include/components/builder.hpp
+++ b/include/components/builder.hpp
@@ -76,6 +76,8 @@ class builder {
 
   string m_background{};
   string m_foreground{};
+
+  bool m_ignore_syntax_tags{false}; // tagfix:
 };
 
 POLYBAR_NS_END
--- a/include/components/types.hpp
+++ b/include/components/types.hpp
@@ -37,6 +37,7 @@ enum class syntaxtag {
   R,  // flip colors
   o,  // overline color
   u,  // underline color
+  X   // tagfix: ignore syntax tags
 };
 
 enum class mousebtn { NONE = 0, LEFT, MIDDLE, RIGHT, SCROLL_UP, SCROLL_DOWN, DOUBLE_LEFT, DOUBLE_MIDDLE, DOUBLE_RIGHT };
--- a/include/drawtypes/label.hpp
+++ b/include/drawtypes/label.hpp
@@ -39,11 +39,12 @@ namespace drawtypes {
     side_values m_margin{0U,0U};
     size_t m_maxlen{0_z};
     bool m_ellipsis{true};
+    bool m_ignore_syntax_tags{false};
 
     explicit label(string text, int font) : m_font(font), m_text(text), m_tokenized(m_text) {}
     explicit label(string text, string foreground = ""s, string background = ""s, string underline = ""s,
         string overline = ""s, int font = 0, struct side_values padding = {0U,0U}, struct side_values margin = {0U,0U},
-        size_t maxlen = 0_z, bool ellipsis = true, vector<token>&& tokens = {})
+        size_t maxlen = 0_z, bool ellipsis = true, vector<token>&& tokens = {}, bool ignore_syntax_tags = false)
         : m_foreground(foreground)
         , m_background(background)
         , m_underline(underline)
@@ -53,6 +54,7 @@ namespace drawtypes {
         , m_margin(margin)
         , m_maxlen(maxlen)
         , m_ellipsis(ellipsis)
+        , m_ignore_syntax_tags(ignore_syntax_tags)
         , m_text(text)
         , m_tokenized(m_text)
         , m_tokens(forward<vector<token>>(tokens)) {}
--- a/src/components/builder.cpp
+++ b/src/components/builder.cpp
@@ -19,6 +19,7 @@ builder::builder(const bar_settings& bar) : m_bar(bar) {
   m_tags[syntaxtag::T] = 0;
   m_tags[syntaxtag::o] = 0;
   m_tags[syntaxtag::u] = 0;
+  m_tags[syntaxtag::X] = 0;
 
   m_colors[syntaxtag::B] = string();
   m_colors[syntaxtag::F] = string();
@@ -58,6 +59,10 @@ string builder::flush() {
     cmd_close();
   }
 
+  if(m_tags[syntaxtag::X]) {
+    tag_close(syntaxtag::X);
+  }
+
   string output{m_output};
 
   // reset values
@@ -66,7 +71,9 @@ string builder::flush() {
   m_output.clear();
   m_fontindex = 1;
 
-  return string_util::replace_all(output, BUILDER_SPACE_TOKEN, " ");
+  // tagfix: do this at the end of builder::node(..) instead
+  // return string_util::replace_all(output, BUILDER_SPACE_TOKEN, " ");
+  return output;
 }
 
 /**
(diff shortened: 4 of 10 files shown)


NBonaparte:
This is a serious vulnerability, good catch.

I think I've spotted an issue with this solution, however. Suppose you have a label:
```
label = %{X7}%title% %{F#00F}other_stuff%{F-}
```
When `%title%` is expanded, the parser does not know how long it will be. Thus, if `%title%` is, let's say, `moretext %{A1:command:}click%{A}` it will still be clickable. The skip count also cannot be set to a high value because the other stuff still uses tags.

ozelis:
My idea is that this tag should not be used like this directly. It is inserted automatically with correct size in `builder::node()` after the whole label is already expanded (and if it has `-notags` set). See i first tried to do something like `%{X}%title%{X-}` but this wouldn't work because you could escape its scope, eg. if

`%title%=%{X-}%{A1:command:}click me%{A}%{X}`

so it expands to

`%{X}%{X-}%{A1:command:}click me%{A}%{X}%{X-}`

patrick96:
Neat solution with a tag to ignore all other tags. Mine would just have been to somehow escape all tags within the `%title%` token of the xwindow module.

However, I'm not a fan of using the length in the tag, this leaves too much room for error whenever we work on that part of the bar later on, you already saw that when you needed to make some code changes because of text replacement here and there. The only other safe way I see here, is to use tag pairs with random numbers attached to them, for example something like `%{X1234cdef}...%{X1234cdef}`. Here I used 32bit numbers represented in hex, which I think should suffice. An attacker would need to guess the random number that was used to be able to properly set the end tag early. Also important here is that as soon as an X tag was read, no other tag (especially no other X tags) should be interpreted, otherwise an attacker could disable the tags for the rest of the bar, by just putting a start X tag in there.
What do you guys think about that?

patrick96:
Hey @ozelis are you still around for this?

patrick96:
Closing in favor of #1704

PR #1075, opened 2018-03-03
feat(pulseaudio): port icons

### LINKED ISSUE #1071: Pulseaudio get active port
I used to use a custom script with Lemonbar to display volumes, where I had a ramp-volume / mute icon as well as an icon that represented whether the active port of the current sink was set on "analog-output-speaker" or "analog-output-headphones". Is there any way to get this functionality (with a bar, ramp icon, etc.) without using a script?

NBonaparte:
Fixes #1071 by adding icons for headphones, speakers, and HDMI ports using the tag `%port-icon%`.


araspik:
In `src/.../pulseaudio.cpp:102`, shouldn't it be `m_label_muted`?
Haven't looked at your source code, but that line seems suspicious.

NBonaparte:
Thanks, I forgot to change it.

araspik:
I'm testing this now ...
Results:

NBonaparte:
I've changed it to `<icon-port>` to be consistent with other icon tags. The options are now `icon-headphones`, `icon-speaker`, and `icon-hdmi`.

src/adapters/pulseaudio.cpp
```diff
@@ -293,6 +300,8 @@ void pulseaudio::sink_info_callback(pa_context *, const pa_sink_info *info, int
   if (!eol && info) {
     This->m_index = info->index;
     This->s_name = info->name;
+    if (info->active_port)
+      This->p_name = info->active_port->name;
```
patrick96: Please use curly brackets even for single line if blocks

araspik:
Thanks so much! It works perfectly!

bbaserdem:
I was doing this in a custom script, this is very well appreciated.

Can bluetooth sink detection be added as well for custom bluetooth icons?

NBonaparte:
Is the sink called `a2dp_sink`? I'm not too familiar with Bluetooth devices in Pulseaudio, but if the sink or port name is different, we could parse for that and apply a different icon.

NBonaparte:
I think the easiest way to implement this is to check if the sink name has `a2dp_sink` and then define separate icons for each port. 
From [here](https://github.com/pulseaudio/pulseaudio/blob/f6fe411b32c0cf5932fb4f169f5288c76bc6923d/src/modules/bluetooth/module-bluez5-device.c#L207) I'm assuming the possible port names would contain `headset`, `handsfree`, `speaker`, `headphone`, `portable`, `car`, `hifi`, and `phone`. (Is anyone running polybar in their car?)

NBonaparte:
@patrick96 do you happen to have Bluetooth audio devices which can be used to test this?

patrick96:
I do have bluetooth devices and my computer even has bluetooth but I wasn't able to get them to pair, let alone actually let the device act as a sink. I also don't have that much time to do testing, when I have time I reply to issues and try to fix bugs, new features aren't a top priority right now for me. But I will try to find a free weekend to do some testing in the near future.

NBonaparte:
It would be great if you could test icons for as many device types as possible.

Mange:
So I had some issues. Here is the config I used:

```
[module/volume]
inherit = module/base
type = internal/pulseaudio

format-volume = <bar-volume> <icon-port> <label-volume>
format-muted = <label-muted> <icon-port>
# ... snip ...
# icon-port = "icon-port"
icon-headphones = "icon-headphones"
icon-speaker = "icon-speaker"
icon-hdmi = "icon-hdmi"
icon-bluetooth-headset = "icon-bluetooth-headset"
icon-bluetooth-handsfree = "icon-bluetooth-handsfree"
icon-bluetooth-speaker = "icon-bluetooth-speaker"
icon-bluetooth-headphones = "icon-bluetooth-headphones"
icon-bluetooth-portable = "icon-bluetooth-portable"
icon-bluetooth-car = "icon-bluetooth-car"
icon-bluetooth-hifi = "icon-bluetooth-hifi"
icon-bluetooth-phone = "icon-bluetooth-phone"
```

(I based this on all listed icons in this PR.)

## Problems

1. It seems that changing sink does not refresh the `pulseaudio` module at all. I had to restart polybar every time I changed output. Am I doing something wrong? Not really related to this specific branch, but might turn out to be useful for other testers, maybe.
2. My speakers are using a "line-out" port on my sound card (normal analog 3.5 stereo plug), which does not have an icon as far as I can tell. This caused me to try to debug my icon settings for a long time before I realized.

```
	Ports:
		analog-output-lineout: Linje ut (priority: 9900, available)
		analog-output-headphones: Hörlurar (priority: 9000, not available)
	Active Port: analog-output-lineout
	Formats:
		pcm
```

I also have a analog headphone jack on the front of the tower, which is identified as `analog-output-headphones` here. That one would automatically get the `icon-headphones` icon.

## Usability suggestions

This is a lot of icons to add. Would it be possible to have defaults for them such that there is just a few basic icons you could set, and then override more specific icons if you wanted.

Maybe something like this:

```
icon-bt-speaker = <icon-speaker>
icon-bt-headset = <icon-headphones>
icon-bt-handsfree = <icon-headphones>
icon-bt-headphones = <icon-headphones>
icon-bt-hifi = <icon-speaker>
icon-bt-phone = <icon-other>
icon-bt-cat = <icon-other>
icon-hdmi = <icon-other>
```

Then you only have to set `icon-speaker`, `icon-headphones`, `icon-other`  to get an icon in all cases, and then you may override a single one that you care about specifically.

doronbehar:
What is holding this PR from being merged? I think some of the problems reported in https://github.com/jaagr/polybar/pull/1075#issuecomment-386836196 were fixed in https://github.com/jaagr/polybar/issues/1167 and obviously the [last requested change](https://github.com/jaagr/polybar/pull/1075/files/79eca4803b4b4a76ed7128db1d63eaf811f3c3fc#diff-bc4c51096e0f8cb63d028420a2422671) is easily fixable. I've tested it on my machine and it works fine for me.

Mange:
I'd love to help out more if needed.

NBonaparte:
I believe I fixed this in ec7b528 (although it's been a while), could anyone test it out?

doronbehar:
I've tested it after I ran `git rebase master`, it works just great.

One last thing I think we need is adding Unicode symbols for the following variables in the [example configuration file](https://github.com/jaagr/polybar/blob/master/doc/config.cmake)?
```
icon-bt-speaker
icon-bt-headset
icon-bt-handsfree
icon-bt-headphones
icon-bt-hifi
icon-bt-phone
icon-bt-cat
icon-hdmi
```

araspik:
Bump: Is there anything holding this back? Maybe @patrick96 could take down the earlier change that is being alerted for. There doesn't seem to be any conflicts as far as I can see. Let's merge this!

NBonaparte:
@patrick96 This is ready to be reviewed as well, should be finished except for the example config (can't remember what font is used for the icons).

doronbehar:
Hey guys, I would really like this feature to be merged, As far as I understand, the only missing piece is an addition to the example configuration file that will include pulseaudio ports icons.

I have found many icons which may suit our needs at https://material.io/tools/icons/ . But this will require us to suggest this font in the other documentation.

What do you think?

src/modules/pulseaudio.cpp
```diff
@@ -119,6 +145,34 @@ namespace modules {
...
+        } else if (string_util::contains(port, "phone")) {
+          builder->node(m_port_icons->get("bt-phone"));
+        }
+      } else if (string_util::contains(port, "headphones")) {
+        builder->node(m_port_icons->get("headphones"));
+      } else if (string_util::contains(port, "speaker")) {
+        builder->node(m_port_icons->get("speaker"));
+      } else if (string_util::contains(port, "hdmi")) {
+        builder->node(m_port_icons->get("hdmi"));
+      }
```
patrick96: should we display `icon-other` if we can't match anything?

patrick96:
@doronbehar I don't think we should add more fonts to the default config it would only be another thing we need to support. I think, if we add them to the default config, they should be commented out with plain text labels or icons from siji

doronbehar:
So what is currently holding back this feature from being merged? I would really like to commit my changes to my configuration and use my systems `polybar` executable.

src/modules/pulseaudio.cpp
```diff
@@ -119,6 +145,36 @@ namespace modules {
       builder->node(m_label_volume);
     } else if (tag == TAG_LABEL_MUTED) {
       builder->node(m_label_muted);
+    } else if (tag == TAG_ICON_PORT) {
+      auto port = m_pulseaudio->get_port_name();
+      auto sink = m_pulseaudio->get_name();
+      if (string_util::contains(sink, "a2dp_sink")) {
```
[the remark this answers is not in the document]
patrick96: Also that function should log which icon it has chosen.

Mange:
I like that idea. Icons would then be consistent with PA GUIs too, which is a plus.

NBonaparte:
I originally considered icon names but found out that not all ports have icon names, so I switched to using port names instead.

patrick96:
@NBonaparte When I suggested this, I was thinking about using `device.icon_name` property of the sink which is set [here](https://github.com/pulseaudio/pulseaudio/blob/b4a36453da7647ece7797dcca1628f7878aab870/src/pulsecore/sink.c#L3427) and should be available for every sink. It takes into account which bus it runs on, which helps us differentiate between `isa`, `pci`, `usb`, `firewire`, and `bluetooth`. It also has information about the form-factor, except for `car`, `hifi`, and `internal`.

I know those `device.icon_name`s are only sink specific. If we tried to match the `device.icon_name` property for both the current port and current sink what devices are you afraid would fall through the cracks?

I have found that the `alsa` module sets `icon_name` for quite a few ports as well: 
```
src/modules/alsa/mixer/paths/hdmi-output-4.conf
7:device.icon_name = video-display

src/modules/alsa/mixer/paths/analog-output-speaker-always.conf
26:device.icon_name = audio-speakers

src/modules/alsa/mixer/paths/analog-output-headphones.conf
25:device.icon_name = audio-headphones

src/modules/alsa/mixer/paths/hdmi-output-2.conf
7:device.icon_name = video-display

src/modules/alsa/mixer/paths/hdmi-output-5.conf
7:device.icon_name = video-display

src/modules/alsa/mixer/paths/hdmi-output-1.conf
7:device.icon_name = video-display

src/modules/alsa/mixer/paths/hdmi-output-0.conf
7:device.icon_name = video-display

src/modules/alsa/mixer/paths/analog-output-headphones-2.conf
25:device.icon_name = audio-headphones

src/modules/alsa/mixer/paths/analog-input-mic.conf.common
21:device.icon_name = audio-input-microphone

src/modules/alsa/mixer/paths/hdmi-output-7.conf
7:device.icon_name = video-display

src/modules/alsa/mixer/paths/hdmi-output-3.conf
7:device.icon_name = video-display

src/modules/alsa/mixer/paths/analog-output-speaker.conf
25:device.icon_name = audio-speakers

src/modules/alsa/mixer/paths/hdmi-output-6.conf
7:device.icon_name = video-display
```

So on my desktop it would be possible to uniquely identify almost all ports. Except for when I manually switch to `iec958-stereo-output`, I don't actually have a device that uses that, so I don't know if it would change, if I attached such a device.
On my laptop I can uniquely identify all ports.

NBonaparte:
The ports `analog-output-lineout` and `analog-output` don't have any icons on my setup. Also, bbaserdem's example with `headset-output` doesn't have an icon either.

Another annoying thing which led me to use ports: [`pa_sink_port_info`](https://freedesktop.org/software/pulseaudio/doxygen/structpa__sink__port__info.html) doesn't have a property list; the only way I found to retrieve it is to get `pa_sink_info->card`, then retrieve through `pa_context_get_card_info_by_index()` and search through the list of ports (`pa_card_info` doesn't even have an `active_port` pointer). I can now confirm this works though, so it is an option although unwieldy.

With this alternative setup, we'd have to look at sink icons for Bluetooth (since the port in bbaserdem's example doesn't have an icon), and then fall back to port icons for alsa, then to port names if there aren't any icons assigned. This is better structured, but we'd have to ensure that we catch all of the possibilities.

Some things that need to be answered:
* Is there any standard list for the XDG icon naming spec?

* How do other pulseaudio interfaces handle this? All the icon themes I've seen don't have any bluetooth-specific icons, so I assume the applications using them just cut off the "bluetooth" at the end and treat them as a regular device.

patrick96:
In bbaserdem's example the sink has the property though. And I'm sure for you that's the same.

>With this alternative setup, we'd have to look at sink icons for Bluetooth (since the port in bbaserdem's example doesn't have an icon), and then fall back to port icons for alsa, then to port names if there aren't any icons assigned. This is better structured, but we'd have to ensure that we catch all of the possibilities.

My idea was to use the port `device.icon_name` if it's defined, otherwise use the active sink's `device.icon_name` because it is defined for all sinks. I don't think I completely understand the motivation behind differentiating between alsa and bluetooth. My approach should work regardless of the API used.

NBonaparte:
That's almost what I meant, I just used it in reverse order.

The problem with using sink icons is that Bluetooth sinks apparently describe the device form factor (headsets, etc.) while ALSA sinks only describe the card (`audio-card-pci`). If our output is `analog-output` which has no port icon, `audio-card-pci` doesn't help at all.

If we instead checked for Bluetooth, we can use a separate fallback for non-Bluetooth ports and use the port name if no port icon name is available.
```
 if Bluetooth:
  icon_name = sink.icon_name //audio-headset-bluetooth
else:
  if port.icon_name:
    icon_name = port.icon_name //audio-headphones
  else:
    icon_name = name_to_icon(port.name) //analog-output
```
vs.
```
if port.icon_name:
  icon_name = port.icon_name //audio-headphones
else:
  icon_name = sink.icon_name //audio-headset-bluetooth, audio-card-pci
```

patrick96:
I forgot to answer that question previously. Yes there is an xdg icon naming spec ([here](https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html)). The pulseaudio maintainer actually wanted to get some names added to the standard but got rejected and so just used his custom names ([here](https://lists.freedesktop.org/archives/xdg/2009-May/010397.html)).

From my viewpoint, the `device.icon_name` property doesn't really have to be descriptive, it just has to be unique to that machine. I guess my point would be that the use case where someone has two audio configuration on the same machine that both only have `icon_name = audio-card-pci` for which they would want different icons in the bar, is so rare that it won't happen in practice. Because with alsa, it provides the icons on the port for many other configurations, so headphones with alsa would actually have the `icon_name` property on the port.

I think we can always implement more and more logic to allow the user to target all their audio configuration very precisely, but it will never be perfect, will add more complexity to the code and will be more difficult for the user to configure.

For me this is a tradeoff between ease of use and expresiveness.

Mange:
Is there any update to this? What are we waiting for? (Acting on feedback, making decisions, a new review, or something else?)

patrick96:
@Mange We still need to decide on how users would identify the different ports in the config. And afterwards of course we'd also have to make the appropriate changes in the code.

Mange:
> For me this is a tradeoff between ease of use and expresiveness.

I assume most users "just" want to be able to have an icon or text label for headset, bluetooth headset, and normal speakers (and possibly for hdmi and bluetooth speaker but that wouldn't be as common).

If there is one solution that makes that easier for users to do, then I think that is the solution that should be preferred even if it's not as expressive.

If they need it to be even more expressive they could use a `script` as an escape hatch instead.

doronbehar:
Too bad this wasn't merged in the last release.. I've been waiting for so long to see it and I hoped I wouldn't have to compile it on my own rebased branch.

doronbehar:
Since this PR is very old but it provides a useful feature, I just wanted to let anyone subscribed that I'm maintaining a fork with it merged. Every release, I rebase the original PR against master and push to the branch `pulseaudio-port` in my fork:

https://github.com/doronbehar/polybar/tree/pulseaudio-port

The last release's rebase wasn't trivial due to some functions which changed their names so I felt it might worth mentioning here my small efforts.

As I said in my previous comment, it's too bad this wasn't merged on time. Not to mention the other very useful PRs waiting to be merged. Thanks again @NBonaparte for making this possible in the first place.

PR #1077, opened 2018-03-04
feat(net): Add unknown-as-up option

### LINKED ISSUE #457: Consider network link state UNKNOWN as UP
Hi, this is a feature request, when I use usb modems / or tethering via usb the state of the interface appears  as UNKNOWN, even trying to put it up with "ip link set" doesnt change the state but the interface works : 

`enp0s20u1: <BROADCAST,MULTICAST,DYNAMIC,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UNKNOWN group default qlen 1000
`

So maybe a property true/false for network module "unknown-as-up" could be useful.

NBonaparte:
Fixes #457

### DIFF
--- a/include/adapters/net.hpp
+++ b/include/adapters/net.hpp
@@ -68,6 +68,7 @@ namespace net {
     string ip() const;
     string downspeed(int minwidth = 3) const;
     string upspeed(int minwidth = 3) const;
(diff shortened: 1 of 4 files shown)


patrick96:
Added to wiki

PR #1079, opened 2018-03-05
refactor(clang-tidy,format): apply fixes

NBonaparte:
The code hasn't been cleaned for a while so I ran clang-tidy and clang-format. Most of it is just reordering of includes, commenting namespaces, and various other formatting stuff. There were a few parts where I refactored the code a little as suggested by clang-tidy.

For future reference, these can be run with `cmake --build build/ --target {codeformat,codecheck,codecheck-fix}` (format, tidy, format + tidy).

### DIFF
--- a/include/adapters/alsa/control.hpp
+++ b/include/adapters/alsa/control.hpp
@@ -36,6 +36,6 @@ namespace alsa {
     snd_hctl_t* m_hctl{nullptr};
     snd_hctl_elem_t* m_elem{nullptr};
   };
-}
+}  // namespace alsa
 
 POLYBAR_NS_END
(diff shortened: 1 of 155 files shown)


patrick96:
I'm totally for this, but right now this will create a merge conflict for every open PR, and after they rebase it will be impossible to review the PR because this commit will show up in the PR.
I suggest we try to merge as many PRs as possible until 4.0.0 at which point we can make a clean break.

I think, after this we should even add a travis job to just run clang-tidy (maybe with less strict rules though).

NBonaparte:
I'll just close this for now.

Issue #762, opened 2017-09-17
i3 module frozen after restarting i3


Metamist:
As the title suggest, once I restart i3 either through the keybinding, or with `i3-msg restart`, the i3 module stops working without any sort of error or log entry.
I'm running Archlinux, by the way.
Clicking on the workspaces in the bar still works, but the indicator no longer updates.
It seems like it is only the visual indicators that stops updating. I have tried this with the stock config found in `/usr/share/doc/polybar/config`, and the issue still happens.
I tried with both polybar and polybar-git from AUR, but the issue remains.
Currently using `polybar-git`, here's the output of `polybar -v`
```
metamist ~ (master*) $ polybar -v
polybar 3.0.4

Features: +alsa +curl +i3 +mpd +network
```

patrick96:
It seems you are not using the lates version of polybar. Please update the newest version of `polybar-git` and check if this still happens.

Metamist:
I literally installed it from AUR moments before posting the issue. Is the AUR version not up to date?

NBonaparte:
It should be on version 3.0.5 if you installed it recently.

Metamist:
So it seems like pacman is not installing the right version. Whenever I try to reinstall, it is telling me that I am on version `3.0.5.r59.g4663d01-1` and that there is an update (Yaourt is telling me the latest is 3.0.5-3). However, I can't update for some reason, it always ends up at the version I wrote before.

koppa:
I experience the same issues. I am currently running the master branch of polybar and i3-gaps 4.14.1. 
Increasing the log level to ```info``` gives no hints to what is happening.

patrick96:
I was able to reproduce this.
I think this happens because polybar connects to i3 using `i3ipc::connection` only at startup and when i3 is restarted like this that connection breaks without either i3ipc being notified, i3ipc notifying polybar, i3ips handling it correctly or polybar handling it correctly. I don't know which one of these applies, as I've never touched that part of polybar. 
But what you can do to fix this is to use
```bash
exec_always launch_polybar.sh
```
in your i3 config and have that script restart the bar. This way the bar is restarted everytime i3 is.

Metamist:
After digging around a little bit and testing some things out, I found out that the issue is most likely just like you say. There is a try catch statement that is hiding the actual exception that occurs after the connection drops, which is
```
error: Unexpected EOF while reading header
```
[...]

PR #1099, opened 2018-03-15
fix(i3): Reconnect i3 IPC socket on restart/error (#762)

Holzhaus:
This fixes issue #762 by reconnecting the socket if an error occurs.


patrick96:
Works fine for me too. Thanks!

PR #1102, opened 2018-03-15
feat(disk_io): added disk I/O module

vilkoz:
Added disk IO speed module
At the moment the module shows sum of read / write speds of all block devices.
Improvement suggestions are welcome.

### DIFF
--- a/doc/config.cmake
+++ b/doc/config.cmake
@@ -394,6 +394,14 @@ menu-2-0-exec = sudo poweroff
 menu-2-1 = cancel
 menu-2-1-exec = menu-open-0
 
+[module/disk_io]
+type = internal/disk_io
+interval = 2
+format-prefix = " "
+format-prefix-foreground = ${colors.foreground-alt}
+format-underline = #4bffdc
+label = R: %speed_read% Mb/s W: %speed_write% Mb/s
+
 [settings]
 screenchange-reload = true
 ;compositing-background = xor
--- a/include/modules/disk_io.hpp
+++ b/include/modules/disk_io.hpp
@@ -1,0 +1,44 @@
+#pragma once
+
+#include <chrono>
+#include "modules/meta/timer_module.hpp"
+#include "settings.hpp"
+
+POLYBAR_NS
+
+namespace modules {
+  /* enum class memtype { NONE = 0, TOTAL, READ, WRITE, SHARED, BUFFERS, CACHE, AVAILABLE }; */
+
+  class disk_io_module : public timer_module<disk_io_module> {
+    public:
+      explicit disk_io_module(const bar_settings&, string);
+
+      bool update();
+      bool build(builder* builder, const string& tag) const;
+
+    private:
+      std::vector<std::string> _get_disk_names(void);
+      std::pair<unsigned long long, unsigned long long> _get_disk_read_write(std::string);
+      void _calculate_disk_io_speeds(std::string);
+      float _get_time_delta(void);
+
+      const static unsigned READ_SECTORS_OFFSET = 3 - 1;
+      const static unsigned WRITE_SECTORS_OFFSET = 7 - 1;
+
+      static constexpr const char* TAG_LABEL{"<label>"};
+      /* static constexpr const char* TAG_BAR_READ{"<bar-read>"}; */
+      /* static constexpr const char* TAG_BAR_WRITE{"<bar-write>"}; */
+
+      label_t m_label;
+      std::chrono::milliseconds m_ms{0};
+      float m_time_delta{0.0};
+
+      std::vector<std::string> m_disk_names;
+      std::map<std::string, float> m_read_speeds;
+      std::map<std::string, unsigned long long> m_read_total;
+      std::map<std::string, float> m_write_speeds;
+      std::map<std::string, unsigned long long> m_write_total;
+  };
+}
+
+POLYBAR_NS_END
--- a/src/modules/disk_io.cpp
+++ b/src/modules/disk_io.cpp
@@ -1,0 +1,157 @@
+#include <fstream>
+#include <iomanip>
+#include <istream>
+#include <vector>
+#include <regex>
+#include <sys/types.h>
+#include <dirent.h>
+
+#include "drawtypes/label.hpp"
+#include "drawtypes/progressbar.hpp"
+#include "modules/disk_io.hpp"
+#include "utils/math.hpp"
+
+#include "modules/meta/base.inl"
+
+POLYBAR_NS
+
+namespace modules {
+  template class module<disk_io_module>;
+
+  disk_io_module::disk_io_module(const bar_settings& bar, string name_) : timer_module<disk_io_module>(bar, move(name_)) {
+    m_interval = m_conf.get<decltype(m_interval)>(name(), "interval", 1s);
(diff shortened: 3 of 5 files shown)


doc/config.cmake
```diff
@@ -394,6 +394,14 @@ menu-2-0-exec = sudo poweroff
 menu-2-1 = cancel
 menu-2-1-exec = menu-open-0
 
+[module/disk_io]
+type = internal/disk_io
+interval = 2
+format-prefix = " "
+format-prefix-foreground = ${colors.foreground-alt}
+format-underline = #4bffdc
+label = R: %speed_read% Mb/s W: %speed_write% Mb/s
```
NBonaparte: The units should be part of `%speed_read%` and `%speed_write%`.

src/modules/disk_io.cpp
```diff
@@ -0,0 +1,157 @@
...
+  std::chrono::milliseconds now = std::chrono::duration_cast<std::chrono::milliseconds>(
+      std::chrono::system_clock::now().time_since_epoch());
+
+  std::chrono::duration<float> ms_delta = now - m_ms;
+  float time_delta = ms_delta.count();
+  m_ms = now;
+  return time_delta;
+}
+
+void disk_io_module::_calculate_disk_io_speeds(std::string disk_name)
```
NBonaparte: You should use `string_util::filesize()` if you wanted (kilo/mega)bytes/s, or create another `string_util` function to calculate it in (kilo/mega)bits/s.

include/modules/disk_io.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+  class disk_io_module : public timer_module<disk_io_module> {
+    public:
+      explicit disk_io_module(const bar_settings&, string);
+
+      bool update();
+      bool build(builder* builder, const string& tag) const;
+
+    private:
+      std::vector<std::string> _get_disk_names(void);
```
NBonaparte: No need to start helper functions with underscores.

src/modules/disk_io.cpp
```diff
@@ -0,0 +1,157 @@
...
+  unsigned long long sec_read_total_new{0ULL};
+  unsigned long long sec_write_total_new{0ULL};
+  try {
+    std::ifstream disk_io_info("/sys/block/" + disk_name + "/stat");
+
+    std::string line;
+    std::getline(disk_io_info, line);
+
+    static const std::regex re{"\\s+"};
+    std::vector<std::string> items{
```
NBonaparte: For easier parsing (without variable length whitespaces) you could use `/proc/diskstats/` and split it into a vector with `string_util::split()`.

src/modules/disk_io.cpp
```diff
@@ -0,0 +1,157 @@
...
+  disk_io_module::disk_io_module(const bar_settings& bar, string name_) : timer_module<disk_io_module>(bar, move(name_)) {
+    m_interval = m_conf.get<decltype(m_interval)>(name(), "interval", 1s);
+
+    m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL}/*, TAG_BAR_READ, TAG_BAR_WRITE}*/);
+    if (m_formatter->has(TAG_LABEL)) {
+      m_label = load_optional_label(m_conf, name(), TAG_LABEL, "%speed_read% Mb/s %speed_write% Mb/s");
+    }
+    m_disk_names = _get_disk_names();
+    for (auto name : m_disk_names) {
+      std::cout << "name: " << name << std::endl;
```
NBonaparte: Not needed, unless you want to use `m_log.info()` to output the names.

src/modules/disk_io.cpp
vilkoz: the `/proc/diskstats` actually also has continuous sequence of spaces so I can't easily split it to needed values without regular expression

vilkoz:
This module allows user to select which disk to monitor, if multiple selected - it will show sum of speeds, to minimize consumption of available space on bar.

NBonaparte:
Are you planning to implement `bar-read` and `bar-write`, or would you like to merge this as-is?

vilkoz:
I've planned to add them as soon as I have a free minute

vilkoz:
FYI: I've added bar-read and bar-write, but they don't seem to be very usefull in context of disk read/write speeds. BTW, I think this feature is ready to be merged.

src/modules/disk_io.cpp
```diff
@@ -0,0 +1,169 @@
...
+  float sum_write{0.0f};
+
+  for (auto disk_name : m_disk_names) {
+    calculate_disk_io_speeds(disk_name);
+    sum_read += m_read_speeds[disk_name];
+    sum_write += m_write_speeds[disk_name];
+  }
+
+  m_max_read_speed = std::max(m_max_read_speed, sum_read);
+  m_max_write_speed = std::max(m_max_write_speed, sum_write);
```
patrick96: Are the max speeds just the highest speeds measured during the time the module was running?

vilkoz: Yes, this method will determine maximum disk read/write speed as soon as read/write to disk will continue longer then module update period. Of course you can perform test write and read from disk for determining this values at initialization of the module, but for me it seems to be a little owerkill for module that will run at long period of time

patrick96: So, if any of the disks don't reach their maximum speeds then the values here are going to be inaccurate?

vilkoz: The situation where disks don't reach max speed is highly unlikely and hard to reproduce. But those calculations only made for `<bar-write>` and `<bar-read>`, which are not seem to be very informative at all for displaying precise numbers, they are more suitable for displaying changes in I/O speed, e.g. when you start download directly to hdd, you will see that bar changed dramatically from 0% to 100%, and when download is finished - effect will be opposite.
I've got an idea when I wrote the last sentence: old computers have LED indicators for disk usage, and they (indicators) are quite informative, so you can easily detect if your machine performs some I/O intense actions. Maybe it would be more useful to implement analog for such indicator, using ramp drawtype, as a replacement for bar-read bar-write?

vilkoz:
Hello, I have replaced bar-read/write with indicator-read/write, they informative to the same extent but take much less space on bar. Also, I've added some explanatory comments to the module in config, so configuring of it can be done without exploring the source code.
In my opinion the module is ready to be merged.

jvandenbroek:
Any change this gets merged?

vilkoz:
@jvandenbroek They seems to make plugin architecture for providing additional features, so now you can just use bash script to replace this feature

vilkoz:
@NBonaparte does this pull-request needs some changes, or it should be abandoned?

ironyman:
@vilkoz can we address the late emperor's comments and get this feature merged? :)

patrick96:
This is on my TODO list to be reviewed. I don't have a lot of free time and there are a bunch of higher priority issues, so that's why this hasn't been reviewed yet.

lockie:
Any chance of adding this to `master`?

derrgill:
Really looking forward to a disk I/O module, to have some semblance of parity with the Gnome System Monitor applet / extension that's been around in some form since the Gnome 2.x days

PR #1107, opened 2018-03-19
aur: Add i3-wm dependency

patrick96:
The actual i3 package is also needed to build polybar with i3 support.
It is also listed as a makedepends, like that polybar will always work
with i3, even if the user only decides to install i3 later on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ A compiler with C++14 support ([clang-3.4+](http://llvm.org/releases/download.ht
 **Optional dependencies for extended module support:**
 - alsa-lib *required by `internal/alsa`*
 - libpulse *required by `internal/pulseaudio`*
+- i3-wm *required by `internal/i3`*
 - jsoncpp *required by `internal/i3`*
 - libmpdclient *required by `internal/mpd`*
 - libcurl *required by `internal/github`*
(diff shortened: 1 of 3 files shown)


NBonaparte:
We can't assume everyone's using `makepkg`. In that case, shouldn't we have all the optional module dependencies or at least jsoncpp as makedepends too?

patrick96:
I already had that discussion about making the optdepends makedepends over at the [AUR](https://aur.archlinux.org/packages/polybar), the problem with this is that, if the optdepends are removed after the build polybar will not run anymore because the shared libraries aren't there anymore. Also jsoncpp is already makedepends because cmake relies on it

>We can't assume everyone's using makepkg

Why not? It's the way a package should be built from a PKGBUILD. And if they build the package another way, they are responsible for removing the makedepends after the build anyways.

NBonaparte:
I think the functionality should remain the same as the other optdepends (in the current case, rebuilding the package). Having i3-wm as a makedepends seems kind of misleading.

patrick96:
Fair enough. I also prefer a cleaner makedepends

NBonaparte:
Could you please fix the commit message? It still mentions i3-wm as a makedepends.

patrick96:
Done

NBonaparte:
Already pushed both to AUR.

patrick96:
Great, thank you!

PR #1111, opened 2018-03-21
Ensures that libraries from dependencies in LOCALBASE are used.

lebarondemerde:
@patrick96 

This is building fine in here.

Thank you for your patience.


patrick96:
Thanks!

PR #1118, opened 2018-03-24
fix(pulseaudio): get default sink on new events when using default, add mutex

### LINKED ISSUE #1086: Pulseaudio module doesn't detect a new default sink
(Note: This is similar to #693, but using the new pulseaudio module)

When I connect my bluetooth headphones, pulseaudio automatically switches to that sink as default. However, polybar doesn't notice this switch and still displays the volume for the previous sink.

Is there any way to make the module automatically aware of this change?

NBonaparte:
The new event case should now fall through if there was no sink specified in the config, so it gets the current default sink.

I also added a mutex for the queue because the callback is on a separate thread compared to the module which calls `process_events()`.

Fixes #1086.

### DIFF
--- a/src/adapters/pulseaudio.cpp
+++ b/src/adapters/pulseaudio.cpp
@@ -124,8 +124,8 @@ int pulseaudio::process_events() {
         if (!spec_s_name.empty()) {
           o = pa_context_get_sink_info_by_name(m_context, spec_s_name.c_str(), sink_info_callback, this);
           wait_loop(o, m_mainloop);
+          break;
         }
-        break;
       // get default sink
       case evtype::REMOVE:
         o = pa_context_get_server_info(m_context, get_default_sink_callback, this);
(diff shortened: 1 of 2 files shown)


src/adapters/pulseaudio.cpp
```diff
@@ -124,8 +126,8 @@ int pulseaudio::process_events() {
         if (!spec_s_name.empty()) {
           o = pa_context_get_sink_info_by_name(m_context, spec_s_name.c_str(), sink_info_callback, this);
           wait_loop(o, m_mainloop);
+          break;
         }
-        break;
       // get default sink
```
patrick96: Can you add `//FALLTHRU`, like in commit https://github.com/jaagr/i3ipcpp/commit/57e59bd8bad21fbb9d18bbb233ff33012c74efa9, some compilers don't like implicit fallthroughs.

PR #1122, opened 2018-03-26
feat(xwindows): add pin-output and pin-show-last

### LINKED ISSUE #252: feat(xwindow) Monitor locality: show (last) active window title on bar's monitor

NBonaparte:
Use `_NET_DESKTOP_VIEWPORT` for determining whether a desktop is on a monitor, similar to the xworkspaces module.
Adds options `pin-output` and `pin-show-last`, which determines whether to show the previous title from the monitor or a blank string.

Fixes #252.


jeffmhubbard:
I don't think this is working as expected. I have pin-output and pin-show-last set to true. Everything seems to work fine when switching between monitors and workspaces that have windows on them. But whenever I switch to an empty workspace on either monitor it clears the titles on both bars. If this fixes #252, shouldn't the title remain on the monitor\workspace that still has windows on it? I have tested this with both bspwm and i3, and I get the same results.

He-Ro:
I am using this branch currently as well and also came across the issue @jeffmhubbard describes.
Furthermore I would like to give feedback on `pin-show-last`: I think it would be a nice feature, if there was a new format string for the xwindow module on an inactive workspace. This way one would achieve the `pin-show-last` effect by setting `format-inactive` to an empty string. The option `pin-show-last` would then be obsolete. Additionally one could e.g. change the foreground color of the title to a light gray, if the workspace is not active right now. Or all the other customization options that are possible with a separate format: `format-inactive-foreground`, `format-inactive-background`, `format-inactive-overline`, etc.
I have no idea how feasible this is, but I wanted to give feedback on an awesome feature I am eager to see in the master branch.

He-Ro:
@NBonaparte I would love to see this merged. I am not proficient in the code base, so I would not expect myself to be much of a help in coding (with some pointers however it might be doable). But what I definitely can help with is some testing, if you know what I should test out, since I have a second monitor and the problems in the above comments mainly originate from using multiple monitors.
Cheers!

betaboon:
@NBonaparte can this move forward as #576 has been closed in favor of this ?

Lomadriel:
I agree with @He-Ro, adding a new state, and so a new formatter, when the monitor isn't focus seems to be more customizable then using a boolean.
@NBonaparte Are you still planning to work on this? If you don't want to work on this PR or you don't have the time, I would like to continue your work to finish this feature.

BeaudanBrown:
@Lomadriel Bumping this to see if you ended up working on it?
I would love to see this feature

patrick96:
I don't think anyone is working on this PR. I have gone ahead and resolved the merge conflicts, if anyone wants to take over and address some of the issue mentioned by the people in this thread, let me know.

I don't think you can push to this PR, so the best way would be to [pull the changes from Nbonaparte's repo](https://stackoverflow.com/a/42157071/5363071), create your own branch, and open a PR with that branch.

Issue #1055, opened 2018-02-26
mpd module not updating


SuperPrower:
```
[module/mpd]
type = internal/mpd

host = 127.0.0.1
port = 6600
password =

interval = 2

label-song-maxlen = 72

format-online =  <label-song>
format-playing-underline = ${colors.underline}
label-offline =  mpd is offline
```

```
]$ polybar -vvv
polybar 3.1.0-64-g4f6e374

Features: +alsa +curl +i3 +mpd +network +pulseaudio

X extensions: +randr (+monitors) -render -damage -sync -composite +xkb +xrm +xcursor

Build type: RelWithDebInfo
Compiler: /usr/bin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -Wall -Wextra -Werror -Wno-noexcept-type -O2 -pedantic -pedantic-errors
Linker flags: -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now
```

I use Mopidy as my server.

Status, i.e. playing/paused, track name doesn't update and stay like that until I restart polybar. Polybar binary doesn't seem to say anything about this module - it just freezes like that.

Also, not configured here, but clicking on shuffle label seemed to cause huuge problems with my player ncmpcpp - it can't enable/disable shuffle mode anymore. Pressing pause (<toggle>) seemed to be very delayed.

patrick96:
Do other modules still update (like the date module)? The pause button still works in polybar, even if the module is frozen.

SuperPrower:
Hm, I have restarted Mopidy, because ncmpcpp also started acting weird, and problem seems to go away. I suspect it has something to do with Polybar's interaction with MPD - will try to add pause button back and click it, or do something else.

SuperPrower:
So yeah, it has something to do with clickability of the module. I checked output of mopidy - it keeps getting incoming connections from local machine every few seconds, even after polybar hanged.

Here is proper steps to re-produce issue:

1. Have `<toggle>` and `<icon-random>` in `format-online`
2. (use mopidy as mpd server? is it causing the issue?)
3. tap random icon to enable random mode. It is reflected in ncmpcpp.
4. tap random icon again. nothing happens. (optional?)
5. tap pause button. Music pauses. Polybar module doesn't change at all. Everything else - time and stuff - keeps working.
6. tap (still) pause button. Nothing happens.
[...]

PR #1135, opened 2018-03-31
mpd logging

patrick96:
While debugging #1055, I added some trace logging to the mpd module that I
think could be useful. They will also not spam the output too much, as they
are only executed on user clicks

### DIFF
--- a/src/adapters/mpd.cpp
+++ b/src/adapters/mpd.cpp
@@ -10,7 +10,7 @@
 
 POLYBAR_NS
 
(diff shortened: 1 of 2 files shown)


patrick96:
Done

src/adapters/mpd.cpp
```diff
@@ -476,4 +483,6 @@ namespace mpd {
   // }}}
 }
 
+#undef LOG_BOOL
```
NBonaparte: Is this supposed to be `TRACE_BOOL`?

patrick96: Yes, it is. Must have missed that when I renamed the rest.

PR #1136, opened 2018-03-31
Clear entire xwindow label if no window is open

kenogo:
If you use the current [config file](https://github.com/jaagr/polybar/wiki/Module:-xwindow) in the wiki entry for the xwindow module, not having an active window will still result in "Window: " being displayed in the bar. This doesn't really make sense IMO, so I added an if statement to clear the entire label (including the "Window: ") if there is no active window.

I could also add an option to make this configurable, but I don't really see why anyone would want the "Window: " to be displayed.

### DIFF
--- a/src/modules/xwindow.cpp
+++ b/src/modules/xwindow.cpp
@@ -77,7 +77,10 @@ namespace modules {
     m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL});
 
     if (m_formatter->has(TAG_LABEL)) {
-      m_label = load_optional_label(m_conf, name(), TAG_LABEL, "%title%");
+      m_statelabels.insert(
+          make_pair(state::ACTIVE, load_optional_label(m_conf, name(), "label", "%title%")));
+      m_statelabels.insert(
+          make_pair(state::EMPTY, load_optional_label(m_conf, name(), "label-empty", "")));
     }
   }
 
@@ -118,9 +121,12 @@ namespace modules {
       m_active = make_unique<active_window>(m_connection, win);
     }
 
-    if (m_label) {
+    if (m_active) {
+      m_label = m_statelabels.find(state::ACTIVE)->second->clone();
       m_label->reset_tokens();
-      m_label->replace_token("%title%", m_active ? m_active->title() : "");
+      m_label->replace_token("%title%", m_active->title());
+    } else {
+      m_label = m_statelabels.find(state::EMPTY)->second->clone();
     }
   }
 
--- a/src/modules/xwindow.cpp
+++ b/src/modules/xwindow.cpp
@@ -77,10 +77,8 @@ namespace modules {
     m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL});
 
     if (m_formatter->has(TAG_LABEL)) {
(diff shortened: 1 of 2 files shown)


NBonaparte:
It already does here: `m_active ? m_active->title() : ""`. Just leave the label as `%title%` (or don't set it at all so it will be set as default to that) and no "Window: " will be shown.

kenogo:
Sorry, maybe I didn't express myself clearly. I want the "Window: " part to be shown, but it only makes sense to be shown when there actually is an active window. IMO, the desired behavior should be as follows:

Terminal window active -> output: "Window: Terminal"
No window active -> no output

while the current behavior is

No window active -> output: "Window: "

I doubt anyone wants that stray "Window: " part or a window icon without any text to replace `%title%`

kenogo:
Maybe it's easiest explained with pictures. So this is what I want, see how there's a window icon in front the window's name:

bildschirmfoto_2018-04-01_00-49-07

However, the current behavior when there isn't an active window is the following (the icon is still there and looks ugly), and there's currently no way to change this:

bildschirmfoto_2018-04-01_00-52-00

I changed this behavior to the following:

bildschirmfoto_2018-04-01_00-49-32

NBonaparte:
I understand now. I think it should be set as an option (like `no-window-blank = true`) or create another label just for this purpose, because otherwise this behaviour deviates from what is normally expected.

patrick96:
I would prefer, if another label or format was used to keep it consistent with what other modules do

kenogo:
Ok, I guess the most consistent way would be to use two labels `label-active` and `label-unactive`? I can start working on this.

patrick96:
Yeah, almost. Either create a second format `format-empty` and let it have a separate label like `label-empty` (similar to what the network module does), or just have a second label and use `label`, if there is a window title and `label-empty` when there is not (this is what the xworkspaces module does wher `<label-state>` is replaced depending on the module state). The main label needs to be called `label` though, because it would break polybar configs when changed.

kenogo:
Alright this should do :)

src/modules/xwindow.cpp
```diff
@@ -118,9 +121,12 @@ namespace modules {
...
     }
 
-    if (m_label) {
+    if (m_active) {
+      m_label = m_statelabels.find(state::ACTIVE)->second->clone();
       m_label->reset_tokens();
-      m_label->replace_token("%title%", m_active ? m_active->title() : "");
+      m_label->replace_token("%title%", m_active->title());
+    } else {
+      m_label = m_statelabels.find(state::EMPTY)->second->clone();
```
patrick96: Using `.at(..)` instead of `.find(..)` saves you `->second`

kenogo:
Here ya go. Hit me up if you want me to squash commits and/or follow some kind of commit naming scheme!

patrick96:
We're gonna squash and merge this one anyway, but in the future try to follow the de factor [standard git commit message rules](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) and also try to structure structure the first line as `purpose(target): msg`

Where purpose is what the commit does and is usually `fix` or `feat` (for feature) and target is the name of the component your commit is affecting. Just `target: msg` is also fine. See our commit history for some examples

PR #1152, opened 2018-04-08
Migrate to googletest and add coverage reports

patrick96:
For my work on the config parser, I wanted an easier to use testing framework as
we had before. Right now only two tests are ported but the others will be quite
straight-forward

I have also linked this repo with [coveralls.io](https://coveralls.io/github/jaagr/polybar) so that we can see
what our tests missed.

I have also done some cleanup in the travis config file and created several
build configurations: [Debug, Release] X [gcc, clang] and a separate one for the
tests (using gcc because gcov needs gcc).


patrick96:
I see your concern and cloning googletest would be quite unnecessary especially on the build server.
If we made cloning optional, we would only be able to build the tests, if we specifically enable it when calling cmake, which is fine, it would just be a break from the current functionality. I will push another commit to make this happen.

As for the error, it seems to be only occuring with clang, because we have `-Wmissing-field-initializers` and clang complains about that and gcc doesn't. I will add `-Wno-missing-field-initializers` to the `CXX_FLAGS`

patrick96:
Done, I don't actually know how to give you access to the coveralls repo. This is strange, since it seems to be the first one to register a repo becomes the only one with access.

NBonaparte:
If you're not against trying out other suites we could also try [codecov](https://codecov.io/).

patrick96:
I'm open to that. Do you have any experience with codecov? One thing I noticed with coveralls is that the upload script doesn't work, if I pass the `-a` flag to `gcov`. What `-a` does, is give you coverage reports for each block, so for
```
if(x > 2 || y < 5) {
  ...
}
```
we get coverage reports for both `x > 2` and `y < 5`. Do you know, if codecov supports that? It would be really useful to get coverage for every possible code path.

Would you look into that? I have already spent quite a lot of time on the coveralls integration.

patrick96:
Just looked a little at codecov and it seems it has a lot more features (and documentation) than coveralls. So switching to it may be beneficial to our testing efforts

NBonaparte:
I’ll take a look at migrating to it in the next few days.

PR #1153, opened 2018-04-09
(pulseaudio) Add max volume configuration

### LINKED ISSUE #1125: Pulseaudio module max volume
With alsa module 100% was max volume I could go to, with pulseaudio I can go way higher, is there a way to cap it at 100% (or some other configurable value in the config)  ?

kajzersoze:
Adds `max-volume` configuration.
If set to `true`  volume is capped at 150%, 
if `false`it's capped at 100%, 
unset = `true`

Fixes #1125

### DIFF
--- a/include/adapters/pulseaudio.hpp
+++ b/include/adapters/pulseaudio.hpp
@@ -26,7 +26,7 @@ class pulseaudio {
   using queue = std::queue<evtype>;
 
   public:
-    explicit pulseaudio(const logger& logger, string&& sink_name);
+    explicit pulseaudio(const logger& logger, string&& sink_name, bool m_max_volume);
     ~pulseaudio();
 
     pulseaudio(const pulseaudio& o) = delete;
@@ -74,6 +74,8 @@ class pulseaudio {
     string spec_s_name;
     string s_name;
     uint32_t m_index{0};
+
+    pa_volume_t max_volume{PA_VOLUME_UI_MAX};
 };
 
 POLYBAR_NS_END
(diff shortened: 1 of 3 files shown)


NBonaparte:
I'd prefer to pass `m_max_volume` only once (i.e. in the constructor), so we don't have to keep passing it for every volume change. Also, a better and more descriptive setting name might help.

kajzersoze:
Agreed, will do that.
As for setting name I can't think of a better one and yet to be short enough, description for it would be in the Wiki. Though I'll think about it more tomorrow, maybe I'll come up with something.

kajzersoze:
KDE / Plasma calls that volume overdrive.

> Volume overdrive. Pulseaudio allows the volume to be increased beyond 100%. This is an often requested feature and should be supported in some form. It likely needs visual cues as to where 100% is and where the overdrive limit is (usually applications will allow up to 130% although technically much more is possible).

It could be called volume-boost maybe, I really don't know, I'm not good at naming things.
maximal-possible-volume-level
volume-cap-above-normal
..

I think you should pick / produce the right one.

include/adapters/pulseaudio.hpp
```diff
@@ -74,6 +74,8 @@ class pulseaudio {
     string spec_s_name;
     string s_name;
     uint32_t m_index{0};
+
+    pa_volume_t max_volume = PA_VOLUME_UI_MAX;
```
NBonaparte: Could you please use brace initialization here?

kajzersoze: Yeah, was in a hurry this morning...

NBonaparte:
How about `use-ui-max` for the setting name?

PR #1166, opened 2018-04-14
xworkspaces: Deprecate pin-workspaces

patrick96:
The ewmh spec ([1]) doesn't have a notion of mapping a workspace to a
monitor. Whatever polybar does when pin-workspaces is true, may be
totally different than what the WM does, because it can't tell us how
the workspaces are mapped with ewmh atoms.

[1]: https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html

### DIFF
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -32,6 +32,12 @@ namespace modules {
     m_click = m_conf.get(name(), "enable-click", m_click);
     m_scroll = m_conf.get(name(), "enable-scroll", m_scroll);
 
+    if(m_pinworkspaces) {
+      m_log.warn("%s: The config parameter `pin-workspaces = true` is deprecated, "
+          "as defining a workspace on an output only is not a general WM concept. "
+          "Please stop using `pin-workspaces`, you may experience weird behaviour", name());
+    }
+
     // Initialize ewmh atoms
     if ((m_ewmh = ewmh_util::initialize()) == nullptr) {
       throw module_error("Failed to initialize ewmh atoms");


NBonaparte:
Don't you mean deprecate `pin-workspaces`?

patrick96:
Whoops, of course.

NBonaparte:
I think viewports are more of a de facto standard, as numerous WMs support its usage while others don't (see https://github.com/xmonad/xmonad-contrib/issues/150#issuecomment-298169758). It is still useful for those that support it, so I don't think it would be wise to remove its functionality.

patrick96:
How would a WM define a monitor-workspace mapping by setting EWMH atoms? I don't see how that would be possible.

NBonaparte:
[_NET_DESKTOP_VIEWPORT](https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472713248):

> Array of pairs of cardinals that define the top left corner of each desktop's viewport.

Basically coordinates for each workspace/desktop. On bspwm with 5 desktops on my left monitor and 5 on my right, I have
```
_NET_DESKTOP_VIEWPORT(CARDINAL) = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1920, 0, 1920, 0, 1920, 0, 1920, 0, 1920, 0
```
By cross-referencing with monitor geometries, we can determine which monitor the desktop is on.
Since the x-values of desktops 6-10 are greater than my monitor width, we know that they are on the right monitor.

This should work similarly for i3 and other WMs that support this.

patrick96:
Ah, I see. The two WMs I tested this with didn't do this. i3 only set a coordinate pair per monitor (so two in my case) and hlwm always sets `0, 0`. Also I had some quite weird behaviour in i3 where polybar showed workspaces on the wrong monitor. And on hlwm the workspaces are only displayed on the left monitor (the one with `(0, 0)` as the coordinates), but this is to be expected.
I'll have a closer look at this, but I agree that if a viewport is defined for each workspace, then this feature makes a lot of sense. I'm going to close this.

PR #1168, opened 2018-04-15
fix(pulseaudio): listen for default sink changes, and use @DEFAULT_SINK@

### LINKED ISSUE #1167: When pulseaudio's default sinks changes, the volume bar doesn't truly reflect the volume although the mouse bindings work
The title is a bit long but I hope it was informative. Very similar to what @indeedwatson has [pointed out](https://github.com/jaagr/polybar/issues/1086#issuecomment-380833286), the problem I'm experiencing is this:

If I change pulseaudio's default sink with a script like [@indeedwatson's](https://gist.github.com/47cfe7ca5f83a77ccdd6b970581daeca) or through any another API (like [this](https://github.com/stefano-m/lua-pulseaudio_dbus) in my case), and I put the mouse on the volume bar and scroll up or down in order to raise or lower the volume, the default sink's volume is truly the one that is changed thanks to https://github.com/jaagr/polybar/pull/1118 (I guess) but the volume bar doesn't really change it's appearance and it **looks** like the volume doesn't really change although one can hear it change.

NBonaparte:
Fixes #1167 by listening for changes in the default sink. Also, we can reduce the number of callbacks by using `@DEFAULT_SINK@` as the default sink name, so we don't have to retrieve the actual name.

While we're at it, I also fixed `max_volume` naming to be consistent.

### DIFF
--- a/include/adapters/pulseaudio.hpp
+++ b/include/adapters/pulseaudio.hpp
@@ -74,7 +74,7 @@ class pulseaudio {
     string s_name;
     uint32_t m_index{0};
 
-    pa_volume_t max_volume{PA_VOLUME_UI_MAX};
+    pa_volume_t m_max_volume{PA_VOLUME_UI_MAX};
 };
 
 POLYBAR_NS_END
--- a/include/adapters/pulseaudio.hpp
+++ b/include/adapters/pulseaudio.hpp
@@ -22,7 +22,7 @@ DEFINE_ERROR(pulseaudio_error);
 
 class pulseaudio {
   // events to add to our queue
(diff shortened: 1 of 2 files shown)


include/adapters/pulseaudio.hpp
```diff
@@ -63,7 +62,7 @@ class pulseaudio {
     pa_cvolume cv;
     bool muted{false};
     // default sink name
-    string def_s_name;
+    static constexpr auto DEFAULT_SINK{"@DEFAULT_SINK@"};
```
patrick96: Where is `@DEFAULT_SINK@` defined? Or is it a magic string within pulseaudio?

PR #1171, opened 2018-04-16
fix(net): Mark not connected on querying failure

### LINKED ISSUE #1163: Detect network interface removal
I am using a USB to Ethernet adapter, since my laptop does not have an ethernet port.

If I disconnect the ethernet cable from the adapter, the state of the network interface is set to down, which changes the polybar network module normally.

However, if I disconnected the adapter from the laptop without first disconnecting the ethernet cable from the adapter (i.e, the network interface disappears, without first changing its state do down) the polybar network module stops updating itself, continuing to display the last state before the adapter was removed instead of displaying the disconnected format.

patrick96:
I don't see a reason why not to treat a negative result from `network::query` as a reason why the interface is down.
Is there something I'm missing here @NBonaparte?

Fixes #1163

### DIFF
--- a/src/modules/network.cpp
+++ b/src/modules/network.cpp
@@ -86,6 +86,7 @@ namespace modules {
 
     if (!network->query(m_accumulate)) {
       m_log.warn("%s: Failed to query interface '%s'", name(), m_interface);
+      m_connected = false;
       return false;
     }
 


patrick96:
Not sure either. But I have again looked at all the `query` methods and it seems all their `return false` paths justify setting the interface to disconnected

Issue #1195, opened 2018-04-25
Use Unicode ellipsis instead of three dots


M1dgard:
In monospaced fonts, the three dots `...` are too wide: ![screenshot_2018-04-26_01-28-48](https://user-images.githubusercontent.com/4283376/39278025-72f9bce6-48e0-11e8-9722-66de7b1e6290.png)

The single glyph `…` looks better: ![screenshot_2018-04-26_01-32-52](https://user-images.githubusercontent.com/4283376/39278138-0683d082-48e1-11e8-92db-f15ec5f5cef9.png)

patrick96:
We could maybe even add an option to the label to set the ellipsis string. But you're right, the the character ellipsis is quite long

M1dgard:
Yes, being able to do
```ini
label-song-ellipsis = …
label-song-ellipsis = ...
label-song-ellipsis = ..
label-song-ellipsis = $
label-song-ellipsis = $
```
would be nice

patrick96:
Exactly, but we can't name it `label-song-ellipsis`. That is already taken for the boolean.

> I don't know if it's currently taken into account, but make sure to subtract the length of the chosen ellipsis string from maxlen!

I think we already have logic for that in the token handling, so we can use that ;)

M1dgard:
> Exactly, but we can't name it `label-song-ellipsis`. That is already taken for the boolean.

I think that #1194 proves that no-one really used it. :p So wouldn't it be possible to change the type to string? You can use an empty string to disable the ellipsis.

patrick96:
It's technically possible to change the type, but it would break all existing configs that use it. Right now we only know that nobody who has it in their config noticed that it doesn't work. But `label-song-ellipsis = true` is part of the config shipped with polybar, so people would suddenly have 'true' added to their cut off labels.

What we will probably need to do is define another key, deprecate `label-*-ellipsis`, and maybe in the next major release use `label-*-ellipsis` the way you described

kronn:
Isn't this issue resolved by merging #1198?

PR #1198, opened 2018-04-28
fix(builder): Respect label-ellipsis option

### LINKED ISSUE #1194: Can't disable ellipsis
From my MPD module config:
```ini
label-song          = %artist% - %title%
label-song-maxlen   = 22
label-song-ellipsis = false
```

But the ellipsis is still shown: screenshot_2018-04-26_01-28-48

patrick96:
As described in #1194 polybar never actually checks the m_ellipsis flag
in labels.

Right now, I have only refactored the builder to determine the actual
label text inside a function and added failing tests for that bug and
the general expected behaviour of get_label_text. This is why the
tests in the current build will fail.

This will probably decrease test coverage quite drastically, as codecov
doesn't track builder.cpp yet, so this basically adds a bunch of not
covered lines that codecov is now tracking

Also @NBonaparte, do you think the builder should throw an exception,
if the maxlen is less than the length of the ellipsis and ellipsis is
enabled, because it would be impossible to follow both the label-maxlen
and the label-ellipsis directive?

EDIT1: Seems like the new test doesn't build on travis, and for some reason the travis build doesn't fail when this happens.
EDIT2: I have fixed the compilation error, now the test should fail for the right reasons
EDIT3: Fixes #1194

### DIFF
--- a/src/drawtypes/label.cpp
+++ b/src/drawtypes/label.cpp
@@ -221,6 +221,17 @@ namespace drawtypes {
       }
     }
 
+    size_t maxlen = conf.get(section, name + "-maxlen", 0_z);
+    bool ellipsis = conf.get(section, name + "-ellipsis", true);
+
+    if(ellipsis && maxlen > 0 && maxlen < 3) {
+      logger::make().err(sstream() << "Label " << section << "." << name
+          << " has maxlen " << maxlen
+          << ", which is smaller than length of ellipsis (3), disabling ellipsis...");
+
+      ellipsis = false;
+    }
+
     // clang-format off
     return factory_util::shared<label>(text,
         conf.get(section, name + "-foreground", ""s),
@@ -230,8 +241,8 @@ namespace drawtypes {
         conf.get(section, name + "-font", 0),
         padding,
         margin,
-        conf.get(section, name + "-maxlen", 0_z),
-        conf.get(section, name + "-ellipsis", true),
+        maxlen,
+        ellipsis,
         move(tokens));
     // clang-format on
   }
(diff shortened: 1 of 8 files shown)


patrick96:
Hold off on merging this, I think we should just log a warning and disable ellipsis, if the maxlen < 3

patrick96:
I think this is ready for review now.

patrick96:
> What if we used an ellipsis character ("…") instead of three periods?

This will be what I implement next, in a separate PR, a config option to define the ellipsis.

M1dgard:
Configurable ellipsis = #1195

M1dgard:
I tested this patch as a user, I didn't find any issue. Beware that it changes behaviour (that's intended, and I requested this change in #1195): `maxlen 4` used to yield `abcd...`, now it yields `a...`

src/drawtypes/label.cpp
```diff
@@ -221,6 +221,17 @@ namespace drawtypes {
       }
     }
 
+    size_t maxlen = conf.get(section, name + "-maxlen", 0_z);
+    bool ellipsis = conf.get(section, name + "-ellipsis", true);
+
+    if(ellipsis && maxlen > 0 && maxlen < 3) {
+      logger::make().err(sstream() << "Label " << section << "." << name
```
NBonaparte: Couldn't we throw an `application_error` or some other type of error to avoid creating a new logger?

patrick96: `logger::make` uses the singleton pattern, so only the first call to `logger::make()` will create a new instance, all others always return the same instance. I used it directly because I didn't want to add the `m_log` field just for that.. Btw. I really don't like the use of a singleton for the logger but that is a discussion for another time.

If we threw an `application_error` here, wouldn't it be caught in the controller and would disable the whole module? I was debating, if we wanted invalid maxlen be a fatal error, would you prefer that?

NBonaparte: Yeah, I think that would be much neater.

patrick96: Alright, gonna revert back to throwing an error when I have the time

patrick96: Done

PR #1213, opened 2018-05-05
refactor(battery): simplify animation subthread

cs:
I came up with this while trying to make the animations "smoother" => currently, it's skipping animation icons quite often due to timing issues on my machine. Haven't been able to improve animation smoothness, though, as it appears to be a deeper issues. In any case, the calls of `broadcast()` appear to be timed correctly.

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -338,24 +338,18 @@ namespace modules {
    * same time.
    */
   void battery_module::subthread() {
-    chrono::duration<double> dur{0.0};
-
-    if (battery_module::state::CHARGING == m_state && m_animation_charging) {
-      dur += chrono::milliseconds{m_animation_charging->framerate()};
-    } else if (battery_module::state::DISCHARGING == m_state && m_animation_discharging) {
-      dur += chrono::milliseconds{m_animation_discharging->framerate()};
-    } else {
-      dur += 1s;
-    }
+    m_log.trace("%s: Start of subthread", name());
 
     while (running()) {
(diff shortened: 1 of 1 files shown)


patrick96:
Thanks for contributing. People in #568 also complained about uneven animations. I will review this and look into why the animations has some deeper issues.

src/modules/battery.cpp
```diff
@@ -338,24 +338,18 @@ namespace modules {
...
-      for (int i = 0; running() && i < dur.count(); ++i) {
-        if (m_state == battery_module::state::CHARGING ||
-            m_state == battery_module::state::DISCHARGING) {
-          broadcast();
-        }
-        sleep(dur);
+      int framerate = 1000; // milliseconds
+      if (m_state == battery_module::state::CHARGING) {
+        broadcast();
+        framerate = m_animation_charging->framerate();
```
patrick96: `m_animation_charging` may be NULL

src/modules/battery.cpp
```diff
@@ -338,24 +338,18 @@ namespace modules {
...
+      } else if (m_state == battery_module::state::DISCHARGING) {
+        broadcast();
+        framerate = m_animation_discharging->framerate();
```
patrick96: `m_animation_discharging` may be NULL

Lomadriel:
> Maybe the uneven animation come from the fact that we don't use a steady clock (the PR for this is #1725).

Nop the problem is the synchronization between the update and the animation thread. I fixed that in my PR (#1683). It also includes this PR.

cs:
Closing in favor of #1683.

PR #1217, opened 2018-05-06
fix(builder): Don't truncate colors with same channels

### LINKED ISSUE #1183: i3 module ignores transparency if color is black
If I set the background colour for the i3 module to `#00000000` it is opaque. Setting it to anything other than zero causes it to behave as expected. Other modules / the bar itself do not have this problem.

`#00000000` -> opaque black
`#00010101` -> fully transparent
`#01000000` -> (pretty much) fully transparent

Version: Built locally off tag 3.1.0
System: Fedora 27

patrick96:
The builder would convert bg and fg colors like #00000000 to #000 which
would make it loose its alpha channel.

Fixes #1183

### DIFF
--- a/src/components/builder.cpp
+++ b/src/components/builder.cpp
@@ -356,8 +356,6 @@ void builder::background(string color) {
     string bg{background_hex()};
     color = "#" + color.substr(color.length() - 2);
     color += bg.substr(bg.length() - (bg.length() < 6 ? 3 : 6));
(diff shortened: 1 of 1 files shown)


patrick96:
Yeah, was wondering about that too. A lot of the code that handles colors seems to be suffering from premature optimizations. For example, the `simplify_hex` function tries to 'optimize' `#AABBCC` to `#ABC` and will also strip the alpha channel, if it's `FF`. As far as I can tell the code I deleted didn't server any particular purpose.

PR #1221, opened 2018-05-07
fix(aur): Backport gcc fix

patrick96:


patrick96:
I have gone ahead and pushed the fix to the AUR already so that people can still build

Issue #675, opened 2017-07-23
--reload doesn't monitor include-files


bricewge:
When using the flag `--reload` if a file has been included with `include-file` and edited the bar will not be updated with the new settings.

patrick96:
This seems to have been forgotten when the `include-file` option was introduced, I'll take a look at it

PR #1237, opened 2018-05-15
New config parser

### LINKED ISSUE #412: Nested inherit no longer works?
I just updated polybar and my config seems broken. I rely a lot on inherit to make my config tidy. It goes something like this.

```
[bar/bar]
monitor = ${env:MONITOR:LVDS1}
override-redirect = true
bottom = true

[bar/bg]
inherit = bar/bar
background = ${colors.white}
font-0 = ${fonts.null}
height = 16
modules-center = null

[bar/fg]
inherit = bar/bar
foreground = ${colors.black}
height = 12
border-bottom-size = 2

[bar/icon]
inherit = bar/fg
font-0 = ${fonts.icon}

[bar/text]
inherit = bar/fg
font-0 = ${fonts.text}

[bar/icon_blue]
inherit = bar/icon
background = ${colors.blue-0}

[bar/icon_teal]
inherit = bar/icon
background = ${colors.teal-0}

[bar/icon_orng]
inherit = bar/icon
background = ${colors.orng-0}

[bar/text_blue]
inherit = bar/text
background = ${colors.blue-1}

[bar/text_teal]
inherit = bar/text
background = ${colors.teal-1}

[bar/text_orng]
inherit = bar/text
background = ${colors.orng-1}

...more bars with inherits in them
```

This works on the previous version polybar (i'm not sure which one since I haven't updated it for a while). I noticed it has something to do with `inherit` since the config fails to refer to the bar that also inherits another bar.

For example from the config above, `bar/bg` will succesfully inherit `bar/bar` and will render a normal bar. `bar/icon` and `bar/text` however, will fail to inherit `bar/fg` because it also inherits `bar/bar`. `bar/icon_blue` will also fail to inherit `bar/icon` and so on, basically makes my config obsolete.

### LINKED ISSUE #783: Self referenced colours caused a segfault
Definitely not a normal thing to do, but I accidentally copied colours from the bottom bar to the top bar like this:  
```
[bar/top]
background = ${bar/top.background}
foreground = ${bar/top.foreground}
```
  

Definitely not urgent, as this was user-error, but segfaults are bad.  

On Arch linux:  
```
polybar 3.0.5-70-g5b7d7b8

Features: +alsa +curl +i3 +mpd +network
```

### LINKED ISSUE #954: Setting font from variable disables further font options
I load my font information from xresources.
`.Xresources`:  `budstyle.font: FixedFixedsys`  

polybarconfig:  
this works:   
`font-0 = ${xrdb:budstyle.font:monospace}`  

these two doesn't work:  
font-0 = ${xrdb:budstyle.font:monospace};2`  
font-0 = ${xrdb:budstyle.font:monospace}:size=12:antialias=false;1`  

Is there some workaround other then accepting no styling or harcoding the font?

patrick96:
I have taken the first step to implement a new config parser. For now it's
only a skeleton. But it should help fix a bunch of issues and make the
parser as well as the dependency checking for variable references as well as
included files much more robust

It will also support giving more helpful error messages for syntax errors, I
think the current implementationn doesn't say anything about syntax errors.

Will add more info later with what my plans are

Should
Fix #412
Fix #783
Fix #954

And make the following obsolte:
Closes #1032
Closes #784

It should also make it easier to add an inotify watch onto included files,
as described in #675

### DIFF
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -116,7 +116,7 @@ def FlagsForFile(filename, **kwargs):
     # Bear in mind that compilation_info.compiler_flags_ does NOT return a
     # python list, but a "list-like" StringVec object
     final_flags = MakeRelativePathsInFlagsAbsolute(
-        compilation_info.compiler_flags_,
+        [x for x in compilation_info.compiler_flags_ if x != "-Werror"],
         compilation_info.compiler_working_dir_)
   else:
     # We use default flags if GetCompilationInfoForFile can't find any flags
--- a/include/components/config_parser.hpp
+++ b/include/components/config_parser.hpp
@@ -1,0 +1,262 @@
+#pragma once
+
+#include <algorithm>
+#include <unordered_map>
+
+#include "common.hpp"
+#include "errors.hpp"
+#include "components/logger.hpp"
+#include "utils/file.hpp"
+#include "utils/string.hpp"
+
+POLYBAR_NS
+
+/**
+ * \brief Generic exception for the config_parser
+ */
+DEFINE_ERROR(parser_error);
+
+/**
+ * \brief Exception object for syntax errors
+ *
+ * Contains filepath and line number where syntax error was found
+ */
+class syntax_error : public parser_error {
+
+  public:
+    /**
+     * Default values are used when the thrower doesn't know the position.
+     * parse_line has to catch, set the proper values and rethrow
+     */
+    syntax_error(string msg, string file = "", int line_no = -1)
+      : parser_error(msg), file(file), line_no(line_no), msg(msg) {}
+
+    virtual const char* what() const throw() {
+      return ("Syntax error at " + file + ":" + to_string(line_no) + " " + msg).c_str();
+    }
+
+  private:
+    string file;
+    int line_no;
+    string msg;
+};
+
+/**
+ * \enum line_type
+ * \brief All different types a line in a config can be
+ */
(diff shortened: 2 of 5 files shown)


swillner:
I don't want to intrude on your work, but are there plans to change the configuration format to JSON or YAML (I have made very good experiences with jbeder/yaml-cpp). This would much simplify the specification of lists and eliminate the need to implement a parser just for this project... I would be willing to help there.

patrick96:
@swillner I have not considered this before. I think the big advantage of the ini format is that it is very easy to use for the users since we actually require them to write the config themselves.
I think we'll stick to the ini format for the time being since it would also break all existing configs to switch formats.

FichteFoll:
I believe the toml format is more appropriate for a tool like polybar. It's not as complex as yaml, is similar to ini but properly specified and has proper lists.

patrick96:
We're not going to switch config formats anytime soon.

sudoforge:
Hey @patrick96 -- I've recently sat down and "modularized" my polbar configuration, which led to finding out that included files don't propagate a reload of the configuration like the root conffile does (this would be related to #675, #1909).

It looks like that portion of this effort never landed in trunk -- can you describe whether or not you think completing this effort or tackling #675 independently makes more sense to you? I don't have much experience with CPP and haven't contributed to Polybar before but would like to take a stab at it.

patrick96:
@sudoforge I think tackling this separately makes perfect sense. It should only tangentially touch the config code (to get all config files that need to be watched). I haven't done any work on that issue at all, so feel free to tackle this yourself. Thanks :smiley:

Issue #1234, opened 2018-05-13
IPv6 support in net module


sqwishy:
It might be nice to show IPv6 addressing status with a `%local_ip6%` label or something. `%local_ip%` only displays IPv4 addresses. One way is to use `getnameinfo` for AF_INET6 addresses, as is done with AF_INET currently, and ignore link local addresses. But I don't believe it's terribly uncommon to have multiple global IPv6 addresses on an interface and it's not entirely clear how to pick the "best" one.

i3status [creates a UDP socket](https://github.com/i3/i3status/blob/master/src/print_ipv6_addr.c) to determine the address that the operating system wants to use to reach some global address. That might be a good approach.

patrick96:
Good idea, we should probably do this. I don't have any ipv6 addresses on any of my interfaces right now (except for loopback), but I'd like to get this done definitely by the next major release, maybe earlier.

sqwishy:
I put [a commit](https://github.com/sqwishy/polybar/commit/f0639ab5eb5d2035c077982f13afe7485ec0405b) together to try to implement this. I don't really know if it does the right thing in other environments and it looks a bit goofy. I can fix up the styling and make a pull request if the implementation seems sound. Otherwise, whoever else does this might find it useful as a reference.

patrick96:
Thanks for the effort. I just configured my interface to also have an ipv6 address and it seems to work fine for me. I think with a little bit of cleaning up, this could be merged. Would you be willing to work a little more on this? If so, please add that commit to a PR, I will look at it a little more thorough and see what we need to change before merging.

sqwishy:
https://github.com/jaagr/polybar/pull/1239 Let me know what your suggestions are. I also put some of my own concerns in the description.

PR #1239, opened 2018-05-17
Teach net module how to query an IPv6 address.

sqwishy:
This should probably throw network_error instead of call `perror()` maybe?

Also, I think if you lose addressing but the interface is still up, the value on m_status isn't unset so the address is still displayed. I think the IPv4 thingy does that too.

I'm also not sure if this works on FreeBSD.

EDIT: Fixes #1234

### DIFF
--- a/src/adapters/net.cpp
+++ b/src/adapters/net.cpp
@@ -37,12 +37,14 @@ namespace net {
     return file_util::exists("/sys/class/net/" + ifname + "/wireless");
   }
 
+  static const string NO_IP6 = string("N/A");
+
   // class : network {{{
 
   /**
    * Construct network interface
    */
-  network::network(string interface) : m_interface(move(interface)) {
+  network::network(string interface) : m_log(logger::make()), m_interface(move(interface)) {
     if (if_nametoindex(m_interface.c_str()) == 0) {
       throw network_error("Invalid network interface \"" + m_interface + "\"");
     }
@@ -103,7 +105,14 @@ namespace net {
 
     freeifaddrs(ifaddr);
 
-    this->query_ip6();
+    try {
+      this->query_ip6();
+    } catch (const network_error& err) {
+      if (m_ip6_last_error != err.what()) {
+        m_ip6_last_error = err.what();
+        m_log.warn("net/query_ip6: %s", err.what());
+      }
+    }
 
     return true;
   }
@@ -125,35 +134,31 @@ namespace net {
     struct sockaddr_in6 a;
     socklen_t l = sizeof (struct sockaddr_in6);
 
+    m_status.ip6 = NO_IP6;
+
     memset(&a, 0, l);
     a.sin6_family = AF_INET6;
     a.sin6_port = htons(0);
     if (inet_pton(AF_INET6, "2001:7fd::1", &a.sin6_addr) == -1) {
-      perror("inet_pton()");
-      return;
+      throw network_error("inet_pton() " + string(strerror(errno)));
     }
 
     if ((fd = socket(AF_INET6, SOCK_DGRAM, 0)) == -1) {
-      perror("socket()");
-      return;
+      throw network_error("socket() " + string(strerror(errno)));
     }
 
     if (connect(fd, (const sockaddr*)&a, l) == -1) {
-      perror("connect()");
-      close(fd);
-      return;
+      throw network_error("connect() " + string(strerror(errno)));
     }
 
     if (getsockname(fd, (struct sockaddr*)&a, &l) == -1) {
-      perror("getsockname()");
-      close(fd);
-      return;
+      throw network_error("getsockname() " + string(strerror(errno)));
     }
 
     close(fd);
+
     if (inet_ntop(AF_INET6, &a.sin6_addr, ip6_buffer, NI_MAXHOST) == 0) {
-      perror("inet_ntop()");
-      return;
+      throw network_error("inet_ntop() " + string(strerror(errno)));
     }
 
     m_status.ip6 = string(ip6_buffer);
--- a/src/adapters/net.cpp
+++ b/src/adapters/net.cpp
@@ -7,6 +7,7 @@
 #include <sstream>
 #include <utility>
 
+#include <arpa/inet.h>
 #include <linux/ethtool.h>
 #include <linux/if_link.h>
 #include <linux/sockios.h>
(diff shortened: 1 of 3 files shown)


patrick96:
Thank you very much for contributing, I've added my comments above.

sqwishy:
>I wonder if the AF_PACKET case in the query method also counts the data sent via the ipv6 address or if we need extra logic to capture the up and down speeds for ipv6

I'm pretty sure those statistics include everything on that link regardless of the protocol. It increases accordingly if I do a speed test over IPv6. You can send raw ethernet frames full of null bytes with something like, `seq 3000 | xargs -n1 -P12 -i socat -u OPEN:/dev/zero,readbytes=1234 INTERFACE:eno1`. Even though those packets don't mean anything, they appear to be included in the link statistics and bump up the reported send rate a little bit in polybar.

patrick96:
Alright, so we don't need extra logic for this, I'm relieved. This makes this really only about finding the ipv6 address

sqwishy:
It seems like the message deduplication should be a logging feature, rather than live in the network module adapter thing where I put it. What I imagine is a new `logger` subclass where output tests the message against the previous, exits if they match, otherwise saves that messages the last and either calls the base class `output()` or calls `output()` on a "parent" logger in order to go through filtering (such as log level) of the singleton logger. But I didn't try this because I didn't want to get carried away.

Anyway, I think my last changeset addressed your concerns, let me know if I missed something or there's anything else.

src/adapters/net.cpp
```diff
@@ -102,9 +105,67 @@ namespace net {
 
     freeifaddrs(ifaddr);
 
+    try {
+      this->query_ip6();
```
patrick96: Please reset the `m_ip6_last_error` after every successful ipv6 query so that we get warnings message after we lose the ipv6 address *again*

sqwishy: Forgot about that basic part while I was busy overthinking this. :)

>on my machine, I get ::1 for the ipv6 address
That seems weird. That's loopback and should only be on your `lo` interface. These changes don't do anything interface specific (although they probably ought to). But it sounds like Linux is saying your IPv6 address is going to be sourced from your loopback address, which is weird.

sqwishy: Yeah I'm not entirely sure what to do about the IPv6 address information showing up for all interfaces. It's almost better if the IPv6 thing is a separate module to do with routing or something that shows you what source address and interface it would use to get somewhere.

patrick96: Oh man! Totally didn't occur to me that this will give the same ipv6 address independent of the interface :frowning_face:
So this could only be used for generic ipv6 connectivity (without specifying the interface)
From what I've seen, to open a socket bound to a specific interface you need to have root and so I think the only option is to just use `getifaddrs` as we already do for ipv4 and use any of the ipv6 addresses on that interface. Technically that's what we do for ipv4 as well, because an interface can also have multiple ipv4 addresses.

sqwishy: The problem with getifaddrs is that I kept getting the wrong address. Even if I filtered out link local and filtered by their scope. But curl seems to use getifaddrs successfully. They're probably doing something right that I missed. So I'll look at that later and if I figure it out I'll rewrite this.

patrick96: If an interface has multiple ipv6 addresses (with global scope), how can any of them be the *right* one?

From looking at curl [`lib/if2ip.c`](https://github.com/curl/curl/blob/master/lib/if2ip.c#L142), it filters addresses out by scope, but depending on the destination address (which we don't have).
How many ipv6 addresses does your interface have, that you are getting so many different addresses?
I would have said that it is a safe heuristic it to always use an address with global scope (if available) and otherwise choose any of the other addresses, because most of our users will have an ipv6 address with global scope (if they have an ipv6 address). But if you still pick the "wrong" address after that, then maybe it's not such a good heuristic.

sqwishy: Turns out maybe avahi-daemon wasn't responsible for the extra addresses, maybe it was just logging about them innocently, because I stopped the service and rebooted and still have these extra weird addresses. I'm thinking that NetworkManager is doing something. But I'm not sure why. As far as I can tell, SLAAC shouldn't be enabled. With the previous change, polybar displays the wrong address. Something I might look at is what `ip route get to 2001:7fd::1 dev eno1` does, but netlink can be a bit gross to deal with.

patrick96: > The one I use to talk to the internet. :)

Yeah obviously, but isn't it possible that a single interface has multiple IPs with which it can reach the internet? On my Raspberry PI, I have three global ipv6 addresses on the same interface.

> Something I might look at is what `ip route get to 2001:7fd::1 dev eno1` does, but netlink can be a bit gross to deal with.

Yeah, or `ping -6 -I eno1 2001:7fd::1` will also give you an ipv6 address that can reach the internet.

patrick96: Just checked and I can ping `2001:7fd::1` with two of the three addressed. It seems we might have to inspect some routing tables in order to determine which IPv6 can reach the internet. Worst case, we can call `ip route get to 2001:7fd::1 dev eno1` and inspect the output

src/modules/network.cpp
```diff
@@ -117,6 +117,7 @@ namespace modules {
       label->reset_tokens();
       label->replace_token("%ifname%", m_interface);
       label->replace_token("%local_ip%", network->ip());
+      label->replace_token("%local_ip6%", network->ip6());
```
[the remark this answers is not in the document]
sqwishy: Maybe if you treat "local" here is being used to mean "private". I believe the compliment of "local" is "remote" not "global". And technically you can have a public/globally routable ipv4 address on the interface and it would show up in polybar. So "local_ip" really does mean local, as opposed to meaning "private". So I kinda think local is technically correct.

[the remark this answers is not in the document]
patrick96: Fair enough :)

sqwishy:
Sorry for the delay. On my machine, it doesn't yet display how I'd like it to, but I don't know if I'll have much time to work on this in the near future. It's also possible that my machine's configuration is a little bananas and this will work for most everyone who cares. So I'd be happy if it were merged it in. If I revisit it in the future I can open a new pull request or something.

patrick96:
Alright, going to merge this then. I think the solution is good enough. Thank you!

PR #1241, opened 2018-05-18
updated the Readme file with the new logo & Uploaded the SVG & PNG files inside New Icon-Logo folder

Tobaloidee:


NBonaparte:
It would probably be best if we had the SVG and PNG files in the repo itself.

Tobaloidee:
I just uploaded the SVG files inside "New Icon-Logo" folder Sir. Thanks a lot!

Tobaloidee:
You're most welcome Sir @patrick96  thank you for accepting the Logo design.. I am happy to have contributed in your great project!

NBonaparte:
How did the repo size increase that much? At the very least we could have just the SVG file and display it through markdown in the README and that would only cost a few KB.

patrick96:
The size didn't increase that much, I was looking at how much space the files I downloaded from google drive took, but here only the pngs and svgs were uploaded (no `.ai` files), which only add a few hundred KB, I already added a note to my original comment. But nonetheless, I don't like cluttering the main repo with files unrelated to the actual code.

NBonaparte:
Probably doesn't matter too much, as bspwm and i3 keep their logo in their repo. Right now, I think we should pick one of them and make that the official one, and we can store the rest elsewhere if needed.

patrick96:
Only storing the image we use for the README in the repo is fine with me.

patrick96:
I have now uploaded the banner that was in the original commit, but now stored it in the repo.

From the logos presented here:


It is the banner in the left column, I find the one in the middle column quite appealing as well. Which one do you prefer @NBonaparte?

NBonaparte:
I think the one on the left works best as well, especially when displaying in the README. In that case, we can use the one in the top right as an icon (for the Github organization, etc.)

PR #1249, opened 2018-05-20
Add a question regarding compiler to `build.sh`.

doronbehar:
Make question enable the user to use GCC as the compiler even if clang
is installed on the machine.


patrick96:
LGTM

patrick96:
Thanks!

PR #1261, opened 2018-05-29
feat(builder) minlen

M1dgard:
The counterpart of maxlen, that forces a minimum length on a string.
Pads right with spaces if a string is too short.

This allows for cool stuff like this (note the title):
screenshot_2018-05-29_15-34-52
screenshot_2018-05-29_15-35-03

Using the `%{O-…}` tag, you can hack complex stuff like this together in configs. The bar below is the position in the song. The minlen and maxlen of the song-label are both set to 25 in this case, to make the length fixed.

I also applied a patch to use `…` instead of `...`, but that's not included in this PR.

---

This PR depends on (please merge those first):
* [ ] PR #1198

---

My C++ has rusted a bit and the technical quality of this code may be insufficient. Please review thoroughly. Don't hesitate to request changes or make them yourself.


M1dgard:
> This should probably be consistent with that

Do you mean the `minlen` option should pad on the left too?

> unless a new option is added for padding side.

That would be preferable to padding on the left. Padding on the left wouldn't be nice for my use case.

M1dgard:
Heads up: I am going to delete my GitHub account in a few days to prepare for Microsoft's GitHub acquisition. If this PR isn't converted to a branch on your project, it will close.

patrick96:
I pushed the branch to [`jaagr/minlen`](https://github.com/jaagr/polybar/tree/minlen). I don't have the time to finish this PR though. We will keep it in the repo though, if we ever want to get back to it.

patrick96:
Closing in favor of #1546

PR #1263, opened 2018-05-31
add support for MPD_TAG_ALBUM_ARTIST

Noiseko:


Noiseko:
It doesn't want to compile with the string having a hyphen in it.

PR #725, opened 2017-09-05
[WIP] - Add Unknown battery status

krakn:
Initial work to  make `polybar` aware of an "Unknown" battery status.

I believe there needs to be a slight change made to the logic of how `polybar` gets the battery's state:

```
 » cat /sys/class/power_supply/BAT0/status 
Unknown
 » cat /sys/class/power_supply/BAT1/status 
Charging
 » 
```

I need to grok the code a bit more before I can complete this PR. Any help is appreciated, as my knowledge of CPP isn't the strongest.

PR #1286, opened 2018-06-15
Add unknown battery status

TomKiesow:
contains all changes from [this pull request](https://github.com/jaagr/polybar/pull/725), plus some changes to the state_reader of the battery module, which now returns an int instead of a boolean. Tested on a dual battery laptop.

if youre wondering why i didnt commit to the pull request about this (if thats even possible), its because i dont know how to do that.

### DIFF
--- a/include/modules/battery.hpp
+++ b/include/modules/battery.hpp
@@ -13,6 +13,7 @@ namespace modules {
       CHARGING,
       DISCHARGING,
       FULL,
+      UNKNOWN,
     };
 
     enum class value {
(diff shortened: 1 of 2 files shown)


include/modules/battery.hpp
```diff
@@ -39,7 +40,7 @@ namespace modules {
       const function<ReturnType()> m_fn;
     };
 
-    using state_reader = mutex_wrapper<value_reader<bool /* is_charging */>>;
+    using state_reader = mutex_wrapper<value_reader<int /* current state */>>; // 0 = discharging; 1 = charging; 2 = full; 3 = unknown
```
Lomadriel: The state enum should be used here.

PR #1304, opened 2018-07-03
extend build.sh

laur89:
- add option flags;
  - among others, flag --auto allows executing
    script in a non-interactive mode, where all
    unset options are defaulted automaticaly;
- add help/usage;

This allows installing polybar non-interactively from various system scripts, eg users could
`./build.sh --auto --all-features --no-install-config`

Also fixes bug with install prompts - prompt shows default Y (`[Y/n]`), but in reality `Y` is not assigned to $p when no input is given, meaning subsequent ` $p == Y` check fails:

```
  read -r -p "$(msg "Execute 'sudo make install'? [Y/n] ")" -n 1 p && echo
  [[ "${p^^}" == "Y" ]] && {   # <-- !! $p would be empty if user gave no input !!
    sudo make install || msg_err "Failed to install executables..."
  }

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -44,8 +44,8 @@ usage() {
           Remove existing build dir; disabled by default.
       ${COLORS[GREEN]}-I, --no-install${COLORS[OFF]}
           Do not execute 'sudo make install'; enabled by default.
-      ${COLORS[GREEN]}-C, --no-install-config${COLORS[OFF]}
-          Do not install example configuration; enabled by default.
+      ${COLORS[GREEN]}-C, --install-config${COLORS[OFF]}
+          Install example configuration; disabled by default.
       ${COLORS[GREEN]}-A, --auto${COLORS[OFF]}
           Automatic, non-interactive installation; disabled by default.
           When set, script defaults options not explicitly set.
@@ -68,7 +68,7 @@ install() {
 
   if [[ "$AUTO" == ON ]]; then
     [[ -z "$INSTALL" ]] && INSTALL="ON"
-    [[ -z "$INSTALL_CONF" ]] && INSTALL_CONF="ON"
+    [[ -z "$INSTALL_CONF" ]] && INSTALL_CONF="OFF"
   fi
 
   if [[ -z "$INSTALL" ]]; then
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: build.sh [hunk 1]
```diff
@@ -1,23 +1,180 @@
...
+
+  ${COLORS[GREEN]}${COLORS[BOLD]}Usage:${COLORS[OFF]}
+      ${COLORS[CYAN]}${SELF}${COLORS[OFF]} [options]
+
+  ${COLORS[GREEN]}${COLORS[BOLD]}Options:${COLORS[OFF]}
+      ${COLORS[GREEN]}-3, --i3${COLORS[OFF]}
+          Include support for internal/i3 (requires i3); disabled by default.
+      ${COLORS[GREEN]}-a, --alsa${COLORS[OFF]}
+          Include support for internal/alsa (requires alsalib); disabled by default.
+      ${COLORS[GREEN]}-p, --pulse-audio${COLORS[OFF]}
```
NBonaparte: Should be kept as `--pulseaudio`, or maybe `--pulse` for short?

build.sh
```diff
@@ -28,80 +185,80 @@ function main
...
+    -A|--auto)
+      AUTO=ON; shift ;;
+    -h|--help)
+      usage
+      exit 0
+      ;;
+    --) shift; break ;;
+    *)
+      usage
+      [[ "$1" =~ ^-[0-9a-zA-Z]{2,}$ ]] && msg_err "don't combine options: ie do [-c -i] instead of [-ci]" || msg_err "unknown option [$1]"
```
NBonaparte: Probably would be nicer if this was supported, `getopt` would work for this.

laur89: But then we lose long format; arg parsing has never been too straight-forward with bash. I'd rather leave as-is - not what users might expect, but at least error message explicitly instructs not to bundle opts.

laur89:
@NBonaparte are you OK with current state?

build.sh
```diff
@@ -1,23 +1,180 @@
...
+    [[ -z "$INSTALL_CONF" ]] && INSTALL_CONF="ON"
+  fi
+
+  if [[ -z "$INSTALL" ]]; then
+    read -r -p "$(msg "Execute 'sudo make install'? [Y/n] ")" -n 1 p && echo
+    [[ "${p^^}" != "N" ]] && INSTALL="ON" || INSTALL="OFF"
+  fi
+
+  if [[ -z "$INSTALL_CONF" ]]; then
+    read -r -p "$(msg "Install example configuration? [Y/n]: ")" -n 1 p && echo
```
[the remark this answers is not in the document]
NBonaparte: We should definitely change this to `[y/N]`. Other than that, this can be merged.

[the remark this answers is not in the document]
laur89: Pushed. If all good, i'll squash 'em together.

Issue #1181, opened 2018-04-19
Battery Module - Inotify fails after supend


koppa:
Version stable as well as current master branch is affected

My device is a Thinkpad X220i. When using the battery,  the battery module sometimes fails to track the battery level. Error message:

```
error: module/battery: Error while creating inotify watch (what: Failed to attach inotify watch (reason: No such file or directory))
```

On further debugging I found out that the battery module has two inotify watches:

* /sys/class/power_supply/AC/online,1
* /sys/class/power_supply/BAT0/charge_now,1           

and after the suspend the charge_now device doesn't exist anymore
```
 % ls
alarm               energy_now     status
capacity            manufacturer   subsystem@
capacity_level      model_name     technology
cycle_count         power/         type
device@             power_now      uevent
energy_full         present        voltage_min_design
energy_full_design  serial_number  voltage_now
```

Probably this code has to be executed again (excerpt from battery_module::battery_module):

    if ((m_fcapnow = file_util::pick({m_path_battery + "charge_now", m_path_battery + "energy_now"})).empty()) {
      throw module_error("No suitable way to get current capacity value");
    } else if ((m_fcapfull = file_util::pick({m_path_battery + "charge_full", m_path_battery + "energy_full"})).empty()) {
      throw module_error("No suitable way to get max capacity value");
    }

patrick96:
Does the `charge_now` file exist before you suspend, but vanishes after?

koppa:
It seems so, but it doesn't vanish on every suspend.

I'm not sure, if this is some bug inside the `battery` linux module

patrick96:
I'm assuming when it doesn't vanish you don't get an error in polybar, is that right?

Seems odd to me that such a file would vanish, don't know if there are some specifications for those files. In the beginning, does your system have both `energy_now` and `charge_now` files? Or does `charge_now` vanish and `energy_now` appear?

> BTW: the battery state is reported correctly with missing charge_now and restarted polybar

Polybar will choose between `charge_now` and `energy_now` (whichever exists) when the module is loaded.

koppa:
My distro is a current arch linux 64bit with a 4.16.3-1-ARCH kernel, but the problem exists for a while.

> I'm assuming when it doesn't vanish you don't get an error in polybar, is that right?

Yeah it works correctly.

[...]

Issue #839, opened 2017-11-12
Battery remaining time gets stuck at 00:00 until bar is restarted


AppIns:
I noticed that when I suspend my laptop, my battery's remaining life will get stuck at 00:00 (formatted %hour%:%minute%) and will stay that way until the bar is restarted.

patrick96:
What is the output of `acpi` when that happens?

AppIns:
`Battery 0: Discharging, 96%, 24:19:36 remaining`
Also I got a screenshot, sometimes it keeps the percentage, other times it doesn't (this time it went to 0%)
![screenshot from 2017-11-12 14-29-41](https://user-images.githubusercontent.com/13598541/32704192-0c4eb074-c7b6-11e7-86ef-e62fc742c46f.png)

EnzDev:
It happens for me when the computer get out of sleeping.

AppIns:
Yep. If someone wants to reproduce this issue, try plugging in your laptop, sleeping it, unplugging, and then waking it up. If that doesn't work, try mixing up the order. Sometimes all it takes is just sleeping the laptop.

patrick96:
What is the output of the following command, when polybar behaves this way:
```bash
find -L /sys/class/power_supply -maxdepth 2 -type f -print0 2>/dev/null | xargs -0 -I '{}' sh -c 'echo "{}:"; cat {}; echo'
```
This command lists the contents of all the files in the `/sys/class/power_supply` folder.

EnzDev:
Everythin seems normal when it happen. See [Gist](https://gist.github.com/EnzDev/9ebd76d09cde35a5494f10a71c874097).

patrick96:
Yes indeed, it seems everything is in order in those files. What version of polybar are you running? (`polybar -vvv`)
Can you also run polybar with `-l trace`, reproduce this and post the whole output here. If trace logging is not enabled, you'll need to recompile polybar in debug mode (You can do this by adding the `-DCMAKE_BUILD_TYPE=Debug` flag to cmake)

EnzDev:
I'm using the last git tag (3.0.5).
While trying to reproduce the bug with a freshly compiled polybar, I ran into a very similar problem (which I guess has the same origin). 
The battery module just stopped updating (Stucked at `46% (00:57)`) and looped with the debug message
`* module/battery: Polling values (inotify fallback)`
so polybar just stopped rendering (because the config just contains the battery module).
[...]

PR #1306, opened 2018-07-06
Fix battery refresh when swapping power /(issue #839 and #1181)

koppa:
Here is a fix for correcting the behavior of polybar when suspending.

The paths for monitoring the battery state can change in a running system,
for example after a suspend.

Therefore now the battery module now tries to find the correct battery paths
after a failing inotify watch.

For the implementation i merged the logic of the inotify_module into the battery modules
for an easy fix.

### DIFF
--- a/include/modules/battery.hpp
+++ b/include/modules/battery.hpp
@@ -1,12 +1,12 @@
 #pragma once
 
 #include "common.hpp"
-#include "modules/meta/inotify_module.hpp"
+#include "meta/base.hpp"
 
 POLYBAR_NS
 
 namespace modules {
-  class battery_module : public inotify_module<battery_module> {
+  class battery_module : public module<battery_module> {
    public:
     enum class state {
       NONE = 0,
@@ -88,12 +88,15 @@ namespace modules {
     animation_t m_animation_discharging;
     progressbar_t m_bar_capacity;
     ramp_t m_ramp_capacity;
+    map<string, int> m_watchlist;
 
     string m_fstate;
     string m_fcapnow;
     string m_fcapfull;
     string m_frate;
     string m_fvoltage;
+    string path_battery;
+    string path_adapter;
 
     state m_state{state::DISCHARGING};
     int m_percentage{0};
@@ -104,7 +107,11 @@ namespace modules {
     chrono::duration<double> m_interval{};
     chrono::system_clock::time_point m_lastpoll;
     thread m_subthread;
+
+    void runner();
+    void poll_events();
+    void recreate_watches();
   };
-}
+}  // namespace modules
 
 POLYBAR_NS_END
--- a/include/modules/meta/inotify_module.hpp
+++ b/include/modules/meta/inotify_module.hpp
@@ -8,16 +8,7 @@ POLYBAR_NS
 namespace modules {
   template <class Impl>
   class inotify_module : public module<Impl> {
-   public:
-    using module<Impl>::module;
-
-    void start() {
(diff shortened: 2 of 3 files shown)


include/modules/battery.hpp
```diff
@@ -104,7 +107,11 @@ namespace modules {
     chrono::duration<double> m_interval{};
     chrono::system_clock::time_point m_lastpoll;
     thread m_subthread;
+
+    void runner();
+    void poll_events();
+    void recreate_watches();
   };
-}
+}  // namespace modules
```
f0s3: Remove useless comment please, of provide a reason for it to stay here.

include/modules/meta/inotify_module.hpp
```diff
@@ -8,16 +8,7 @@ POLYBAR_NS
...
-    using module<Impl>::module;
-
-    void start() {
-      this->m_mainthread = thread(&inotify_module::runner, this);
-    }
-
-   protected:
-    void runner() {
-      this->m_log.trace("%s: Thread id = %i", this->name(), concurrency_util::thread_id(this_thread::get_id()));
+   public: using module<Impl>::module; void start() { this->m_mainthread = thread(&inotify_module::runner, this); } protected: void runner() { this->m_log.trace("%s: Thread id = %i", this->name(), concurrency_util::thread_id(this_thread::get_id()));
```
f0s3: Leave the indentation as it was, it is difficult to review and, possibly, introduces useless changes.

src/modules/battery.cpp
```diff
@@ -91,16 +67,16 @@ namespace modules {
...
+        // if it was power, just use as is
       } else {
         unsigned long power{std::strtoul(file_util::contents(m_frate).c_str(), nullptr, 10)};
 
         consumption = power / 1e6;
       }
 
       // convert to string with 2 decimmal places
-      string rtn(16, '\0'); // 16 should be plenty big. Cant see it needing more than 6/7..
+      string rtn(16, '\0');  // 16 should be plenty big. Cant see it needing more than 6/7..
```
f0s3: Please do not introduce indentation changes if it does not fix the issue. Code cleanup and refactoring is a good thing, but if you want your branch to be merged, you need to keep it clean and create another branch and PR for code cleanup.

patrick96:
@f0s3 Thanks for the review. I will also take a last look once I have some time.

The formatting changes are introduced by `clang-format` and we *want* people to run it on the files they have changed: https://github.com/polybar/polybar/wiki/Style-Guide#code-formatting

We are currently trying to get all our code formatted with `clang-format` but because we don't want to introduce a bunch of conflicts in all open PRs we didn't want to just format everything at once, so we ask contributors to format the files they changed with `clang-format`. This way all files get formatted one-by-one and at some point we can then run `clang-format` to also format the files that are never touched by anyone.

Though the one where everything moved onto the same line seems out of place and shouldn't be something that `clang-format` does.

include/modules/battery.hpp
patrick96: This is fine

include/modules/meta/inotify_module.hpp
patrick96: Strange that it was all put on the same line. Did `clang-format` do that?

src/modules/battery.cpp
patrick96: This is also fine, see my other comment :)

pstch:
I'm using this patch on my laptop, and it seems to be working fine (fixes the problem in #1937).

f0s3 (approved):

Issue #758, opened 2017-09-16
i3 workspace back_and_forth on mouse click


schaden-freude:
In i3bar, it is possible to switch to the previously active workspace by clicking on the current workspace icon. Is it possible to replicate that in polybar?

I tried `format = %{A1:i3-msg workspace back_and_forth:}<label-state>%{A}` but that doesn't work flawless. e.g. It switches to previous workspace only when clicked on the ws-1 icon.

patrick96:
Theoretically this should be possible by surrounding `label-focused` with the `back_and_forth` action. But this would nest actions which isn't handled that great currently, so this doesn't work right now.

I'll open up a separate issue to address the nesting issue, once that is fixed, this issue should be resolved too.

patrick96:
Hi, I have just pushed a fix for the nested action issue. Now using
```dosini
label-focused = %{A1:i3-msg workspace back_and_forth:}  %index%  %{A}
label-focused-padding = 0
```
in your config should work as expected.
*Note:* You need to set the padding to 0 because the padding is not part of the label and thus there would be no action block on the padding only on the text itself.

Please reopen if this still doesn't work with the newest version

schaden-freude:
Hey! This is not exactly related, but how do I update to the newest version?

Should I remove the entire thing, clone and build again? Because it says it's already up to date.

`$ git pull`
`From https://github.com/jaagr/polybar`
`   0a5ad9c..5b7d7b8 master     -> origin/master`
`You are not currently on a branch.`
`Please specify which branch you want to merge with.`


`$ git pull https://github.com/jaagr/polybar 3.0.5`
`From https://github.com/jaagr/polybar`
 `* tag               3.0.5      -> FETCH_HEAD`
`Already up-to-date.`

PR #1312, opened 2018-07-12
fix(i3): Send workspace cmd unconditionally on click

softmoth:
This makes polybar play nice with the i3wm feature `workspace_auto_back_and_forth`, and it is harmless when that option isn't set.


patrick96:
Thanks, this should make the approach in #758 obsolete.

Issue #99, opened 2016-10-15
Clickable areas in custom/script


birjolaxew:
Would it be possible to create a way for external scripts to provide multiple clickable areas in the shown text? Using lemonbar-like tokens would probably be the easiest.

My specific use case is a wifi menu; a clickable `internal/network` which shows available networks when clicked, and connects when one is clicked, similar to a `custom/menu`. This would require clickable areas to simulate a menu, which isn't possible using only left/middle/right clicking.

jaagr:
You can already use lemonbar tokens. All %{...} tags used in lemonbar is implemented.

birjolaxew:
Ah, neat. Is that mentioned on the wiki?

If I were to use a menu, how would I make it so that the menu items (and `label-open`) are given by the output of an external script? Based on the wiki, I don't see a way to do that.

jaagr:
Oh I must have misunderstood you. That's not possible with the current module..

birjolaxew:
Thanks for the help anyway - support of lemonbar tokens lets me code it myself :)

Tmplt:
Reopening this: is this still possible in 2.0 and above now that the bar is internal?

NBonaparte:
Yes.

PR #1325, opened 2018-07-20
Added swap memory support for bar-used

roel0:
bar-used currently only supports internal memory, added support to have a bar for swap

### DIFF
--- a/src/modules/memory.cpp
+++ b/src/modules/memory.cpp
@@ -18,9 +18,9 @@ namespace modules {
   memory_module::memory_module(const bar_settings& bar, string name_) : timer_module<memory_module>(bar, move(name_)) {
     m_interval = m_conf.get<decltype(m_interval)>(name(), "interval", 1s);
 
-    m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL, TAG_BAR_USED, TAG_BAR_FREE, TAG_RAMP_USED, TAG_RAMP_FREE});
+    m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL, TAG_BAR_USED, TAG_BAR_FREE, TAG_RAMP_USED, TAG_RAMP_FREE, 
+                                                 TAG_BAR_SWAP_USED, TAG_BAR_SWAP_FREE, TAG_RAMP_SWAP_USED, TAG_RAMP_SWAP_FREE});
 
-    m_type = m_conf.get<decltype(m_type)>(name(), "mem-type", "RAM");
     if (m_formatter->has(TAG_BAR_USED)) {
       m_bar_memused = load_progressbar(m_bar, m_conf, name(), TAG_BAR_USED);
     }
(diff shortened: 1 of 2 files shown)


src/modules/memory.cpp
```diff
@@ -107,7 +107,12 @@ namespace modules {
 
   bool memory_module::build(builder* builder, const string& tag) const {
     if (tag == TAG_BAR_USED) {
-      builder->node(m_bar_memused->output(m_perc_memused));
+      string m_type = m_conf.get<decltype(m_type)>(name(), "bar-used-type", "RAM");
```
NBonaparte: This retrieves the variable from the config every single time the module is built, which isn't very efficient. It would be better if m_type was set from the initialization and only checked in `build()`, or create a new tag like `TAG_BAR_USED_SWAP` altogether.

maxdevjs:
Will this be merged?

src/modules/memory.cpp
roel0: done

NBonaparte:
The only issue with this method is that RAM and swap memory can't both be displayed in the same bar. I'm not sure how many people would use this though. @patrick96 do you think having both types available would be better?

maxdevjs:
screenshot-2018-08-28_15 08 17_-03-03 00_1535479697

sigh...

roel0:
Example config:

`[module/memory]
type = internal/memory
interval = 0.2
format = <label> <bar-used>

label = RAM
label-font = 3
label-foreground = #99

bar-used-width = 15
bar-used-gradient = true
bar-used-foreground-0 = #449f3d
bar-used-foreground-1 = #2f8419
bar-used-foreground-2 = #f5a70a
bar-used-foreground-3 = #ed5456
bar-used-indicator = █
bar-used-indicator-font = 5
bar-used-indicator-foreground = #fff
bar-used-fill = █
bar-used-fill-font = 5
bar-used-empty = █
bar-used-empty-font = 5
bar-used-empty-foreground = #44

[module/swap]
type = internal/memory
interval = 0.2
format = <label> <bar-used>

label = SWAP
label-font = 3
label-foreground = #99

bar-used-type = SWAP
bar-used-width = 15
bar-used-gradient = true
bar-used-foreground-0 = #449f3d
bar-used-foreground-1 = #2f8419
bar-used-foreground-2 = #f5a70a
bar-used-foreground-3 = #ed5456
bar-used-indicator = █
bar-used-indicator-font = 5
bar-used-indicator-foreground = #fff
bar-used-fill = █
bar-used-fill-font = 5
bar-used-empty = █
bar-used-empty-font = 5
bar-used-empty-foreground = #44
`

maxdevjs:
Thank you @roel0 , currently works only with your fork, right?

roel0:
@maxdevjs  correct

patrick96:
Why not use `<bar-swap-used>` and `<bar-swap-free>` etc.? This would be in the same vein as the swap tokens (`%percentage_swap_used%` etc.) we added. I think this would be cleaner codewise anyways, even if it uses some more code than the current approach. And people could use both in the same module.

roel0:
@NBonaparte  @patrick96  done

patrick96:
Much better! Code looks good, will run it with some configurations in a few days just to make sure.

patrick96:
Thanks for contributing!

Issue #1335, opened 2018-07-21
KeyError: 'eventstruct' error when trying to compile


EgbertMedia:
I'm getting an KeyError when I try to build using the build script.
The fix in the known issues list is not working for me, I'm still getting the same error.

[  1%] Linking CXX static library libxpp.a
Traceback (most recent call last):
  File "/home/egbert/Downloads/polybar/lib/xpp/generators/cpp_client.py", line 3163, in <module>
    from xcbgen.state import Module
  File "/usr/lib/python2.7/site-packages/xcbgen/state.py", line 7, in <module>
    from xcbgen import matcher
  File "/usr/lib/python2.7/site-packages/xcbgen/matcher.py", line 12, in <module>
    from xcbgen.xtypes import *
  File "/usr/lib/python2.7/site-packages/xcbgen/xtypes.py", line 1221, in <module>
    class EventStruct(Union):
  File "/usr/lib/python2.7/site-packages/xcbgen/xtypes.py", line 1239, in EventStruct
    out = __main__.output['eventstruct']
KeyError: 'eventstruct'
make[2]: *** [lib/xpp/CMakeFiles/xpp.dir/build.make:70: lib/xpp/libxpp.a] Error 1
make[1]: *** [CMakeFiles/Makefile2:403: lib/xpp/CMakeFiles/xpp.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
** Failed to build project

patrick96:
See [Known-Issues](https://github.com/jaagr/polybar/wiki/Known-Issues)

EgbertMedia:
Like I said, git checkout in the lib/xpp folder didn't fix the issue for me.
I'm using fedora.

EgbertMedia:
Like I said, git checkout in the lib/xpp folder did not fix the issue for me.
Is there something else I should try? I'm using fedora.

patrick96:
Did you also exectue `rm include/xpp/proto/*` in the `lib/xpp` folder?

patrick96:
#1330 had the same problem. The fix described doesn't work when you use `build.sh`.

EgbertMedia:
Thanks! I have compiled it manually now and it does work fine now.

Issue #1330, opened 2018-07-20
KeyError: 'eventstruct'


JsonFn:
Following the instruction to build from source, but build.sh script `build.sh -f` fails with 

    Scanning dependencies of target xpp
    [  1%] Linking CXX static library libxpp.a
    Traceback (most recent call last):
      File "/home/jason/github.com/polybar/lib/xpp/generators/cpp_client.py", line 3163, in <module>
        from xcbgen.state import Module
      File "/usr/lib/python2.7/dist-packages/xcbgen/state.py", line 7, in <module>
        from xcbgen import matcher
      File "/usr/lib/python2.7/dist-packages/xcbgen/matcher.py", line 12, in <module>
        from xcbgen.xtypes import *
      File "/usr/lib/python2.7/dist-packages/xcbgen/xtypes.py", line 1221, in <module>
        class EventStruct(Union):
      File "/usr/lib/python2.7/dist-packages/xcbgen/xtypes.py", line 1239, in EventStruct
        out = __main__.output['eventstruct']
    KeyError: 'eventstruct'
    make[2]: *** [lib/xpp/CMakeFiles/xpp.dir/build.make:70: lib/xpp/libxpp.a] Error 1
    make[1]: *** [CMakeFiles/Makefile2:403: lib/xpp/CMakeFiles/xpp.dir/all] Error 2
    make: *** [Makefile:130: all] Error 2
    ** Failed to build project

How to fix this error? 

All modules are answered with Y 

    ** Include support for "internal/i3" (requires i3) ------------------- [Y/n]: Y
    ** Include support for "internal/volume" (requires alsalib) ---------- [Y/n]: Y
    ** Include support for "internal/network" (requires wireless_tools) -- [Y/n]: Y
    ** Include support for "internal/mpd" (requires libmpdclient) -------- [Y/n]: Y
    ** Include support for "internal/github" (requires libcurl) ---------- [Y/n]: Y
    ** Build "polybar-msg" used to send ipc messages --------------------- [Y/n]: Y

And targets configured

    --  Targets:
    -- [X]   polybar-msg
    -- [ ]   testsuite
    --  Module support:
    -- [X]   alsa
    -- [X]   curl
    -- [X]   i3
    -- [X]   mpd
    -- [X]   network
    --  X extensions:
    -- [X]   xcb-randr
    -- [X]   xcb-randr (monitor support)
    -- [ ]   xcb-render
    -- [ ]   xcb-damage
    -- [ ]   xcb-sync
[...]

Issue #1294, opened 2018-06-21
error


TasX-error404:
Hello, I would like to show up after sudo make install
out = __main__.output['eventstruct']
KeyError: 'eventstruct'
make[2]: *** [lib/xpp/CMakeFiles/xpp.dir/build.make:70: lib/xpp/libxpp.a] Error 1
make[1]: *** [CMakeFiles/Makefile2:403: lib/xpp/CMakeFiles/xpp.dir/all] Error 2
make: *** [Makefile:130: all] Error 2

patrick96:
See [Known-Issues](https://github.com/jaagr/polybar/wiki/Known-Issues#keyerror-eventstruct)

Issue #1252, opened 2018-05-22
Failed to build package


DevGrohl:
I'm having problems building the package in Archlinux since a few weeks ago.
I don't usually post issues so let me know if more info it's needed.

> Traceback (most recent call last):
  File "/home/mike/.cache/pacaur/polybar/src/polybar/lib/xpp/generators/cpp_client.py", line 3163, in <module>
    from xcbgen.state import Module
  File "/usr/lib/python3.6/site-packages/xcbgen/state.py", line 7, in <module>
    from xcbgen import matcher
  File "/usr/lib/python3.6/site-packages/xcbgen/matcher.py", line 12, in <module>
    from xcbgen.xtypes import *
  File "/usr/lib/python3.6/site-packages/xcbgen/xtypes.py", line 1221, in <module>
    class EventStruct(Union):
  File "/usr/lib/python3.6/site-packages/xcbgen/xtypes.py", line 1239, in EventStruct
    out = __main__.output['eventstruct']
KeyError: 'eventstruct'
make[2]: *** [lib/xpp/CMakeFiles/xpp.dir/build.make:70: lib/xpp/libxpp.a] Error 1
make[1]: *** [CMakeFiles/Makefile2:403: lib/xpp/CMakeFiles/xpp.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
==> ERROR: A failure occurred in build().
    Aborting...
:: failed to build polybar package(s)

patrick96:
See [Known-Issues](https://github.com/jaagr/polybar/wiki/Known-Issues)

On Arch it is recommended to build `polybar` from the AUR, which doesn't have that issue

PR #1338, opened 2018-07-22
Release 3.2.0

patrick96:
I think the 3.2.0 release is badly needed right now, for one because xcb-proto 1.13 is getting more and more common, making the 'eventstruct' issue (#973) catch on (see #1335, #1330, #1294, #1252, #1233, #1090, #1114) and also 3.1.0 cannot be compiled with gcc 8 without manually updating the i3ipcpp submodule.

The [3.2.0 Milestone](https://github.com/jaagr/polybar/milestone/2) contained quite a few unfinished issues, I have moved almost all of them into the [3.3.0 Milestone](https://github.com/jaagr/polybar/milestone/4) because this release now has some urgency to it.
The only issue I kept was #1194, because there is already a fix for it (#1198), it just needs to be reviewed and merged.

With this release, I'd like to also propose some guidelines on how we do releases. The base idea is that each minor (and major) release (like 3.0.0, 3.1.0, 3.2.0, etc) gets its own branch. The name of the branch is the version number with the patch number dropped (e.g. 3.2, 4.0, etc) so that we don't get any conflicts with the git tags for the versions (that's why in the README I used `--branch 3.2`). Development continues on `master`, no actual work is done directly in the release branches. If we need to release a patch version (e.g. 3.2.1), for example because of a critical bug, we cherry-pick the commit that fixes the bug onto the current release branch (*Note:* The 3.2.1 tag is only present in the release branch, the `master` branch now only contains the tags for minor and major releases). So we basically backport the fixes, this way we don't have to create minor releases for every bug, just because we implemented a feature in the meantime. It also prevents the behaviour we have now, where we have to patch up the PKGBUILDs and write extra instructions just to get the current version to compile.

I'd like to get your feedback on these release guidelines, if you approve, I will write a page on the wiki with a release protocol.

**Things to do before merging:**
- [x] Write changelog and add it to the github release tag
- [x] Check if all new features are documented on the wiki

**Things to do after this is merged:**
- [ ] Remove `(unreleased)` note from new features in the wiki
- [ ] Create the `3.2` branch with this PR as its base commit.
- [ ] Publish drafted release with target on the `3.2` branch
- [ ] Ensure AUR packages build properly and have proper version tags
- [ ] Push updated PKGBUILDs (and .SRCINFO)

@NBonaparte could you please review #1198 so that we can fix that bug for this release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Find a more complete list on the [dedicated wiki page](https://github.com/jaagr/
 Please [report any problems](https://github.com/jaagr/polybar/issues/new) you run into when building the project.
 
   ~~~ sh
-  $ git clone --branch 3.1.0 --recursive https://github.com/jaagr/polybar
+  $ git clone --branch 3.2 --recursive https://github.com/jaagr/polybar
   $ mkdir polybar/build
   $ cd polybar/build
   $ cmake ..
--- a/contrib/polybar-git.aur/PKGBUILD
+++ b/contrib/polybar-git.aur/PKGBUILD
@@ -2,8 +2,8 @@
 # Contributor: Michael Carlberg <c@rlberg.se>
 _pkgname=polybar
 pkgname="${_pkgname}-git"
-pkgver=3.1.0
-pkgrel=3
+pkgver=3.2.0
+pkgrel=1
 pkgdesc="A fast and easy-to-use status bar"
 arch=("i686" "x86_64")
 url="https://github.com/jaagr/polybar"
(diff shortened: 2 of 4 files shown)


NBonaparte:
Sounds good.

It might also be nicer to do more frequent minor releases because we usually end up having a large backlog of features, fixes, etc. We should have some metric to determine when to release, especially if some PRs in the milestone take longer than expected.

patrick96:
I agree, we should maybe limit our milestones to 5 or 6 features and if they're done, just do a release. If only one of them is holding the others back, we just move it to the next milestone and continue with the release.

patrick96:
Will try to finally release this thing tomorrow and nicely write up the release guidelines.

patrick96:
I have written down the [guidelines](https://github.com/jaagr/polybar/wiki/Release-Guidelines) in the wiki, if you want to read them

Issue #1332, opened 2018-07-21
Zero padding


memeplex:
Specially for the cpu module that tends to be below 10% and even with `:2:3` looks inconsistent side-to-side other modules:

![image](https://user-images.githubusercontent.com/2845433/43031555-4868c778-8c7a-11e8-815a-c8522a97a33e.png)

Here, the spacing between the icon and the percentage appears a bit like the margin between modules so it's not only unaesthetic but also confusing. The alternative of not reserving space for a second digit is distracting when the glyphs are frequently moving back and forth.

The real solution is to zero pad that percentage.

Then the **real** solution is to support printf-style format strings but zero-padding is the most important use case AFAICS.

memeplex:
Here is a suggestion for a very simple syntactical extension: using `%value:02` instead of `%value:2% will left zero pad `%value`, in general using `%value:0n` will zero pad to fill at least n positions.

This is not a general padding mechanism, it adds only left zero padding. But since:

1. Left zero padding is by far the most common case for aligning numerical fields and left space padding is the most common case for text fields (and also desirable for numerical fields), so both important cases will be covered.

2. Eventually printf-style format strings will generalize this anyway.

I think is a worthy simple addition.

memeplex:
@patrick96 now that I'm on that, would you accept a PR for this too?

NBonaparte:
What is your font order in your config? If your icon font is first, it might be using the space from that which is not monospace. For me, I have my mono font first and then the icon font, and I use `%label:3%` without any problems.

memeplex:
@NBonaparte the problem is with whitespace-padding, font is monospeced and that's working fine, I just want `0` instead of ` `.

I've submitted a very simple PR to add the functionality described above. I could add a couple of lines to the documentation also.

Btw, what's the best way to quick dump some debug info while developing? cout and cerr shows nothing and it's a bit cumbersome to add and remove a logger just for that. I ended up writing to a file but I guess you have better ways.

NBonaparte:
Deleted comment because I misread it, sorry.

memeplex:
Could you help me understand why adding a couple of lines is reducing coverage by 50% (!!!!)?

I think that's due to adding the zpad field to label but I fail to see a lot to test here.

patrick96:
I agree, printf style tokens are the way to go eventually. Your suggestion sounds fine for the moment.

`cout` and `cerr` should work (they do for me). You can also use the logger, by just typing `logger::make.err(...)`.

[...]

PR #1341, opened 2018-07-22
Zero pad tokens when min spec starts with '0'

memeplex:
Fixes https://github.com/jaagr/polybar/issues/1332


patrick96:
Please add some documenting comments inside the code. Looks good though, will do some testing and merge this, after #1338 is merged

patrick96:
@NBonaparte I would have squashed this anyway because of the format of the commit message

For the first part, do whatever you feel like is clearest

PR #1342, opened 2018-07-22
Make text module a label

### LINKED ISSUE #1331: Make text module a label
The text module is too inflexible: it's not possible to set left/right padding, it's not possible to change the font, etc. Why not making it a label with a %text% token by default? This would be much more flexible and also consistent with most of the other modules (consider script, for a similar one).

memeplex:
Fixes https://github.com/jaagr/polybar/issues/1331

**Reviewer Note:** This PR introduces breaking changes to the text module. It should be merged in the merge window where breaking changes can be introduced for the next major release (generally directly before the next major release).

```
refactor(text): Make content a label (#1342)

This adds the `format` key to the text module with `<content>` as its 
only (label) tag.

This change is not backwards compatible for all configs that make use of 
the `-margin`, `-spacing`, `-prefix` or `-suffix` properties of 
`content`, because `content` is converted from a format to a label.
It would be otherwise because by default `format = <content>` and if 
only `content` is set, the module will behave as before.
This will allow users to use all the label formatting options in the 
text module as well.

Sample configuration:
[module/text]
type = custom/text
format = <content>
content = Content of the text module

Fixes #1331

```

### DIFF
--- a/src/modules/text.cpp
+++ b/src/modules/text.cpp
@@ -1,4 +1,5 @@
 #include "modules/text.hpp"
+#include "drawtypes/label.hpp"
 
 #include "modules/meta/base.inl"
 
@@ -8,18 +9,10 @@ namespace modules {
   template class module<text_module>;
 
   text_module::text_module(const bar_settings& bar, string name_) : static_module<text_module>(bar, move(name_)) {
-    m_formatter->add("content", "", {});
-
-    if (m_formatter->get("content")->value.empty()) {
-      throw module_error(name() + ".content is empty or undefined");
+    m_formatter->add(DEFAULT_FORMAT, TAG_CONTENT, {TAG_CONTENT});
+    if (m_formatter->has(TAG_CONTENT, DEFAULT_FORMAT)) {
+      m_label = load_label(m_conf, name(), TAG_CONTENT);
     }
-
-    m_formatter->get("content")->value =
-        string_util::replace_all(m_formatter->get("content")->value, " ", BUILDER_SPACE_TOKEN);
-  }
-
-  string text_module::get_format() const {
-    return "content";
   }
 
   string text_module::get_output() {
@@ -54,6 +47,16 @@ namespace modules {
 
     return m_builder->flush();
   }
+
+  bool text_module::build(builder* builder, const string& tag) const {
+    if (tag == TAG_CONTENT) {
+      builder->node(m_label);
+    } else {
+      return false;
+    }
+    return true;
+  }
+
 }
 
 POLYBAR_NS_END
--- a/include/modules/text.hpp
+++ b/include/modules/text.hpp
@@ -10,8 +10,14 @@ namespace modules {
     explicit text_module(const bar_settings&, string);
 
     void update() {}
-    string get_format() const;
+
     string get_output();
+    bool build(builder* builder, const string& tag) const;
+
+   private:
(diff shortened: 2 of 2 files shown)


memeplex:
Could you help me understand what else is missing here? I'm getting no output for the text module after the change. I assume I have to implement `build` or do something else inside `get_output` (because it's a static module (?)) but this is my first change that requires a more throughout knowledge of the overall architecture and I'm a bit lost.

memeplex:
Also I don't know what would be `m_label->replace_token(" ", BUILDER_SPACE_TOKEN);` for, I just adapted it from the previous code.

memeplex:
Ok, forget the previous questions, I had to add the label to the builder at construction time since it's a static module.

memeplex:
Well, I ended up implementing `build` since it seems the best way to avoid assuming too much about what will be in `format`. Nevertheless, the user is not intended to modify the default `<content>` but in case he/she does it everything should still work fine if:

1. The label `content` is defined
2. The format `format` includes `<content>`

Anything else will be reported as an error.

src/modules/text.cpp
```diff
@@ -8,18 +9,13 @@ namespace modules {
   template class module<text_module>;
 
   text_module::text_module(const bar_settings& bar, string name_) : static_module<text_module>(bar, move(name_)) {
-    m_formatter->add("content", "", {});
-
-    if (m_formatter->get("content")->value.empty()) {
-      throw module_error(name() + ".content is empty or undefined");
```
patrick96: This looks like very old code. No other module actually does that anymore. I don't think this needs to be adapted and can be removed. What all the other modules do, is to load the format, as you did, and if the label tag occurs in the format string, load the label. For example the script modules does it like this:
```cpp
modules/script.cpp:
-------------------
m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL});
  m_label = load_optional_label(m_conf, name(), "label", "%output%");
}
```

This avoids throwing unnecessary errors because one should be able to also define a format without including any tags (even if it's not that useful).

memeplex: In your last example

```
if (m_formatter->has(TAG_LABEL))
```

I'm unsure about removing the `DEFAULT_FORMAT` argument, as in:

```
if (m_formatter->has(TAG_CONTENT, DEFAULT_FORMAT))
```

Because the formatter could contain more than one format and the only API I found to check if a tag is inside a format is to request the formatter for that tag in one of its formats. But maybe the dafault format is considered when no other is explicitly passed?

patrick96: From looking at the code (`base.cpp:143`), if you pass only one argument to `module_formatter::has`, then it searches through all the defined formats. So in this case it shouldn't matter, because the text module only defines a single format `format`.

memeplex: Ok, I've already updated and rebased everything according to your comments.

patrick96:
I just realised that this breaks compatibility for all configs that use `content-prefix`, `-suffix`, `-spacing`, or `-margin` because right now `content` is a format and afterwards it would be a label, which doesn't have all those properties. We will keep this PR open and will merge it for the 4.0.0 release. This may be a while though.

memeplex:
Yeah, I commented a bit about that in the issue, so I was expecting this to
enter the next major version, if any. It's ok, since I posted I discovered
many workarounds like prefix/suffix labels and font tags.

PR #1343, opened 2018-07-23
Release 3.2.1

patrick96:
I want to try to release a patch release with the new process as soon as possible to work out any kinks.

This adds #1340, ~#1341~, and #1312 to version 3.2

### DIFF
--- a/contrib/polybar-git.aur/PKGBUILD
+++ b/contrib/polybar-git.aur/PKGBUILD
@@ -2,7 +2,7 @@
 # Contributor: Michael Carlberg <c@rlberg.se>
 _pkgname=polybar
 pkgname="${_pkgname}-git"
-pkgver=3.2.0
+pkgver=3.2.1
 pkgrel=1
 pkgdesc="A fast and easy-to-use status bar"
 arch=("i686" "x86_64")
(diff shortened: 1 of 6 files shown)


NBonaparte:
Should we add #1322 to this as well?

patrick96:
Sure thing

patrick96:
Done

patrick96:
Alright, I just noticed something. In the [guidelines for a patch release](https://github.com/jaagr/polybar/wiki/Release-Guidelines#patch-release), I specifically say to have a PR with the bugfixes and the PR with the release commit separate so that we can squash the commits. This introduces the problem that when the bug fixes are merged into the release branch, they are not part of a tagged release but in the README we tell the people to check out the `3.2` branch when cloning, so they would get a non-release version of polybar. I am going to change the guidelines so that we have both the bug fix commits and the release commit in the same PR. This also ensures that everytime we add some bugfixes to the release branch, we also directly release a new version. I know this is something very small, but I really like to have the release process ironed out so that we don't run into trouble later.

I will try to do all those things tomorrow.

patrick96:
Alright, this can be reviewed now

patrick96:
Seems like we shouldn't update the `polybar-git` PKGBUILD for patch releases, because the `polybar-git` package always pulls from master, and the 3.2.1 tag is not on master. Since there is no work done on this branch, I will amend and force push the release commit to remove the changes to the `polybar-git` PKGBUILD. I will also update the documentation

PR #1350, opened 2018-07-27
Fix font resolution

schra:
Situation: In Arch Linux the Siji and Fixed fonts from the example configuration didn't load. With my fixes they do.

The fixed font wasn't part of the optional dependencies. So I added them to the PKGBUILD.

In Arch Linux the fixed font is named `misc fixed` and siji `wuncon siji`. `queryfont` takes care of this naming issue. The problem was, however, that the font names weren't updated. So eg `queryfont` would recognize that `$FONT_SIJI` should be set to `wuncon siji` instead of the default value `siji` but it didn't overwrite the value of `$FONT_SIJI` and rather set (literally) `$output_variable`.


patrick96:
Thanks!

Issue #1172, opened 2018-04-17
Multiple MPD modules: second one controls wrong server


ghost:
I have a local MPD server and one on the internet and added modules for both of them, like this:
```ini
[bar/mybar]
modules-left = mpd-local mpd-online

[module/mpd-local]
type = internal/mpd

; I didn't expect this to change anything but tried with and without this anyway, no change:
host = localhost
port = 6600

format-online = local <toggle> <label-song>
icon-play     = ⏵
icon-pause    = ⏸
label-song    = %title%

[module/mpd-online]
type = internal/mpd

host     = mpd.example.org
port     = 1234
password = example

format-online = online <toggle> <label-song>
icon-play     = ⏵
icon-pause    = ⏸
label-song    = %title%
```

The first module works fine. The second one however acts up weird:

1. Initial state:  
![both players are paused](https://user-images.githubusercontent.com/4283376/38867324-3d9d2dca-4233-11e8-99fc-d66452db31c2.png)

2. Click the play button for the **online** player:  
![the local player started playing, the online one is still paused](https://user-images.githubusercontent.com/4283376/38867331-41daa02a-4233-11e8-86f0-b4337322bec2.png)  
The wrong one starts playing. It's indeed the local server that started producing sound.

3. Click the same button again:  
![both the online and the local player are now playing](https://user-images.githubusercontent.com/4283376/38867332-41fe5f4c-4233-11e8-9120-b95031c00383.png)

4. And again:  
![the local player is now paused, the online one is playing](https://user-images.githubusercontent.com/4283376/38867472-afc3e61e-4233-11e8-9b69-2a21c2712143.png)

5. Once more pauses the online server:  
![we're back to the initial state: both players are paused](https://user-images.githubusercontent.com/4283376/38867324-3d9d2dca-4233-11e8-99fc-d66452db31c2.png)

NBonaparte:
If you have only the second one running, does it work normally?

patrick96:
I was afraid of this. For modules with custom buttons (that don't execute a shell command), like the mpd, backlight, volume and even the date module, polybar sends an internal command on each button click. The problem: from the command itself you can only determine from which module type it came from not from which exact module (if you have more than one of the same type). Having more than one module of the same type can cause the modules to interfere with eachother, because the commands are generally handled by the first module that can handle them. So having two date modules will cause only the first one to toggle its label, even if you click on the second one.

[...]

PR #1353, opened 2018-07-30
Added timezones

### LINKED ISSUE #596: Select time zone for date module

### LINKED ISSUE #1293: [module/date] Zone and UTC feature support request
Hello,

That's why I request to:

- Each user may want to see their own time zone. (Like I would like to see the time of Turkey and my brother may want to see Germany time.)
- We don't have to change the `zoneinfo` information from `/usr/share/` specially
- What time zone do we want to see ? LocalTime, UTC, Network Synched, etc...
- TimeZone offset support? UTC+1, UTC+2 like that.

Edit: Oh I forgot that trick, each user can have their own time zone. :)

Currently only localtime support is available with the [date] architecture. When we do `timedatectl`, there are few outputs:

```
Local time: Wed 2018-06-20 14:02:29 +03
Universal time: Wed 2018-06-20 11:02:29 UTC
RTC time: Wed 2018-06-20 14:02:29
Time zone: Europe/Istanbul (+03, +0300)
System clock synchronized: no
```

But Polybar uses localtime everytime as long as I do not sync with localtime to network time using timezones from mentioned here https://wiki.archlinux.org/index.php/Time#Time_zone

I don't think that's a rational solution for Polybar. The time module should be further improved. The module I designed in my mind is:

```
[module/date]
type = internal/date

; New date features
timezone = Europe/Istanbul ; from /usr/share/zoneinfo
timezoneoffset = +3, +2, 0, bla bla...
datetype = UTC, RTC, Local, Universal etc...
autosync = true ; Possibility of system time not synced


; Seconds to sleep between updates
interval = 1.0

; See "http://en.cppreference.com/w/cpp/io/manip/put_time" for details on how to format the date string
; NOTE: if you want to use syntax tags here you need to use %%{...}
date = %Y-%m-%d%

; Optional time format
time = %H:%M

; if `date-alt` or `time-alt` is defined, clicking
; the module will toggle between formats
date-alt = %A, %d %B %Y
time-alt = %H:%M:%S
```

All comments are welcomed. Thanks in advance! :)

thomastay:
Added the ability to display two clocks on the date module and the ability to set the timezone on those two clocks. This solves #596 and also addresses the problem of #788 in that the fix does not use any external modules, instead relying on setting the TZ environment variable. 

Currently the options for customization are such that users can set the setting **time-TZ** and **time-2-TZ** on their polybar config file to set the timezones of two separate clocks. The options date-2 and time-2 can be added to the label to give the ability to display a dual clock. date-2 is set using the time-2-TZ variable and is not independent of time-2. 

Here's an example using my polybar config file:
```
type = internal/date
interval = 1

date =  %a, %b %d
date-alt =  %Y-%m-%d 

time = "%H:%M"
time-TZ = "Asia/Singapore"
time-2-TZ = "America/Detroit"
time-alt = "%H:%M:%S"

label = %date%   %time% | %date-2%   %time-2%
```
My config

Minor bugs: currently, setting the option time-2-TZ but _not_ time-TZ results in both clocks having the secondary time zone. I don't know how to fix this bug at the moment.

Improvements: I intend to work on a version of this module that allows users to have different timezone when they click on the module, instead of just changing the format.

Fixes #596 
Fixes #1293

### DIFF
--- a/src/modules/date.cpp
+++ b/src/modules/date.cpp
@@ -17,7 +17,6 @@ namespace modules {
     m_dateformat_alt = string_util::trim(m_conf.get(name(), "date-alt", ""s), '"');
     m_timeformat = string_util::trim(m_conf.get(name(), "time", ""s), '"');
     m_time_timezone = string_util::trim(m_conf.get(name(), "time-TZ", ""s), '"');
-    m_time_timezone_2 = string_util::trim(m_conf.get(name(), "time-2-TZ", ""s), '"');
     m_timeformat_alt = string_util::trim(m_conf.get(name(), "time-alt", ""s), '"');
 
     if (m_dateformat.empty() && m_timeformat.empty()) {
@@ -47,7 +46,6 @@ namespace modules {
     const char* timezone_format_original = getenv("TZ");
 
     string timezone_format = m_time_timezone;
-    string timezone_format_2 = m_time_timezone_2;
     if (timezone_format != ""){ 
       setenv("TZ", timezone_format.c_str(), 1);
     }
@@ -64,38 +62,17 @@ namespace modules {
     datetime_stream << std::put_time(localtime(&time), time_format.c_str());
     auto time_string = datetime_stream.str();
 
-    //Secondary time zone
-    string date_string_2, time_string_2;
-    if(timezone_format_2 != ""){
-      setenv("TZ", timezone_format_2.c_str(), 1);
-      auto date_format = m_toggled ? m_dateformat_alt : m_dateformat;
-      // Clear stream contents
-      datetime_stream.str("");
-      datetime_stream << std::put_time(localtime(&time), date_format.c_str());
-      date_string_2 = datetime_stream.str();
-
-      auto time_format = m_toggled ? m_timeformat_alt : m_timeformat;
-      // Clear stream contents
-      datetime_stream.str("");
-      datetime_stream << std::put_time(localtime(&time), time_format.c_str());
-      time_string_2 = datetime_stream.str();
-    }
-
     if (m_date == date_string && m_time == time_string) {
       return false;
     }
 
     m_date = date_string;
-    m_date_2 = date_string_2;
     m_time = time_string;
-    m_time_2 = time_string_2;
 
     if (m_label) {
       m_label->reset_tokens();
       m_label->replace_token("%date%", m_date);
-      m_label->replace_token("%date-2%", m_date_2);
       m_label->replace_token("%time%", m_time);
-      m_label->replace_token("%time-2%", m_time_2);
     }
 
     //Reset TZ so other modules work fine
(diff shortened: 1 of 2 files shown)


patrick96:
What is the benefit of dual clocks in the same module over having two separate date modules?

thomastay:
@patrick96 I suppose you're right. I originally designed the dual clock system with the intention of having a click-to-change timezone module, that's why I implemented the option for dual clock. However that proved difficult without completely rewriting the module so that's why I added the option to display two clocks at once.

thomastay:
hmm I tried having two modules with one timezone each, but I ran into a problem that having two date modules means that the click handler only applies to first of the modules defined. This means that on clicking either of the modules, only the first module defined changes to 

Is this a known bug? I've looked through the code and I'm not sure why the

patrick96:
> Is this a known bug?

Yes, this is the same issue as #1172. If you have multiple modules of the same kind that have internal click events, they are all caught and processed by the first module and don't reach the others.

thomastay:
Ok! Have done so.

thomastay:
@patrick96 are there any more changes that need to be made to the PR before it can be accepted?

patrick96:
Probably. I haven't had the time to properly review this.

src/modules/date.cpp
```diff
@@ -41,6 +42,14 @@ namespace modules {
   bool date_module::update() {
     auto time = std::time(nullptr);
 
+    //Timezone format business
+    const char* timezone_format_original = getenv("TZ");
+
+    string timezone_format = m_time_timezone;
+    if (timezone_format != ""){
```
patrick96: Just a style thing: Please use a space after the if parenthesis

NBonaparte:
We might need to use a mutex to prevent race conditions between multiple modules.

thomastay:
@patrick96 ok i will look into it. Using a mutex like @NBonaparte mentioned is probably the best solution.

thomastay:
@patrick96 I've thought about this and here's a solution I've come up with to prevent the race condition (it is indeed a race condition, I checked). Wanted to hear your input on this as I'm not an expert on multithreading.

Have a central mutex held by the bar's _controller_. When a date module wants to access / set an environment variable, it needs to use the mutex held by the bar's controller. This ensures that only one module is accessing the "TZ" environment variable at once. 

Right now the modules cannot directly call this central mutex since they don't inherit a pointer to the controller. I was thinking of modifying the start function to include a pointer to the controller that started it, but that might be unwieldy. Any ideas / opinions?

patrick96:
Yeah right now we don't really have central place to store such things where everyone has access to. I think storing the mutex in `bar_settings` in `types.hpp` for now should suffice, every module already has access to that. At some point we should probably rethink how things are coupled in general and maybe use a service locator for such things.

patrick96:
@thomastay are you still working on this?

thomastay:
sorry, I haven't had time to work on it. Not working on it for now.

patrick96:
No worries. Maybe someone else will pick this up in the future :)

PR #1371, opened 2018-08-09
fix(net): Allow all characters for the SSID

### LINKED ISSUE #1367: Wireless Module fails when the SSID is an emoji
I have my phone's hotspot SSID set to "😶". This displays fine in my terminal and other wifi indicators.
On Polybar's `wireless-network` module, when connected to this network, it shows the `label-disconnected` label, despite the internet being connected. This does not happen when I change the SSID of my phone's hotspot to something other than "😶".

Yes, I realise how niche of an issue this is. Is this a Polybar problem or something else? I believe that I updated Polybar quite recently, and that may have been the cause of this problem.

taschenb:
Don't filter out any character that is not part of the current C locale, since it might drop some wanted characters.

Fixes #1367

### DIFF
--- a/src/adapters/net_nl.cpp
+++ b/src/adapters/net_nl.cpp
@@ -182,9 +182,7 @@ namespace net {
         auto essid_begin = ies + hdr_len;
         auto essid_end = essid_begin + ies[1];
 
-        // Only use printable characters of the current locale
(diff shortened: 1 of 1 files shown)


lyneca:
Thanks!

PR #1382, opened 2018-08-14
replaces if with contidional statement

ghost:


src/drawtypes/ramp.cpp
```diff
@@ -34,11 +34,7 @@ namespace drawtypes {
     vector<icon_t> vec;
     vector<string> icons;
 
-    if (required) {
-      icons = conf.get_list<string>(section, name);
-    } else {
-      icons = conf.get_list<string>(section, name, {});
-    }
+    icons = required ? conf.get_list<string>(section, name) : conf.get_list<string>(section, name, {});
```
patrick96: I think this hurts legibility too much, please revert.

BO41: @patrick96 sure, will do both :)

Issue #591, opened 2017-05-20
System tray sometimes overlapping the rest of the bar  


seysn:
I've seen some weird issues with the systray, since it's overlapping with the rest of the bar on some conditions. 
First it started to overlap when my download module had less than 3 digits, so i replaced the line
`label-connected = %downspeed%` to `label-connected = %downspeed:8%` which add space so it's like the module always have the same width even with 1 digit.

But then it started to overlap depending the number of icon the systray have. Here's some examples with 2 and 3 icons :
![polybar-nok](https://cloud.githubusercontent.com/assets/6440133/26275181/5ad7658c-3d5b-11e7-8743-b89d29e48fd6.png)
![polybar-ok](https://cloud.githubusercontent.com/assets/6440133/26275225/3dcf5782-3d5c-11e7-92d0-90bc5197d58e.png)

I saw some similar issues about systray overlap like #425 but i don't know if the problem come from the same thing.
Also here's my [polybar config](https://gist.github.com/seysn/fa6d1b7478a48cf317e40b784f3f0d62)

Plasticoo:
This sometimes happens to me too! When I get 3 icons, if the first one disappears (because I closed the program), then it will be shown a blank space, overlapping the rest of the bar.

5haffl:
This issue affects me too. It is not easy to track down the cause, since it happens "randomly".
![overlap](https://user-images.githubusercontent.com/6308274/29063340-a37d64e0-7c25-11e7-84ed-45cc4e5dfb75.png)

patrick96:
I started noticing the same thing too. For me it only seems to happen with the dropbox icon.
![2017-09-07-003457_92x29_scrot](https://user-images.githubusercontent.com/2452038/30137615-6f972264-9364-11e7-8587-bda9759446d1.png)
(Note: The overlap is only very small)

SeerUK:
Not sure if this is related, but if I close something, the space isn't reclaimed - as if the tray area doesn't resize completely properly when something disappears. Right now I end up reloading polybar:

![](https://i.imgur.com/2TZfQ0o.png)
![](https://i.imgur.com/jVczhHg.png)

My power icon ends up obscured.

surrsurus:
Happens to me too, tray on right side overlaps the bar... does it happen when the tray is on the left?
![bar wyd](https://user-images.githubusercontent.com/9388076/30838649-cff7bd00-a23a-11e7-8712-2a577b2f71a2.png)

[...]

PR #1392, opened 2018-08-21
fix(tray): Reconfigure tray on error

patrick96:
See #591. This is the patch jaagr worked on. Some people report this is working, or at least making things better.


patrick96:
I don't think it will hurt anything.

PR #1395, opened 2018-08-23
fix(compiling): `map` variable without namespace.

ysiraichi:
It seems that if you don't have `xcb-xkb` installed, you get a compilation error because `cursor.hpp` uses `map` without `std::` in front of it. When you install `xcb-xkb` back, it includes `x11/extensions/xkb.hpp` which has `using std::map`.

I can reproduce this error on branch 3.2, too.

### DIFF
--- a/include/x11/cursor.hpp
+++ b/include/x11/cursor.hpp
@@ -15,7 +15,7 @@
 POLYBAR_NS
 
 namespace cursor_util {
-  static const map<string, vector<string>> cursors = {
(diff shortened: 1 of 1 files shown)


ysiraichi:
@patrick96 Should I send a patch to branch `3.2`, too? It has the same issue.

patrick96:
No, we'll add this commit to the next upcomming patch release ourselves.

Issue #1393, opened 2018-08-21
Failing Travis Test Build


patrick96:
Currently the travis test build is failing because of a change in googletest, it fails with the following errors (see https://travis-ci.org/jaagr/polybar/jobs/418846511):
```
CMake Error: install(EXPORT "gmockConfigInternal" ...) includes target "gmock" which requires target "xpp" that is not in the export set.
CMake Error: install(EXPORT "gmockConfigInternal" ...) includes target "gmock" which requires target "i3ipc++" that is not in the export set.
CMake Error: install(EXPORT "gmock_mainConfigInternal" ...) includes target "gmock_main" which requires target "xpp" that is not in the export set.
CMake Error: install(EXPORT "gmock_mainConfigInternal" ...) includes target "gmock_main" which requires target "i3ipc++" that is not in the export set.
CMake Error: install(EXPORT "gtestConfigInternal" ...) includes target "gtest" which requires target "xpp" that is not in the export set.
CMake Error: install(EXPORT "gtestConfigInternal" ...) includes target "gtest" which requires target "i3ipc++" that is not in the export set.
CMake Error: install(EXPORT "gtest_mainConfigInternal" ...) includes target "gtest_main" which requires target "xpp" that is not in the export set.
CMake Error: install(EXPORT "gtest_mainConfigInternal" ...) includes target "gtest_main" which requires target "i3ipc++" that is not in the export set.
```
From looking at the recent changes, I think https://github.com/google/googletest/pull/1682 is responsible. I will look into fixing this once I'm done with my exams.

patrick96:
Alright, I was able to confirm that it was definitely https://github.com/google/googletest/commit/880896c6f4814f4c7798355a652dc6167be2b75f, the commit from the PR that broke stuff

PR #1400, opened 2018-08-26
readme: Add release badge

patrick96:


patrick96:
Yeah googletest probably pushed some code that broke our setup, I was going to take a look at it next week (see #1393)

Issue #1406, opened 2018-08-30
Softfail on optional dependencies


FFY00:
## Is your feature request related to a problem? Please describe.
Yes, when polybar is built with an optional dependency, that optional dependency becomes a hard dependency. This is really bad for anyone trying to distribute binaries (ex. distros).

## Why does polybar need this feature?
So that polybar could be available distribution repos.

## Describe the solution you'd like
Instead of trying to blindly launch module, check if the dependencies are present first.

patrick96:
Our current build setup links the shared libraries for the optional dependencies at compile time which makes the whole executable not start, if the libraries it was linked against are missing. The problem is not that we *blindly launch a module*, if the dependency is not present, the while `polybar` executable won't start.

There is a way to load shared libraries during runtime but it would be a major PITA to manually load every library function before we call it.

We had people having issues, because they removed an optional dependency after compilation, before so I'd also like to do this. However, I don't think there is a way to do this without either manually loading the library functions or major architectural changes.

@NBonaparte do you know of any other way to do this?

You can still package polybar for a distro, but you'll have to list the dependencies for all enabled features as hard dependencies.

FFY00:
Why not changing to a plugin architecture? An object would be compiled for each module and then the main executable would try loading it. Your module structure seems relatively easy to port to something like that.

patrick96:
That's what I meant when I was talking about major architectural changes. I don't think it would be that easy to port. We would need to create some way for polybar to interface with the modules and a way for modules to use helper functions inside polybar.

This is definitely something I'd like to see in the long run because it would make it easier to create more complex plugins and would also lessen the burden on us to create those plugins.

skystrife:
I don't think it'll actually be that bad, but I could easily be underestimating the difficulty here.

I think the refactoring steps would be:

1. Update the CMake config to compile each module as a separate shared library, each of which links in whatever optional dependencies it requires. Then, make the `polybar` executable link in every (enabled) module's shared library.
2. Update [`make_module()`](https://github.com/jaagr/polybar/blob/master/include/modules/meta/factory.hpp#L55-L112) to use the "self-registering class" pattern (this will be a nice cleanup step anyway) where we instead look up a factory function from a singleton `map<string, std::function<module_interface*(const bar_settings&, string)>` rather than hard-coding the list of available modules. The basic idea here is to have some statically initialized class within each module's shared library whose constructor registers that module by name with the factory.
[...]

PR #1409, opened 2018-09-01
Plugin architecture for modules

skystrife:
This is my initial draft at using a plugin architecture for loading modules with optional shared library dependencies. This is to address issue #1406.

There are a few major things here:

1. I've switched from using `GLOB_RECURSE` in `src/CMakeLists.txt` to instead having a separate `CMakeLists.txt` that explicitly lists sources within each subdirectory that compiles down to a static library for that subdirectory. I can backtrack here if needed, but I think the end result is a bit cleaner and it's apparently against CMake best practices to use `GLOB_RECURSE` for defining the source files for a target anyway, so I think this should be OK.

2. All modules are built as shared library files. All of the modules that don't have external dependencies are currently all built together as one big shared library that's linked in at compile time. The others are all built as their own standalone shared libraries. 

   Because of how the module factory registration function works, this is currently _mandatory_, but I don't think it is that big of a limitation.

3. Currently the list of plugins is hard-coded in `utils/plugin.hpp`; this is non-ideal but probably OK for now. In the future, we might want to explore how to define a plugin path (or list of paths) to search for modules, but I didn't bother here because there's apparently some goofiness with `dlopen()` with paths that have any slashes in them (where it assumes the path to be absolute).

I'm almost certain to have screwed something up somewhere, so I appreciate any feedback on this (and if people could run it through its paces to root out any remaining bugs).


src/modules/xkeyboard.cpp
```diff
@@ -6,12 +6,15 @@
...
 
 #include "modules/meta/base.inl"
+#include "modules/meta/factory.hpp"
 
 POLYBAR_NS
 
 namespace modules {
   template class module<xkeyboard_module>;
 
+  POLYBAR_MODULE(xkeyboard_module, "xkeyboard");
```
patrick96: This should be `internal/xkeyboard`

patrick96:
Just a quick input, haven't reviewed it thoroughly yet, only looked at the basic approach. 
If we are going to separate out the modules as libraries, would it be a good idea to move all the source files belonging to a module into a single directory? 
So for example for alsa the `adapters/alsa/control.cpp` and `adapters/alsa/mixer.cpp` files would move into the `src/modules/alsa` folder.

To the points you've made:

1. I'm not really a fan of building a library for each subdirectory. What would be the advantage over having each subdirectory add all its files onto a list of source files and pass that list up to their parent? But I agree that explicitly listing files could save some trouble.

2. "*Because of how the module factory registration function works, this is currently mandatory, but I don't think it is that big of a limitation.*" What exactly is mandatory? That the modules without external dependencies are built as one library?

3. I agree. But this can probably wait until we officially support third party compiled modules in this plugin architecture.

I won't be able to properly review this for the next two weeks because I'm going on vacation, but I think getting this merged will be a longer process anyways.

patrick96:
Also, for the next time, please create the branch for a PR in your own fork, if you push directly to this repo it spams the RSS feed.

skystrife:
> If we are going to separate out the modules as libraries, would it be a good idea to move all the source files belonging to a module into a single directory?
> So for example for alsa the adapters/alsa/control.cpp and adapters/alsa/mixer.cpp files would move into the src/modules/alsa folder.

Yeah, this is a good idea.

> I'm not really a fan of building a library for each subdirectory. What would be the advantage over having each subdirectory add all its files onto a list of source files and pass that list up to their parent? But I agree that explicitly listing files could save some trouble.

In hindsight, I agree with you and I've switched to using this approach as it's cleaner.

> "Because of how the module factory registration function works, this is currently mandatory, but I don't think it is that big of a limitation." What exactly is mandatory? That the modules without external dependencies are built as one library?

Building the modules as a shared library is currently mandatory. (They don't need to be together in one shared lib necessarily, but they need to be in _a_ shared lib.) Doing so gets CMake to add the library path in the build directory to the rpath so that the plugins can be found when running `polybar` directly from the build folder. If we don't link against any shared library, the rpath is not modified and it won't search the `build/lib` dir when we try to `dlopen()` the plugins. There might be another way to get around this, but the easiest thing for now is to just build them as a `.so` and let CMake take care of the rest.

> I won't be able to properly review this for the next two weeks because I'm going on vacation, but I think getting this merged will be a longer process anyways.

NBD. This is a big change so we should make sure we do it right.

WithoutCaps:
Any update on this?

Issue #1143, opened 2018-04-05
Repaint issues in the tray area


ntrp:
I am running i3 and the Arc icon theme. I am using compton but the issue appears also without it.

The icons are fine on first paint (after started) and if they change because I muted the audio o disconnected from network they are completely broken:

![2018-04-05-194805_151x43_scrot](https://user-images.githubusercontent.com/720492/38382677-de2a3fb2-390a-11e8-84ce-cd521dcf39be.png)
![2018-04-05-194902_153x43_scrot](https://user-images.githubusercontent.com/720492/38382679-e0faf010-390a-11e8-8f6c-2b3e2f9071a6.png)

Polybar version
```
polybar 3.1.0-78-g224ffa9

Features: +alsa +curl +i3 -mpd +network +pulseaudio

X extensions: +randr (+monitors) -render -damage -sync -composite +xkb +xrm +xcursor

Build type: RelWithDebInfo
Compiler: /usr/bin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -Wall -Wextra -Werror -Wno-noexcept-type -O2 -pedantic -pedantic-errors
Linker flags: -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now
```
Running Xorg and intel drivers.

patrick96:
The tray is quite broken on some systems. We will have a rewrite of the tray at some point

ntrp:
I see, too bad because it was working one week ago :(

patrick96:
That's strange, we didn't change any tray code in a long time.

Any help is appreciated :)

NBonaparte:
Might be caused by updating one or more libraries if it indeed worked before. If that is the case, then we can't do anything about it except roll back that library.

bananflugan:
I have the same issue, the icons can randomly get broken. Another tray issue is that randomly when i start my computer or do changes to the configuration file some tray icons do not show up at all, its like the tray moves into a weird position which also causes my date module to not be fully visible.

Polybar 3.2.1

leiserfg:
The same bug happens in i3bar and there you can "fix it" by using the transparent flag, so it is something fixable but I think related to a library update 'cause doesn't occurred before.

patrick96:
Hmm, the latest development version cleaned up tray transparency a bit in #595, maybe this will help resolve this.

JonnyHaystack:
[...]

PR #1415, opened 2018-09-06
Fixed tray repain issue #1143

moonlightnvkz:
The error occured due to the tray's background not being repainted with `tray-transparent = false`


NBonaparte:
Do you have a good way to verify this through testing? I haven't found a way to reliably reproduce this.

patrick96:
Closing due to inactivity

Issue #1054, opened 2018-02-26
Debian packaging: i3ipcpp and xpp


jpleau:
Hi.

I'm starting to pickup my work at packaging polybar for Debian/Ubuntu. The one issue I have at the moment is the embedded libraries in the lib/ directory. I would like to package them separately (i3ipcpp and xpp). I saw that both are forked, and have more code than the original repositories (drmgc/i3ipcpp and jotrk/xpp)

I have a few options:

A) Use your forks (and point to your repos as upstream)
B) Use the original repos, but we wouldn't have the changes needed to make polybar work at 100% likely
C) Use the original repos and carry the diff from your repos as patches (I don't like this suggestion at all)
D) Ask here if it's possible to merge every back to the original repos (I highly doubt this can happen)

What of the above is the one I should pick? I'm open to other suggestions as well.

NBonaparte:
A is the best option. I don't think B and D would work either, and C just looks like a mess.

Is what you said in #447 still relevant?

jpleau:
Hi.

Yes it's still relevant, but once I package i3ipcpp and xpp (using your forked repos), all should work fine.

I might need a bit of help to make sure I use the installed libraries on the system (instead of the ones shipped in the tarball), but I'll ask when the time comes :)

patrick96:
Hmm, the jsoncpp tarball in the repo is totally useless, I've opened a PR to remove it.

Why do you want to package xpp and i3ipcpp separately? I understand that you can't download anything during the build process, but couldn't you provide the whole polybar repo (including the submodules) to the build servers? I don't know anything about the build process for official debian (or ubuntu for that matter) packages, so what I proposed could not work at all. 
Can you give me a quick run through? Are there any constraints on what type of files you can put on the build server or where they come from?

I agree with @NBonaparte that A would work best, if you want to package them separately.
Option D is unfeasible because the turnaround rate for PRs would be way to low. The original xpp is unmaintained and i3ipcpp has barely seen any activity in the last year. Option B probably also won't work, as the original xpp uses Makefiles instead of cmake. As for using upstream i3ipcpp, this one might actually work, however I see that debian has jsoncpp version 1.7.4, in our i3ipcpp fork, we build jsoncpp 1.7.7, if a more recent version is not available, so it is possible that there is some functionality that i3ipcpp uses but that isn't in 1.7.4 (I wasn't able to determine what that would be, but it is possible that jaagr discovered that there is).

jpleau:
[...]

PR #1421, opened 2018-09-15
readme: Remove GetDeb mention

patrick96:
GetDeb.net has been offline for a month and isn't looking like coming
back up. Now there isn't even a DNS record anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Here are a few screenshots showing you what it can look like:
 [![sample screenshot](http://i.imgur.com/A6spiZZt.png)](http://i.imgur.com/A6spiZZ.png)
 [![sample screenshot](http://i.imgur.com/TY5a5r9t.png)](http://i.imgur.com/TY5a5r9.png)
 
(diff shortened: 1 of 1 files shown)


kronn:
I support this step. There is no use in mentioning this. Even before, it was outdated. Are there any alternatives for deb-packages on the horizon?

patrick96:
In #1054 there was someone trying to package for the debian official repo but I haven't heard anything back yet. Packaging precompiled binaries is an issue anyway right now because if polybar is compiled with let's say i3 support you need to have i3 installed to even run that polybar executable (see #1406).

Until this is resolved it will be quite difficult for someone to actually properly package polybar in any official repository. Had someone on IRC the last few days that ran into the same roadblocks.

kronn:
@patrick96 Ah, thanks for clarifying. I was hoping it was "just" a matter of setting up a build-pipeline or even a PPA that could be filled by `$CI_SAAS`.

patrick96:
I don't know how feasible it would be to set that up but if it's possible we would certainly be open to that once we finished the plugin architecture.

PR #1422, opened 2018-09-15
fix(cpu): Replace %percentage-cores% normally

### LINKED ISSUE #1282: %percentage-cores does not support min-width.
Hey!

I cannot seem to get the min-width syntax working with the %percentage-cores% token.

```ini
[module/cpu]
type = internal/cpu

label = %percentage-cores:4%
```
Just results in `%percentage-cores%` in the bar.

TonCherAmi:
Resolves #1282 

This pretty much restores the way things were done pre 3292cea where the token splitting shenanigans were introduced.

### DIFF
--- a/src/modules/cpu.cpp
+++ b/src/modules/cpu.cpp
@@ -34,16 +34,6 @@ namespace modules {
       m_rampload_core = load_ramp(m_conf, name(), TAG_RAMP_LOAD_PER_CORE);
     }
     if (m_formatter->has(TAG_LABEL)) {
-      // Update the label parameter and replace the %percentag-cores% token with the individual core tokens
-      string key{&TAG_LABEL[1], strlen(TAG_LABEL) - 2};
-      auto label = m_conf.get<string>(name(), key, "%percentage%%");
-      vector<string> cores;
-      for (size_t i = 1; i <= m_cputimes.size(); i++) {
-        cores.emplace_back("%percentage-core" + to_string(i) + "%%");
-      }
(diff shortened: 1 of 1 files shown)


patrick96:
Shouldn't the min-width property be applied to each core percentage individually instead of to the whole token? When it was a compound token that interpretation made the most sense to me, that's why I initially thought this PR didn't really fix the issue.

TonCherAmi:
I don't believe it should. In my opinion that goes against the way tokens are normally expected to work, the implementation is also rather hackish as a result. Furthermore it would make it impossible to set max-width of the token, whereas the behavior you describe is still achievable using individual `%percentage-corex%` tokens.

TonCherAmi:
I do agree however that the behavior you describe would be pretty nice to have in this case. Perhaps we could introduce some kind of new syntax for this sort of thing, e.g. `[%percentage-cores%]` or something like that. We could then deprecate `%percentage-cores%` afterwards.

patrick96:
@NBonaparte what do you think?
I think `%percentage-cores%` should be regarded as a compound token and pass on all its token properties to the `%percentage-coreX%` tokens.

NBonaparte:
I suppose that is more logical. @TonCherAmi's suggestion also makes sense since right now compound tokens don't have different syntax, so it could be confusing.

patrick96:
`%percentage-cores%` is the only compound token in polybar and I can't think of any other situation where a compound token would make sense. Introducing new syntax for just this token seems a bit much to me. As long as it is documented properly this shouldn't create any problems.

PR #1424, opened 2018-09-16
doc: List xcb-xkb as optional dependency

patrick96:


NBonaparte:
Should we also add all xcb extensions to the README as well, since they're bundled separately in some distros?

patrick96:
I don't think we should. Have you found distros other than debian based ones that package the xcb libraries for X extensions separately? Because for debian based systems we already list the required packages on the wiki.

PR #1426, opened 2018-09-18
feat(monitor): use primary monitor as default

### LINKED ISSUE #1412: Support for randr primary output
I use my laptop with different screens which use different outputs. The monitor and monitor-fallback options only let me choose two levels of fallback (e.g. DP1 and LVDS0), whereas practically I would like DP1 or HDMI1, then LVDS0 in the worst case.

I don't always have the main screen on the same side (depends on the desk).
Xrandr has the notion of a primary display, which is supposed to communicate which one is the most important.

It will make it easier to have a polybar that always appears on the main display, regardless of which socket the display is plugged into.

For polybar to support something like `monitor = PRIMARY` or `monitor-primary = true` in monitor selection. I did look at writing a patch but my x programming skills aren't good enough.

Ehh, at the moment I can grep this out of the randr config and then patch the config file and restart polybar on randr events. This obviously disappoints.

NBonaparte:
Fixes #1412

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -78,7 +78,7 @@ int main(int argc, char** argv) {
     if (cli->has("list-monitors")) {
       for (auto&& mon : randr_util::get_monitors(conn, conn.root(), true)) {
         if (WITH_XRANDR_MONITORS && mon->output == XCB_NONE) {
-          printf("%s: %ix%i+%i+%i (XRandR monitor)%s\n", mon->name.c_str(), mon->w, mon->h, mon->x, mon->y,
+          printf("%s: %ix%i+%i+%i (XRandR monitor%s)\n", mon->name.c_str(), mon->w, mon->h, mon->x, mon->y,
               mon->primary ? " (primary)" : "");
         } else {
           printf("%s: %ix%i+%i+%i%s\n", mon->name.c_str(), mon->w, mon->h, mon->x, mon->y, mon->primary ? " (primary)" : "");
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -78,9 +78,10 @@ int main(int argc, char** argv) {
     if (cli->has("list-monitors")) {
       for (auto&& mon : randr_util::get_monitors(conn, conn.root(), true)) {
         if (WITH_XRANDR_MONITORS && mon->output == XCB_NONE) {
-          printf("%s: %ix%i+%i+%i (XRandR monitor)\n", mon->name.c_str(), mon->w, mon->h, mon->x, mon->y);
+          printf("%s: %ix%i+%i+%i (XRandR monitor)%s\n", mon->name.c_str(), mon->w, mon->h, mon->x, mon->y,
+              mon->primary ? " (primary)" : "");
         } else {
-          printf("%s: %ix%i+%i+%i\n", mon->name.c_str(), mon->w, mon->h, mon->x, mon->y);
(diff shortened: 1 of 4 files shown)


NBonaparte:
@patrick96 bump

patrick96:
Sorry about that. I have put it on my calendar next week.

src/x11/extensions/randr.cpp
```diff
@@ -90,13 +92,18 @@ namespace randr_util {
       for (auto&& mon : conn.get_monitors(root, true).monitors()) {
         try {
           auto name = conn.get_atom_name(mon.name).name();
-          monitors.emplace_back(make_monitor(XCB_NONE, move(name), mon.width, mon.height, mon.x, mon.y));
+          auto primary = (mon.primary != 0);
```
patrick96: Why not use mon.primary directly?

NBonaparte:
Yeah, the setting overrides everything else.

patrick96:
Then there is probably some issue with it, on my machine, setting `monitor` will display the bar on that monitor, regardless of `monitor-primary`. Can you reproduce that?

It's probably because you don't break out of the while loop, once you found the primary monitor.
What would probably work better is to first search for the primary monitor and just leave the whole logic for finding monitors afterwards unchanged. And if there is no primary monitor for some reason, fall back to the old logic for selecting the monitor.

Also, could you make `polybar -m` mark the primary monitor in some way?

frebib:
Would the syntax `monitor = primary` not make more sense, instead of adding more keys with confusing names?

patrick96:
@frebib I mentioned this in the feature request [issue](https://github.com/jaagr/polybar/issues/1412#issuecomment-420335964) for this PR:
> I'm not a fan of using magic strings.

I think it's bad style that some values for `monitor` would behave differently while all other values specify a monitor name. Also, it's possible to rename your monitor names to arbitrary strings, so you could get conflicts there.

But to make the whole thing cleaner, we could use the absence of the `monitor` key (or an empty string) to mean that the primary monitor should be used. This makes things less explicit but you need one less key and if you would specify `monitor-primary` you wouldn't need the other `monitor-*` keys anyway.

src/main.cpp
```diff
@@ -78,9 +78,10 @@ int main(int argc, char** argv) {
     if (cli->has("list-monitors")) {
       for (auto&& mon : randr_util::get_monitors(conn, conn.root(), true)) {
         if (WITH_XRANDR_MONITORS && mon->output == XCB_NONE) {
-          printf("%s: %ix%i+%i+%i (XRandR monitor)\n", mon->name.c_str(), mon->w, mon->h, mon->x, mon->y);
+          printf("%s: %ix%i+%i+%i (XRandR monitor)%s\n", mon->name.c_str(), mon->w, mon->h, mon->x, mon->y,
```
patrick96: I think it would look better if primary monitors here were marked with `(XRandR monitor, primary)` instead of `(XRandR monitor) (primary)`

```suggestion
          printf("%s: %ix%i+%i+%i (XRandR monitor%s)\n", mon->name.c_str(), mon->w, mon->h, mon->x, mon->y,
```

src/main.cpp
```diff
@@ -78,9 +78,10 @@ int main(int argc, char** argv) {
...
+              mon->primary ? " (primary)" : "");
```
patrick96:
```suggestion
              mon->primary ? ", primary" : "");
```

PR #1438, opened 2018-09-22
Battery charging state from battery instead of AC

### LINKED ISSUE #1374: [Battery] Get charging status from battery instead of AC
Currently, polybar reports battery status by checking if AC is online ([1](https://github.com/jaagr/polybar/blob/master/src/modules/battery.cpp#L37) & [2](https://github.com/jaagr/polybar/blob/master/src/modules/battery.cpp#L285)). Instead of this, won't it be better to use `/sys/class/power_supply/<battery>/status` which is currently used as fallback?

This will remove the use of `full-at` option, as status automatically changes to 'Discharging' once the battery is fully charged but not 100% (Atleast on mine it does).

PenguinSnail:
Closes #1374 

I've tested this on my own personal laptop and everything appears to still work as far as I can tell. My laptop does also have two batteries hence the need for the `Unknown` state.

 The two batteries were my reason for attempting to fix this, because only one discharges at a time but when the state is pulled from AC, polybar show's them both as discharging at the same time which is not the case.

### Changes
- changed type of `state_reader` from bool to int (`include/modules/battery.hpp` line 42)
- pull state directly from battery and move AC to fallback (this used to be the other way around) (`src/modules/battery.cpp` lines 36-52)
- add some things to state detection to give different states (`src/modules/battery.cpp` lines 40-46)
    - if the battery is `Charging`, state = 0
    - if the battery is `Discharging`, state = 1
    - if the battery has any other state (ex. `Unknown` if a battery is neither charging nor discharging on a laptop with two batteries) then state = 2
    **Note:** These states have the same value as before, 0 for charging - so anything checking the value should still function properly as we have only added states, not changed them
- update the modules state according to this "new" system (`src/modules/battery.cpp` lines 290-298)


NBonaparte:
Just FYI, there are other PRs (#725, #1286) which also implement the Unknown status, although it has gone stale.

patrick96:
#2556 has now basically implemented the same thing.

PR #1442, opened 2018-09-26
Support double-clicks in custom/script.

dkasak:
This adds support for double-click handlers in the custom/script module
(for all buttons; left, right and middle).


patrick96:
Weird that this wasn't a feature before. Will try to take a look it in the next few days

patrick96:
Does this work on your machine? I can't get it to work on mine.

dkasak:
It did work when I wrote it, yes. I'll have to re-check whether something broke in the meantime.

dkasak:
I rebased this on top of the current master and recompiled. It does work for me.

patrick96:
The polybar double click handler doesn't seem to work well in general, so we need to do some fixing there. But the PR generates the correct action tags, so nothing wrong there.

Issue #1447, opened 2018-09-30
Build Issue - can't build with networking support on Alpine


james9837:
## Build Process
Following the [instructions](https://github.com/jaagr/polybar/wiki/Compiling) as precisely as possible.

*Describe how you build polybar, list the exact commands you are using:*
I used the exact commands from that document:
```
git clone --recursive https://github.com/jaagr/polybar
mkdir polybar/build
cd polybar/build
cmake ..
sudo make install
```
**If you build polybar directly from this repository:**

* Output of `git describe --tags`:
3.2.1

**If you use some other way (like the AUR):**
nope
*List exactly where you are building from and which version you are building.*

## Build Log
*Post everything that is output to the terminal while building polybar below. This HAS to include the output of the `cmake` and `make` commands, if you used them.*
```
cmake output...
-- Trying to enable ccache
-- Using compiler cache /usr/bin/ccache
-- The C compiler identification is GNU 6.4.0
-- The CXX compiler identification is GNU 6.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- No build type specified; using RelWithDebInfo
-- Using supported compiler GNU-6.4.0
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.5.3") 
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - found
-- Found Threads: TRUE  
-- Checking for module 'cairo-fc'
--   Found cairo-fc, version 1.15.12
-- Checking for module 'alsa'
--   Found alsa, version 1.1.6
-- Checking for module 'libnl-genl-3.0'
--   Found libnl-genl-3.0, version 3.4.0
-- Checking for module 'xcb-xkb'
--   Found xcb-xkb, version 1.13
-- Checking for module 'xcb-randr'
--   Found xcb-randr, version 1.13
-- Checking for module 'xcb-randr>=1.12'
--   Found xcb-randr, version 1.13
--  Build:
[...]

PR #1452, opened 2018-10-02
replace caddr_t with char*

james9837:
to fix [issue 1447](https://github.com/jaagr/polybar/issues/1447)

Not sure what else is needed for this PR (my first on public github :P)

With this fix, it builds and runs for me - works great.


james9837:
doh! I made the change in my fork's 'master' branch... killing this PR since I can't redirect it to source from a proper branch.

PR #1456, opened 2018-10-04
fix(script): Avoid duplicate action tags for tailed

patrick96:
If a tailed command is used polybar would generate two action tags, one
with %counter% replaced and one with %pid% replaced, but never both

This is a bug that was introduced in #934


patrick96:
Duh :man_facepalming: 

Done

PR #1457, opened 2018-10-04
fix(cmd): Escape all colons in action tag command

### LINKED ISSUE #984: formatting syntax in i3 workspace names should be ignored
My i3wm + polybar setup uses `strip-wsnumbers = true` to set a sorting order for my static workspaces (e.g. browser, music,...).

Since I still want to see numbers for "dynamic" workspaces (mapped to mod+0..9), these end up being named e.g. `1:1:stuff` (shown as `1:stuff` in the bar).

This worked well with i3bar but in polybar this seems to be interpreted as some kind of format syntax? E.g. `1:1:rust` will make the whole i3 module right-aligned, `1:1:R...` will result in inverted foreground and background. All format codes from the [Formatting wiki page](https://github.com/jaagr/polybar/wiki/Formatting) seem to work.

If the first letter of the actual name is no format code, the whole bar will be empty and I get 
```error: Failed to parse contents (reason: Unrecognized token 'b')```

This does not happen with "normal" workspace names like `1:rust`.

Is this an undocumented format code syntax? Regardless of that it should probably not be interpreted in workspace names?

(the obvious workaround is to have workspaces named like `1:1.rust` instead but still…)

patrick96:
The builder::cmd function creates a new action block with a command.
However it fails to escape all colons and only escapes the first colon

~For now I have refactored the code a bit and added failing unit tests to
show the bug. I will implement the proper behaviour in the next few
days (hopefully).
This changes the semantics a bit: We now also escape the escape character so that we get the exact same command back that we put in.~

Fixes #984

### DIFF
--- a/src/components/builder.cpp
+++ b/src/components/builder.cpp
@@ -499,11 +499,7 @@ void builder::underline_close() {
  */
 void builder::cmd(mousebtn index, string action, bool condition) {
   if (condition && !action.empty()) {
-    size_t p{0};
-    while ((p = action.find(':', p)) != string::npos && action[p - 1] != '\\') {
-      action.insert(p, 1, '\\');
-      p++;
-    }
+    action = string_util::replace_all(action, ":", "\\:");
     tag_open(syntaxtag::A, to_string(static_cast<int>(index)) + ":" + action + ":");
   }
 }
--- a/src/components/parser.cpp
+++ b/src/components/parser.cpp
@@ -64,6 +64,33 @@ void parser::codeblock(string&& data, const bar_settings& bar) {
     }
 
     char tag{data[0]};
+
+    /*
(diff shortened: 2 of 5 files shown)


patrick96:
This broke my brain a little yesterday. The parser actually uses the length of the string returned from `parse_action_cmd` to advance the data string. So we cannot unescape inside `parse_action_cmd`. Also I had some doubts about the escaping semantics so I have left the functionality as it was before with just fixing the bug. Backslashes are not also escaped, only the colons

patrick96:
I don't follow, the escape sequence already is `\`, in the code we have to write `\\` to get a literal backslash. Or do you mean something else?

NBonaparte:
Never mind.

PR #1472, opened 2018-10-15
Add ramp-padding capabilities to module/cpu

### LINKED ISSUE #1389: Way to remove spacing between cores in power module
The power module's `<ramp-coreload>` option is really cool, but it takes up way more space on the bar than is needed to get its point across. For example on a 16-core machine:

```
 CPU: ▂ ▄ ▄ ▂ ▂ ▂ ▇ ▄ ▂ ▂ ▂ ▂ ▄ ▂ ▂ (17%)`
```
it would be so much more space efficient to display without those gaps (and look better in many instances!):
```
 CPU: ▂▄▄▂▂▂▇▄▂▂▂▂▄▂▂ (17%)`
```
`ramp-coreload-margin` works for increasing the gap but not removing it entirely. The only solution I've found so far is to use a pixel offset in the format of each individual core label, e.g:

```
ramp-coreload-0 = "%{O-12}%{F#999}▁"
ramp-coreload-1 = "%{O-12}%{F#977}▂"
ramp-coreload-2 = "%{O-12}%{F#955}▃"
ramp-coreload-3 = "%{O-12}%{F#931}▄"
ramp-coreload-4 = "%{O-12}%{F#900}▅"
ramp-coreload-5 = "%{O-12}%{F#B00}▆"
ramp-coreload-6 = "%{O-12}%{F#D00}▇"
ramp-coreload-7 = "%{O-12}%{F#F00}█"
```

but this pixel-specific approach doesn't work with things like changing fonts and breaks other workarounds like having a common bar format with standard resolution and "hi-DPI" bars inheriting from that.

I'm thinking `<ramp-coreload-padding>` would be the most consistent place to implement this? Or maybe even a more global or bar-wide setting to define the spacing between ramp glyphs in this kind of widget?

[edit: oops, `<tags>` need to be escaped or don't show up]

theryangeary:
Fixes #1389 
Allows "ramp-padding" in polybar config file to adjust spacing between core loads.

### DIFF
--- a/contrib/polybar.aur/PKGBUILD
+++ b/contrib/polybar.aur/PKGBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Michael Carlberg <c@rlberg.se>
 # Contributor: Michael Carlberg <c@rlberg.se>
 pkgname=polybar
-pkgver=3.2.0
+pkgver=3.2.1
 pkgrel=1
 pkgdesc="A fast and easy-to-use status bar"
 arch=("i686" "x86_64")
--- a/include/modules/cpu.hpp
+++ b/include/modules/cpu.hpp
@@ -37,6 +37,7 @@ namespace modules {
     ramp_t m_rampload;
     ramp_t m_rampload_core;
     label_t m_label;
+    int m_ramp_padding;
 
     vector<cpu_time_t> m_cputimes;
     vector<cpu_time_t> m_cputimes_prev;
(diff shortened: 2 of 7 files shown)


patrick96:
Oh damn. Because you originally created a PR against the `3.2` branch, it contained a commit that was exclusive to that branch (3.2.1 release commit) and we don't actually want that in `master`. Will manually add your commits to a new PR.

theryangeary:
So is there anything else I need to do resolve that or is it taken care of?

patrick96:
No, everything is taken care of. I have taken the commits you made and applied them to the master branch and openened a PR for that.

PR #1478, opened 2018-10-16
feat(pulseaudio): add headphones detection

julio-b:
Properly detect if current sink is a headphone and change labels or ramps accordingly.
That's the only feature that pulseaudio module was lacking, and i finally found some free time to implement it. I tested this commit and i think everything works as expected, but please review these changes.
Label/ramp replacement is similar to alsa module. 

```dosini
; Available tags:
;   <label-volume> (default)
;   <ramp-volume>
;   <bar-volume>
format-volume = <ramp-volume> <label-volume>

; Available tags:
;   <label-muted> (default)
;   <ramp-volume>
;   <bar-volume>
;format-muted = <label-muted>

; Available tokens:
;   %percentage% (default)
;label-volume = %percentage%%

; If defined will replace <label-volume> when headphones are plugged in
; Available tokens:
;   %percentage%
;   null (default)
;label-headphones-volume =   %percentage%%

; Available tokens:
;   %percentage% (default)
label-muted = 🔇 muted
label-muted-foreground = #66

; If defined will replace <label-muted> when headphones are plugged in
; Available tokens:
;   %percentage%
;   null (default)
;label-headphones-muted = 🔇

; Only applies if <ramp-volume> is used
ramp-volume-0 = 🔈
ramp-volume-1 = 🔉
ramp-volume-2 = 🔊

; If defined, it will replace <ramp-volume> when headphones are plugged in
; If undefined, <ramp-volume> will be used for both
; Only applies if <ramp-volume> is used
ramp-headphones-0 = 
ramp-headphones-1 = 
```

### DIFF
--- a/include/adapters/pulseaudio.hpp
+++ b/include/adapters/pulseaudio.hpp
@@ -43,6 +43,7 @@ class pulseaudio {
     void set_mute(bool mode);
     void toggle_mute();
     bool is_muted();
+    bool is_headphone();
 
   private:
     void update_volume(pa_operation *o);
@@ -63,6 +64,7 @@ class pulseaudio {
     bool muted{false};
     // default sink name
     static constexpr auto DEFAULT_SINK{"@DEFAULT_SINK@"};
+    string active_port_name;
 
     pa_context* m_context{nullptr};
     pa_threaded_mainloop* m_mainloop{nullptr};
--- a/include/modules/pulseaudio.hpp
+++ b/include/modules/pulseaudio.hpp
@@ -31,9 +31,12 @@ namespace modules {
     static constexpr auto FORMAT_MUTED = "format-muted";
 
     static constexpr auto TAG_RAMP_VOLUME = "<ramp-volume>";
+    static constexpr auto TAG_RAMP_HEADPHONES = "<ramp-headphones>";
     static constexpr auto TAG_BAR_VOLUME = "<bar-volume>";
     static constexpr auto TAG_LABEL_VOLUME = "<label-volume>";
     static constexpr auto TAG_LABEL_MUTED = "<label-muted>";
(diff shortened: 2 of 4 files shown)


patrick96:
Neat! Thank you.

Will do a review once I can dedicate some time.

NBonaparte:
Looks like some functionality is repeated from #1075 which also parses the port name; we'll have to figure out how to handle that before merging.

julio-b:
Oh that's awesome! Your approach is much better than mine. You guys should merge it asap. The only issue is that analog-* are handled incorrectly, right? Well, to solve that for good, allow users to config icons for a specific port. I don't know if that's possible but I'm thinking of something like this:
```dosini
[module/vol]
type = internal/pulseaudio
icon-headphones = "icon-headphones"
icon-speaker = "icon-speaker"
icon-hdmi = "icon-hdmi"
icon-alsa_output.pci-0000_12_00.3.analog-stereo  = "FOO"
icon-alsa_output.pci-0000_12_00.3.analog-out-something  = "FOOOO"
```
and then choose which icon to display
```
if is_defined(m_port_icons->get(sink + "." + port))
    icon = m_port_icons->get(sink + "." + port)
else //try to detect headphones,speaker etc
   if bluetooth  
      icon = sink.icon_name
   else if ...
       ....
       // etc
```

albakham:
It would be great to finally see this commit being merged :)

markovichecha:
I need this PR so much, merge it.

kronn:
After seeing this and #1075, I would say: merge this PR first as it contains fewer questions (none, as far as I can see). Then you could use that as base for #1075, which has more open tasks.

Does that make sense? Can/should anyone help with that?

PR #1501, opened 2018-11-05
fix(man): Add correct date and version to manpage

patrick96:
The previous approach relied on manually updating the version and date
using a Makefile

This approach dynamically gathers the correct version from git and uses the
date of the latest commit in the branch


patrick96:
Fixed merge conflict!

PR #1505, opened 2018-11-10
Allow including config files using relative paths

sam-lunt:
This pull request enables relative include paths (currently, a relative include path is prepended with a separator character). Following the convention of the C preprocessor, relative paths are interpreted from the directory of the including file (this is also how the terminal editor kitty does it).

The reason I wanted to add this was so that I could include files without needing the full path. It makes it easier to move configuration folders to different places or share them across machines with different directory layouts.

While I was making changes, I also added a few changes to the expand logic: it will expand variables anywhere in the path component, not just in the beginning, and it allows the use of braces (e.g. `${HOME}` instead of `$HOME`). I added tests to verify the behavior, but to get the tests to build properly, I had to refactor the CMake logic a bit. (this might be a bigger change than the actual path logic).


tests/unit_tests/main.cpp
```diff
@@ -0,0 +1,10 @@
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+int main(int argc, char** argv) {
```
patrick96: Did you change anything that now requires us to now have a main function for the tests?

lib/CMakeLists.txt
```diff
@@ -31,19 +34,19 @@ if(WITH_XKB)
...
   add_subdirectory(i3ipcpp)
-  list(APPEND libs ${I3IPCPP_LIBRARIES})
-  list(APPEND dirs ${I3IPCPP_INCLUDE_DIRS})
+  list(APPEND libs i3ipc++)
 endif()
 
 # }}}
 
 set(libs ${libs} PARENT_SCOPE)
-set(dirs ${dirs} PARENT_SCOPE)
```
patrick96: Why can we remove the dirs variable?

lib/CMakeLists.txt
```diff
@@ -31,19 +34,19 @@ if(WITH_XKB)
...
+list(APPEND libs xpp-interface)
 
 # }}}
 # Library: i3ipcpp {{{
 
 if(ENABLE_I3)
   add_subdirectory(i3ipcpp)
-  list(APPEND libs ${I3IPCPP_LIBRARIES})
-  list(APPEND dirs ${I3IPCPP_INCLUDE_DIRS})
+  list(APPEND libs i3ipc++)
```
patrick96: This should remain `${I3IPCPP_LIBRARIES}` for modularity.

src/utils/file.cpp
```diff
@@ -233,25 +271,108 @@ namespace file_util {
     return ret;
   }
 
+  /*
+   * Return the path without the last component (see dirname(3))
+   */
+  const string dirname(const string& path) {
```
patrick96: Why don't we use `dirname(3)` for this directly?

lib/CMakeLists.txt
sam-lunt: So CMake actually supports 3 types of values as the right-hand argument of `target_link_libraries(<target> <item>)`:

1. The path to a `.a` or `.so` file
2. A CMake target
3. A linker flag (e.g.`-lm`)

A CMake target is an object with properties associated with it. These properties can be things like include directories, compile-time definitions, etc. When a target is used as the right hand argument of target_link_libraries, it passes these properties on the target on the left-hand side.

Before this change, all of the items in the `lib` list were regular `.a` or `.so` files, which is why the include directories needed to be handled separately. I changed them so that some of the libs are now CMake targets, which can contain their include directories directly.

lib/CMakeLists.txt
sam-lunt: I can change it if you'd like, but I actually think this will be more modular.

This relates to the comment about `dirs`. `i3ipc++` is a CMake target defined [here](https://github.com/jaagr/i3ipcpp/blob/d4e4786be35b48d72dc7e59cf85ec34a90d129b5/CMakeLists.txt#L10) that actually already includes the contents of `JSONCPP_INCLUDEDIR` (defined [here](https://github.com/jaagr/i3ipcpp/blob/d4e4786be35b48d72dc7e59cf85ec34a90d129b5/CMakeLists.txt#L47)) as a property [here](https://github.com/jaagr/i3ipcpp/blob/d4e4786be35b48d72dc7e59cf85ec34a90d129b5/CMakeLists.txt#L18) and [here](https://github.com/jaagr/i3ipcpp/blob/d4e4786be35b48d72dc7e59cf85ec34a90d129b5/CMakeLists.txt#L39).

I'd also guess that the `i3ipc++` target is kept more up-to-date than the `I3IPCPP_LIBRARIES`/`I3IPCPP_INCLUDE_DIRS` variables, since if they add a new library or include directory to their project and forget to update the target, the build will fail, but that won't be true for those variables.

tests/unit_tests/main.cpp
sam-lunt: Ah, my bad, it looks like my local git repo got slightly messed up. I meant to delete that file. I'm also missing a few other changes. Once I sort it out, I'll push a few more commits to fix it.

lib/CMakeLists.txt
patrick96:
```cmake
list(APPEND libs ${I3IPCPP_LIBRARIES})
```
and
```cmake
list(APPEND libs i3ipc++)
```
are completely equivalent.

`${I3IPCPP_LIBRARIES}` just contains `i3ipc++` anyway, see [here](https://github.com/jaagr/i3ipcpp/blob/d4e4786be35b48d72dc7e59cf85ec34a90d129b5/CMakeLists.txt#L46) so using the variable will keep us from duplicating the name (even if it probably never changes).

src/utils/file.cpp
patrick96: I'd rather not reimplement a standard library function in our own codebase, even if using it is not so straightforward.

We could use `file_util::dirname(const string&)` as a wrapper around `dirname` that handles memory allocation, copying of the string, and thread safety.

In pseudocode, I think it would look something like this:
```
copy arg to newly allocated char*
lock
dirname(...)
copy result to std::string
unlock
free allocated char*

File: lib/CMakeLists.txt [hunk 5]
```diff
@@ -31,19 +34,19 @@ if(WITH_XKB)
 endif()
 
 add_subdirectory(xpp)
-list(APPEND libs ${XPP_LIBRARIES})
-list(APPEND dirs ${XPP_INCLUDE_DIRS})
+add_library(xpp-interface INTERFACE)
```
patrick96: `xpp` already calls `add_library` itself [here](https://github.com/jaagr/xpp/blob/master/CMakeLists.txt#L173).
We could only do `list(APPEND libs xpp)`

tests/CMakeLists.txt
```diff
@@ -38,67 +38,30 @@ add_subdirectory(${CMAKE_BINARY_DIR}/googletest-src
...
-    if(NOT "${f}" STREQUAL "main.cpp")
-      LIST(APPEND sources "../src/${f}")
-    endif()
-  ENDFOREACH(f)
+# Compile all unit tests with 'make all_unit_tests'
+add_custom_target(all_unit_tests
+    COMMENT "Building all unit test")
 
-  string(REPLACE "/" "_" testname ${file})
+function(add_unit_test source_file)
```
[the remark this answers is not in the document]
sam-lunt: Do you mean the name change or the actual refactor of the CMake function?

If you meant the change in logic: I tried to add a test for config.cpp, but I was getting a lot of link errors due to missing symbols. Some of them were from functions defined in `src`, but others were from symbols defined by entries in the `libs` variable. Instead of selectively recompiling some source files in order to build the unit tests, I split the polybar executable into a library and an executable [here](https://github.com/sam-lunt/polybar/commit/e2605aea2d4607a61d7590e160dcd88fd6019c9b#diff-95e351a3805a1dafa85bf20b81d086e6R72), which allows the tests to link against `polybar-lib` (if you have a better name, I'm not the biggest fan of that), which includes the dependencies as well as the symbols in `src`. In addition to making the unit test logic simpler, it should improve build times (not that that's a big concern).

If you meant the name, I changed it because I thought `unit_test` was a little unclear, and `add_*` fits with the other CMake functions `add_executable`, `add_library`, `add_test`, etc.

lib/CMakeLists.txt
sam-lunt: Good point. I think I was tripped up by the fact that they link privately against `${XPP_LIBRARIES}`, but on a second look, I think that should be fine.

src/utils/file.cpp
sam-lunt: I understand the desire to avoid reinventing the wheel, but the current implementation is only 11 lines of code, and the logic to use the real `dirname` is almost as long and, in my opinion, more complicated (in addition to being slower). Also, even with the lock, it won't be thread-safe if `dirname(3)` is used elsewhere in the executable (for example, by a statically linked library).

Here's a version that uses dirname directly. If you prefer it, I'll put it in.
``` cpp
const string dirname(string path) {
  static std::mutex m;

  // dirname may return a pointer to static memory,
  // use a lock to avoid race condition due to concurrent calls
  std::unique_lock<std::mutex> lock{m};
  auto ptr = dirname(&path[0]);

  if (ptr == path.c_str()) // modified in place
  {
    // dirname does not use static memory, release lock
    lock.unlock();

    path.erase(std::strlen(ptr));
    return std::move(path);
  }

  // cannot use erase here: for example, "usr" maps to "."
  path = ptr; // avoids an extra allocation
  return std::move(path);
}
```

lib/CMakeLists.txt
patrick96: This makes a lot more sense

tests/CMakeLists.txt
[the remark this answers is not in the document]
patrick96: Oh, I see. Originally, the idea was to have each test specify all source file it relied on so that tests didn't need to compile every polybar source file. But some `.cpp` file are not decoupled at all and need to be compiled with nearly all other source file, that's why I used the `${files}` variable in those tests.

So yeah, this solution is a lot cleaner and does reduce compile times when the whole suite is compiled.

Maybe we can call it `poly`, so the library generated is `libpoly`.

`add_unit_test` is fine

patrick96:
Could you please open a new PR where you only have all the cmake changes. I want to nail those before continuing here. Also this PR does more than one thing (expand feature, non-trivial cmake refactoring) and that should be split up anyway.

PR #1507, opened 2018-11-11
fix(monitor): always use crtc monitor, even when WITH_XRANDR_MONITORS is enabled

### LINKED ISSUE #1437: Incorrect position and width on upscaled display
I am using dual monitor setup. The primary monitor DP-0 has a resolution of 3440x1440 and the left monitor HDMI-0 has an upscaled resolution of 1296x2304 (1920x1080 original).

screenshot from 2018-09-22 07-42-17

Bars are positioned correctly with the width set to 100% on the primary monitor. However, the bar on the left monitor is cut off even with width set to 100% and incorrectly vertically positioned. See below screenshot.

screenshot from 2018-09-22 07-34-29

I tried setting the width manually to more than 1080 but polybar logged an error `polybar|error: Resulting bar width is out of bounds`. I am also seeing `polybar|error: Failed to restack bar window` errors. I wasn't able to move the bar to the bottom of the screen through any configuration changes.

## To Reproduce

I am using bspwm. Launching the polybar normally with `polybar left`. The configuration can be viewed by clicking the Github link below.

[`polybar/config`](https://github.com/gnaqvi/dotfiles/blob/master/.config/polybar/config)

NBonaparte:
Fixes #1437.
This works on my setup, but I don't have any other computers to test on. This should work given that the crtc values are generally the same as the monitor values, but someone might have some funky setup where it doesn't.


NBonaparte:
Wait... this conflicts with #1426, need to fix that first.

patrick96:
Yeah, I was wondering about that.

Also if we remove the `#if WITH_XRANDR_MONITORS` blocks, the only thing `WITH_XRANDR_MONITORS` does is to change the output of `polybar -m`

NBonaparte:
I removed all references to `WITH_XRANDR_MONITORS` since it's not being used anymore.

patrick96:
What was the original intentation of `WITH_XRANDR_MONITORS`? Always seemed kind of useless to me.

Will try to reproduce the issue myself and see if this PR also works for me.

patrick96:
I have been doing some digging, and unfortunately `WITH_XRANDR_MONITORS` is here to stay. I have discovered this issue https://github.com/i3/i3/issues/1799 that describes that "monitors" are a new concept in randr and are different from "outputs" (The naming is just spectacular). They also listed [this](https://keithp.com/blogs/MST-monitors/) article, which may give some more info, but I haven't read it yet.

"outputs" are what `xrandr` will give you while `xrandr --listmonitors` will print the "monitors". On my system those will give about the same information, but the new monitor concept can be used to create two monitors inside a single output. When splitting an output in two like this, this patch will still only recognize the outputs (what `xrandr` outputs), while `WITH_XRANDR_MONITORS` will detect both monitors (what `xrandr --listmonitors` outputs).

I was using [this](https://askubuntu.com/a/998435/454032) answer to setup split outputs myself. The commands I used were:

```bash
xrandr --setmonitor H1 960/598x1080/336+0+0 HDMI-0
xrandr --setmonitor H2 960/598x1080/336+960+0 none
```

Where `HDMI-0` is the name of my output and `H1` and `H2` are the names for the newly created monitors. The dimensions of `HMDI-0` is `1920x1080+0+0`, so this splits that in half. The values after the slash seem to be the physical size in `mm`.

With this setup, `polybar -m` gives me

```
HDMI-0: 1920x1080+0+0
H2: 960x1080+960+0 (XRandR monitor)
H1: 960x1080+0+0 (XRandR monitor)
DisplayPort-0: 1920x1080+1920+0 (primary)
```

while this patch gives me
```
HDMI-0: 1920x1080+0+0
DisplayPort-0: 1920x1080+1920+0 (primary)
```

For comparison, this is the output of `xrandr --listmonitors`:
```
Monitors: 3
 0: +*DisplayPort-0 1920/598x1080/336+1920+0  DisplayPort-0
 1: H1 960/598x1080/336+0+0  HDMI-0
 2: H2 960/598x1080/336+960+0
```

I would argue, neither the current approach, no this PR handle this correctly. What i3 does, is to completely rely on "monitors" (`conn.get_monitors`) when RandR 1.5 is used and fall back to the output based approach on lower versions. What we currently do is to always use the outputs and also include "monitors" when RandR 1.5 is used. Switching to the i3 approach is probably the best idea.

I was not able to reproduce the issue in #1437, once I do, I think we will be able to both fix the issue and properly support randr monitors.

patrick96:
Closing this.

We now have better randr monitor support. Don't know if the linked issue is fixed though, I have not been able to reproduce it yet.

PR #1517, opened 2018-11-19
feat(cpu): Add normalize-load

### LINKED ISSUE #1516: Add non-normalized cpu %percentage%
Sometimes the normalized processor load (0% - 100%) is not very informative, especially in systems with a large number of cores.

Consider a system with 16 cores, then 100% load of one core will be drawn as 6% and each core has ~6 load levels. For example, both top and htop displays 100% as full load of only one core, therefore in such system full cpu usage will be displayed as 1600%. I think that this number may be better in some situations.

Just add a "normalize-load" option to the config that defaults to true, for example: https://github.com/BigRedEye/polybar/commit/8c7b4e400659bf7550c473f0247afc146d6b8132

To implement this functionality I have to use a script module that simply copy-pastes code from cpu.cpp with a few modified lines, but it looks like a useful feature that can be easily implemented.

BigRedEye:
Adding option not to normalize cpu usage (#1516)

EDIT: Closes #1516

### DIFF
--- a/include/modules/cpu.hpp
+++ b/include/modules/cpu.hpp
@@ -38,7 +38,6 @@ namespace modules {
     ramp_t m_rampload_core;
     label_t m_label;
     int m_ramp_padding;
-    bool m_normalize_load;
 
     vector<cpu_time_t> m_cputimes;
     vector<cpu_time_t> m_cputimes_prev;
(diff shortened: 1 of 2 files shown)


TheSilvus:
This could alternatively be implemented as a separate token (e.g. `%percentage-sum%`). 

          m_label->replace_token("%percentage-sum%", to_string(static_cast<int>(m_total * static_cast<float>(cores_n) + 0.5)));

I'd argue that this would be a simpler solution - it does not require adding another config option.

patrick96:
@TheSilvus that is what I proposed as well :wink:

PR #1518, opened 2018-11-20
coverage: Allow small drops in coverage

patrick96:
Most PRs change code where no tests have been written so far and where
adding tests is difficult. Codecov fails all those PRs.

These values should be adjusted once we have better integrated testing
into our workflow and have written better testing infrastructure for
modules and very tightly coupled components

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,3 +1,15 @@
+coverage:
+  status:
+    project:
+      default:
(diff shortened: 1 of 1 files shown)


patrick96:
Hold off on merging, I seem to have broken something

patrick96:
Alright, fixed it. Ready for review now.

PR #1523, opened 2018-11-23
Relative includes

sam-lunt:
Enables `include` commands in config files to specify relative file paths. Relative file paths will be evaluated with respect to the location of the including file (as in the C preprocessor).

To support this change, adds `is_absolute` and `dirname` functions to the `file_util` module.

Also provides a minor refactor to the `expand` command:

- only expand tilde when it is the start of a path name (which is the behavior of all shells)
- allow using environment variables anywhere in the path, not just at the start of a component
- allow using curly braces to more explicitly specify the name of the environment variables (e.g. allow `${HOME}/.config` in addition to `$HOME/.config`)

Adds unit tests for file inclusion and functions changed in `file_utils` (i.e. `is_absolute`, `dirname`, and `expand`).


src/components/config.cpp
```diff
@@ -101,16 +101,18 @@ void config::parse_file() {
     }
 
     if (key == "include-file") {
-      auto file_path = file_util::expand(value);
+      auto file_path = file_util::is_absolute(value)
```
patrick96: Can you put this into a seperate function that takes any path and returns an expanded absolute path please.

src/utils/file.cpp
```diff
@@ -210,6 +241,14 @@ namespace file_util {
     return stat(filename.c_str(), &buffer) == 0 && S_ISFIFO(buffer.st_mode);
   }
 
+  /*
+   * Checks if the (potentionally non-canonical) path is absolute
+   */
+  bool is_absolute(const string& filename) {
+    // this is sufficient for Unix-like systems
+    return !filename.empty() && (filename[0] == '/' || filename[0] == '~');
```
patrick96: Wouldn't this say `$HOME/.config/polybar/config` is relative?

src/components/config.cpp
sam-lunt: Good call. I added `expand_relative_to` as a function.

I wasn't sure the best order of the parameters: putting the directory first could make sense, since that's the order they would be appended together (`dir + '/' + path`). I put the path to be expanded first, since that kept the ordering consistent with normal `expand`.

I also don't call `dirname` on the `base_dir` argument. There's no way to inspect a path to decide if it's a file or a directory, and calling `stat` on the path seemed like overkill. Always calling dirname wouldn't work, since if someone had a directory and not a file argument, they'd need to add a fake file name component to get the result they want.

I don't call expand on the base_dir either, since it's likely already been done (plus, the caller can do that if they want to).

I don't assert that `base_dir` is an absolute path, because I could imagine a scenario where you want to expand a file relative to an absolute path.

src/utils/file.cpp
sam-lunt: Good catch. I made the `expand_relative_to` function expand the path before checking if it's absolute.

src/utils/file.cpp
```diff
@@ -233,27 +276,139 @@ namespace file_util {
...
+
+          auto env = env_util::get(path.substr(pos, count));
+          path.replace(replace_start, replace_end, env);
+
+          pos += count;
+        }
+
+        // erase any separators at the end (due to expansion) to avoid double '//'
+        while (!path.empty() && path.back() == '/') {
+            path.pop_back();
```
patrick96: Indentation
```suggestion
          path.pop_back();
```

src/utils/file.cpp
```diff
@@ -233,27 +276,139 @@ namespace file_util {
...
+        }
+
+        // remove separators at the start to avoid double '//'
+        string::size_type start = 0;
+        while (start < path.length() && path[start] == '/') {
+            ++start;
```
patrick96:
```suggestion
          ++start;
```

src/utils/file.cpp
```diff
@@ -233,27 +276,139 @@ namespace file_util {
...
+        }
+
+        if (start > 0) {
+            path = path.substr(start);
```
patrick96:
```suggestion
          path = path.substr(start);
```

src/utils/file.cpp
```diff
@@ -233,27 +276,139 @@ namespace file_util {
...
+        }
+
+        ++it;
       }
     }
+
+    if (p_exploded.empty()) {
+        return "/";
```
patrick96:
```suggestion
      return "/";
```

src/utils/file.cpp
```diff
@@ -233,27 +276,139 @@ namespace file_util {
...
       ret.insert(0, 1, '/');
+    }
+
     return ret;
   }
+
+  /**
+   * Path expansion, making relative paths absolute
+   */
+  string expand_relative_to(const string& path, const string& base_dir) {
```
patrick96: Please also add tests here. I think there may be some issues if path is something like `../bin` or starts with an env variable.

src/utils/file.cpp
```diff
@@ -233,27 +276,139 @@ namespace file_util {
...
+        it = p_exploded.erase(it);
+      } else if (*it == "..") {
+        if (it != p_exploded.begin()) {
+          --it;
+          it = p_exploded.erase(it);
+        }
+
+        it = p_exploded.erase(it);
+      } else {
+        auto& path = *it;
```
patrick96: This is confusing because it has the same name as the parameter

src/utils/file.cpp
```diff
@@ -233,27 +276,139 @@ namespace file_util {
...
-    vector<string> p_exploded = string_util::split(path, '/');
-    for (auto& section : p_exploded) {
-      switch(section[0]) {
-        case '$':
-          section = env_util::get(section.substr(1));
-          break;
-        case '~':
-          section = env_util::get("HOME");
-          break;
+  string expand(const string& path) {
```
patrick96: Please document the code inside this function better as its quite a big chunk of code.

tests/unit_tests/components/config.cpp
```diff
@@ -0,0 +1,143 @@
...
+  class temp_directory {
+  public:
+    temp_directory(string directory) {
+      if (file_util::is_absolute(directory)) {
+        m_directory = directory;
+      } else {
+        m_directory = "/tmp/" + directory;
+      }
+
+      if (m_directory.length() < 6 || m_directory.compare(m_directory.length() - 6, 6, "XXXXXX") != 0) {
```
patrick96: We can also always directly add `_XXXXXX` to the path, why would any caller of this function want to manually add the Xs.

tests/unit_tests/components/config.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  private:
+    temp_filestream(string directory, string filename)
+      : fd_stream<std::iostream>(open_temp_file(directory, filename))
+      , m_directory(std::move(directory))
+      , m_filename(std::move(filename))
+      , m_full_path(m_directory + "/" + m_filename)
+    {}
+
+    static int open_temp_file(string& directory, string& filename) {
```
patrick96: Doesn't this modify the arguments of the `temp_filestream` constructor because we use references here?

tests/unit_tests/components/config.cpp
```diff
@@ -0,0 +1,143 @@
...
+      directory += "/";
+      directory += filename;
+
+      auto fd = mkstemp(&directory[0]);
```
patrick96: Is this a good idea to pass the raw data pointer of `std::string` to a function that will modify the contents?

tests/unit_tests/components/config.cpp
```diff
@@ -0,0 +1,143 @@
+#include <iostream>
```
patrick96: These tests would greatly benefit from using [test fixtures](https://github.com/google/googletest/blob/master/googletest/docs/primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests). The code in the tests is basically the same and the setup code could probably be reused for lots of different tests.

patrick96:
C++17 actually has lots of utility for [paths](https://en.cppreference.com/w/cpp/filesystem/path). Unfortunately we only use C++14

src/utils/file.cpp
```diff
@@ -165,6 +166,40 @@ int fd_streambuf::underflow() {
 // }}}
 
 namespace file_util {
+  namespace {
```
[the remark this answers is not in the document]
sam-lunt: It gives the helper functions internal linkage. Not only does it allow the compiler to be more aggressive when optimizing (mostly in terms of what it will inline), but it also can avoid some weird errors when building with link-time optimization (which are unlikely to happen here, but in general it's helpful).

tests/unit_tests/components/config.cpp
sam-lunt: Yes, but the arguments are passed by value and then moved in to the data member. That ensures that the `m_directory` member has the real path name of the temp directory, not the template passed to `mkstemp`. (This is private and only called in the constructor)

tests/unit_tests/components/config.cpp
sam-lunt: C++11 guarantees that `std::string`'s buffer is null-terminated, and mkstemp only replaces the "X"s from the template with the actual value, so I don't see any big risk.

src/utils/file.cpp
sam-lunt: Good catch, changed it to "component"

src/utils/file.cpp
sam-lunt: Good point, I added more comments, let me know if that helps

tests/unit_tests/components/config.cpp
sam-lunt: Hypothetically, they might want more than six "X"s or a dot instead of an underscore. But even if I did assume it was always added by the function, I'd still want an assert in there, just to be safe

tests/unit_tests/utils/file.cpp
```diff
@@ -7,11 +7,92 @@
...
+  EXPECT_EQ(file_util::expand("/tmp/~root"), "/tmp/~root");
+  EXPECT_EQ(file_util::expand("~/~"), expand_with_shell("~/~"));
+  EXPECT_EQ(file_util::expand("~/~"), expand_with_shell("$HOME/~"));
+
+  // HOME is expanded anywhere
+  EXPECT_EQ(file_util::expand("/tmp/$HOME"), expand_with_shell("/tmp$HOME"));
+  EXPECT_EQ(file_util::expand("/tmp/$HOME"), "/tmp" + expand_with_shell("~"));
 }
 
+TEST(File, dirname) {
```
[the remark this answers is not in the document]
sam-lunt: I don't think there's a lot of repetition between test cases, since each test has input specific to that particular function. I did refactor it so that the expected input and output are all declared first, and then the test runs for each pair. I think that accomplishes what you're describing without needing to set up a fixture or use the rest of the "parameterized test case" machinery.

src/utils/file.cpp
sam-lunt: Good call with regards to the `../bin` scenario, that didn't work correctly. If the base_dir was something like `/usr/bin` and the relative path was `../lib`, the result wouldn't be `/usr/lib` as the it should have been. I refactored the `is_absolute` check to work on paths with unexpanded variables and then switched up the relative expansion logic so that it handles the above case correctly. Also added some test cases for the scenarios.

sam-lunt:
`std::filesystem` would certainly be nice to use, but it's just barely implemented in the major standard library implementations, so not only would it require a bump of the C++ version, it would also require pretty recent compilers. Personally, I think if you avoid the exotic file system that is Windows, you don't have too pressing a need for the library, haha.

I addressed the above comments and added in more unit tests. I rewrote some of the absolute path checking and relative path expansion logic to address the issue where the path is something like `"../bin"`. This also allows `is_absolute` to handle paths with shell variables at the start.

tests/unit_tests/components/config.cpp
patrick96: I don't think we need to be that flexible in our tests, the main point here is to have a unique temporary directory. Whether or not it has an underscore or a period doesn't matter, I think.

patrick96:
Closing due to inactivity

PR #1528, opened 2018-11-25
Added support for bridge interface

### LINKED ISSUE #1522: Bridge Interface Error - Failed to query interface 'br0'
## Describe the issue
The network interface "br0" is not displayed. But it works with a physical network adapter e.g. "eno1".

## To Reproduce
```dosini
modules-right = network

[module/network]
type = internal/network
interface = "br0"
label-connected = "%ifname% %linkspeed%"
label-disconnected = disconnected
```

## Polybar Log
```
warn: module/network: Failed to query interface 'br0'
warn: module/network: Failed to query interface 'br0'
warn: module/network: Failed to query interface 'br0'
warn: module/network: Failed to query interface 'br0'
warn: module/network: Failed to query interface 'br0'
```

## Environment:
* WM: i3
* PB: polybar 3.2.1

spsarolkar:
Currently the polybar fails to detect bridge interface, Issue reference :- https://github.com/jaagr/polybar/issues/1522.

I have updated the code to add the support for bridge interface

EDIT: Fixes #1522

### DIFF
--- a/src/adapters/net.cpp
+++ b/src/adapters/net.cpp
@@ -254,9 +254,9 @@ namespace net {
     }
 
     if(m_bridge) {
-    /* If bridge network then link speed cannot be computed
-     * TODO: Identify the physical network in bridge and compute the link speed
-     */
+      /* If bridge network then link speed cannot be computed
+       * TODO: Identify the physical network in bridge and compute the link speed
+       */
       return true;
     }
 
--- a/src/adapters/net.cpp
+++ b/src/adapters/net.cpp
@@ -260,13 +260,13 @@ namespace net {
     struct ifreq request {};
     struct ethtool_cmd data {};
 
-	memset(&request, 0, sizeof(request));
-	strncpy(request.ifr_name, m_interface.c_str(), IFNAMSIZ - 1);
-	data.cmd = ETHTOOL_GSET;
-	request.ifr_data = reinterpret_cast<char*>(&data);
+    memset(&request, 0, sizeof(request));
+    strncpy(request.ifr_name, m_interface.c_str(), IFNAMSIZ - 1);
+    data.cmd = ETHTOOL_GSET;
+    request.ifr_data = reinterpret_cast<char*>(&data);
 
     if (ioctl(*m_socketfd, SIOCETHTOOL, &request) == -1) {
-	  return false;
+      return false;
 	}
 
     m_linkspeed = data.speed;
--- a/src/adapters/net.cpp
+++ b/src/adapters/net.cpp
@@ -253,21 +253,23 @@ namespace net {
       return false;
     }
 
-    if(!m_bridge) { // If bridge network then link speed cannot be computed TODO: Identify the physical network in bridge and compute the link speed
-	  struct ifreq request {};
-	  struct ethtool_cmd data {};
-
-	  memset(&request, 0, sizeof(request));
-	  strncpy(request.ifr_name, m_interface.c_str(), IFNAMSIZ - 1);
-	  data.cmd = ETHTOOL_GSET;
-	  request.ifr_data = reinterpret_cast<char*>(&data);
-
-	  if (ioctl(*m_socketfd, SIOCETHTOOL, &request) == -1) {
-	    return false;
-	  }
-
-	  m_linkspeed = data.speed;
+    if(m_bridge) { // If bridge network then link speed cannot be computed TODO: Identify the physical network in bridge and compute the link speed
+      return true;
     }
+
+    struct ifreq request {};
+    struct ethtool_cmd data {};
+
+	memset(&request, 0, sizeof(request));
+	strncpy(request.ifr_name, m_interface.c_str(), IFNAMSIZ - 1);
+	data.cmd = ETHTOOL_GSET;
+	request.ifr_data = reinterpret_cast<char*>(&data);
+
+    if (ioctl(*m_socketfd, SIOCETHTOOL, &request) == -1) {
+	  return false;
+	}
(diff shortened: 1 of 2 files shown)


src/adapters/net.cpp
```diff
@@ -248,19 +253,21 @@ namespace net {
...
 
-    if (ioctl(*m_socketfd, SIOCETHTOOL, &request) == -1) {
-      return false;
-    }
+	  if (ioctl(*m_socketfd, SIOCETHTOOL, &request) == -1) {
+	    return false;
+	  }
 
-    m_linkspeed = data.speed;
+	  m_linkspeed = data.speed;
```
patrick96: Instead of wrapping the whole thing in an if block, please only set the linkspeed if we don't use a bridge interface. We still need to check, if the interface is actually connected.

Also, please use 2 spaces for indentation, no tabs.

spsarolkar:
@patrick96 I have made the changes as per your review comments, also please find the output to command `ethtool -i br0`

▶ ethtool -i br0
driver: bridge
version: 2.3
firmware-version: N/A
expansion-rom-version: 
bus-info: N/A
supports-statistics: no
supports-test: no
supports-eeprom-access: no
supports-register-dump: no
supports-priv-flags: no


Let me know changes required if any

src/adapters/net.cpp
spsarolkar: I have made the changes as per your review comments

src/adapters/net.cpp
```diff
@@ -204,6 +204,11 @@ namespace net {
     } else {
       m_tuntap = false;
     }
+
+    if (strncmp(driver.driver, "bridge", 3) == 0) {
```
patrick96:
```suggestion
    if (strncmp(driver.driver, "bridge", 6) == 0) {
```

Otherwise this only compares the first three bytes

src/adapters/net.cpp
```diff
@@ -258,7 +267,7 @@ namespace net {
 
     if (ioctl(*m_socketfd, SIOCETHTOOL, &request) == -1) {
       return false;
-    }
+	}
```
patrick96:
```suggestion
    }
```

src/adapters/net.cpp
```diff
@@ -248,6 +253,10 @@ namespace net {
       return false;
     }
 
+    if(m_bridge) { // If bridge network then link speed cannot be computed TODO: Identify the physical network in bridge and compute the link speed
```
patrick96: This is not what I meant. We still need to query the interface with `ioctl`, but we only set `m_linkspeed` if we are not using a bridge.

spsarolkar: We are already checking the connected status as a part of the call to update 
`
m_connected = network->connected();
`

And above call makes use of `ioctl` call to check the status and it works in case of bridge interface.

If we compute the link speed on bridge interface the call fails and false is returned

`
if (ioctl(*m_socketfd, SIOCETHTOOL, &request) == -1) {
      return false;
    }
`

Did you mean to return true from above else block in case of bridge interface?

patrick96: Ah, I see, I though we needed that `ioctl` to also check if the interface works properly not just to get the linkspeed.

Say, what is the content of `/sys/class/net/br0/speed` on your system?

patrick96: Please add this to the "if-else-if" block at the beginning. 

```suggestion
    } else if (m_bridge) { 
    /* If bridge network then link speed cannot be computed 
     * TODO: Identify the physical network in bridge and compute the link speed
     */
```

spsarolkar: Hi @patrick96 , I am not sure if you mean this

```suggestion
if (m_tuntap) {
      return true;
    } else if(m_bridge) { // If bridge network then link speed cannot be computed TODO: Identify the physical network in bridge and compute the link speed
      return true;
    } else if (!network::query(accumulate)) {
      return false;
    } 
```

If we insert if-else-if as above `network::query(accumulate)` would never get executed. This call calculates the real time upload and download speed for the interface and it works fine for bridge interface.

We need `network::query(accumulate)` to be executed because this gives some additional information about the current network activity. That's why I have kept below if statement separate after we execute this call.

```
if(m_bridge) { // If bridge network then link speed cannot be computed TODO: Identify the physical network in bridge and compute the link speed
      return true;
}
```

spsarolkar: Please find contents of cat /sys/class/net/br0/speed

$ cat /sys/class/net/br0/speed
cat: /sys/class/net/br0/speed: Invalid argument

patrick96: >  cat: /sys/class/net/br0/speed: Invalid argument

Seems this is consistent with ethtool

> If we insert if-else-if as above network::query(accumulate)

No, I meant
```
if m_tuntap
elseif !network::query
elseif m_bridge
```

spsarolkar: If I add above if-else-if block, the second if block ‘!network::query’ succeeds and it stops returning true in case of bridge interface.  Which proceeds with link speed computation using ‘ioctl’ and fails returning false.

Thats why I have added it as a separate if statement to return true and stop link speed computation

patrick96: :man_facepalming: you're right. I knew there was some flaw in my logic ^^

But please put the comment on a separate line

patrick96:
```suggestion
    if(m_bridge) { 
    /* If bridge network then link speed cannot be computed 
     * TODO: Identify the physical network in bridge and compute the link speed
     */
```

PR #1533, opened 2018-11-27
fix: strict monitor name matching

### LINKED ISSUE #1532: Polybar running in the wrong monitor
## Describe the issue
The application doesn't run in the right monitor.

```config```:
```
[bar/bottom]
monitor = ${env:MONITOR}
...
```

```polybar -m```:
```
HDMI2: 1360x768+0+0
DP1: 1280x1024+4480+824
HDMI-2: 1920x1080+2560+768
HDMI-0: 2560x1080+0+768
```

Polybar can't see the difference over HDMI2 and HDMI-2, as it won't start in HDMI-2, only in HDMI2.

The ```polybar -l info``` at the command ```MONITOR=HDMI2 polybar -l info bottom``` shows:
```* Loaded monitor HDMI2 (1360x768+0+0)```

The ```polybar -l info``` for ```MONITOR=HDMI-2 polybar -l info bottom``` also shows:
```* Loaded monitor HDMI2 (1360x768+0+0)```

**Expected behavior:**
Polybar is expected to load in the right monitor.

**Actual behavior:**
Polybar loads in the wrong monitor.

## To Reproduce
1. Have a output named HDMI2
2. Have another output named HDMI-2
3. Try to run polybar in both outputs

## Additional context
If I remove the replace at the source file ```x11/extensions/randr.cpp```, line 23, it works as expected, but I don't know if it has any other implications in the overall functionality.

### PR DESCRIPTION
**Pull request by patrick96**
~randr_output::match would match regardless of if there is a dash in the
monitor name.
This was done because jaagr's monitors didn't have a dash in them
depending on the graphics driver used. This behavior was introduced in
[1].~

~This behavior creates issues in situations where a user has a monitor
with a dash in its name and one without e.g. HDMI-2 and HDMI2 (see
 #1532). In both cases polybar would be loaded on monitor HDMI2.~

~Monitor name matching should always be strict~

~[1]: 4d7f6c14e65332793e313866ba05539d3776cf65~

EDIT: This now introduces `monitor-exact` in the bar section which defaults to true and if set to false does fuzzy monitor matching (ignores dashes in monitor names).

It also removes `monitor-strict` from the xbacklight module and uses `monitor-strict` from the bar section. This config option was never documented, so it shouldn't break too many setups (which can all easily be fixed).

Fixes #1532

### DIFF
--- a/src/x11/extensions/randr.cpp
+++ b/src/x11/extensions/randr.cpp
@@ -12,15 +12,17 @@
 POLYBAR_NS
 
 /**
- * Workaround for the inconsistent naming
- * of outputs between my intel and nvidia
+ * Workaround for the inconsistent naming of outputs between intel and nvidia
  * drivers (xf86-video-intel drops the dash)
+ *
+ * If exact == false, dashes will be ignored while matching
  */
-bool randr_output::match(const string& o, bool strict) const {
-  if (strict && name != o) {
-    return false;
+bool randr_output::match(const string& o, bool exact) const {
+  if (exact) {
+    return name == o;
+  } else {
+    return string_util::replace(name, "-", "") == string_util::replace(o, "-", "");
   }
-  return name == o || name == string_util::replace(o, "-", "");
 }
 
 /**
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by patrick96** [comment]
This "feature" isn't actually documented anywhere, so I don't know how many people rely on it. But sure, an option to enable it would be nice. I don't like that `monitor-strict` does two things: fuzzy matching and matching disconnected monitors. So we could add an option `monitor-exact`.

The main problem I see here (and also what happened in #1532) is that the monitor matching code does "first-match" instead of "best-match":
```cpp
  for (auto&& monitor : monitors) {
    if (!name_found && (name_found = monitor->match(monitor_name, monitor_strictmode))) {
      m_opts.monitor = move(monitor);
    } else if (!fallback_found && (fallback_found = monitor->match(monitor_name_fallback, monitor_strictmode))) {
      fallback = move(monitor);
    }

    if (name_found && fallback_found) {
      break;
    }
  }
```

But I don't think it would add to much complexity to implement that. I'll get right on that.

patrick96:
Done. I have also done some refactoring because getting the monitor pointer for a name used very similar code everywhere.

src/x11/extensions/randr.cpp
```diff
@@ -12,15 +12,18 @@
...
+ *
+ * If exact == false, dashes will be ignored while matching
  */
-bool randr_output::match(const string& o, bool strict) const {
-  if (strict && name != o) {
-    return false;
+bool randr_output::match(const string& o, bool exact) const {
+  if(exact) {
+    return name == o;
+  }
```
NBonaparte:
```suggestion
  } else
```

PR #1534, opened 2018-11-28
controller: Print error for duplicate modules

patrick96:
You can't use the same name twice inside the module lists

E.g.
```dosini
modules-left = a b c
modules-center = a
modules-right = b
```
would print an error.

We only print an error for now because we don't want to break existing
configs. But in the future this should be properly enforced.

This is in preparation for fixing #1172, because I plan on making modules uniquely identifiable by their name.

### DIFF
--- a/src/components/controller.cpp
+++ b/src/components/controller.cpp
@@ -105,7 +105,9 @@ controller::controller(connection& conn, signal_emitter& emitter, const logger& 
           throw application_error("Inter-process messaging needs to be enabled");
         }
 
-        m_modules[align].emplace_back(make_module(move(type), m_bar->settings(), module_name, m_log));
+        module_t module = shared_ptr<modules::module_interface>(make_module(move(type), m_bar->settings(), module_name, m_log));
+        m_modules.push_back(module);
+        m_blocks[align].push_back(module);
         created_modules++;
       } catch (const runtime_error& err) {
         m_log.err("Disabling module \"%s\" (reason: %s)", module_name, err.what());
@@ -137,15 +139,13 @@ controller::~controller() {
   m_sig.detach(this);
 
   m_log.trace("controller: Stop modules");
-  for (auto&& block : m_modules) {
-    for (auto&& module : block.second) {
-      auto module_name = module->name();
-      auto cleanup_ms = time_util::measure([&module] {
-        module->stop();
-        module.reset();
-      });
-      m_log.info("Deconstruction of %s took %lu ms.", module_name, cleanup_ms);
-    }
+  for (auto&& module : m_modules) {
+    auto module_name = module->name();
+    auto cleanup_ms = time_util::measure([&module] {
+      module->stop();
+      module.reset();
+    });
+    m_log.info("Deconstruction of %s took %lu ms.", module_name, cleanup_ms);
   }
 
   m_log.trace("controller: Joining threads");
@@ -171,26 +171,24 @@ bool controller::run(bool writeback, string snapshot_dst) {
   m_sig.attach(this);
 
   size_t started_modules{0};
-  for (const auto& block : m_modules) {
-    for (const auto& module : block.second) {
-      auto inp_handler = dynamic_cast<input_handler*>(&*module);
-      auto evt_handler = dynamic_cast<event_handler_interface*>(&*module);
+  for (const auto& module : m_modules) {
+    auto inp_handler = dynamic_cast<input_handler*>(&*module);
+    auto evt_handler = dynamic_cast<event_handler_interface*>(&*module);
 
-      if (inp_handler != nullptr) {
-        m_inputhandlers.emplace_back(inp_handler);
-      }
+    if (inp_handler != nullptr) {
+      m_inputhandlers.emplace_back(inp_handler);
+    }
 
-      if (evt_handler != nullptr) {
-        evt_handler->connect(m_connection);
-      }
+    if (evt_handler != nullptr) {
+      evt_handler->connect(m_connection);
+    }
 
-      try {
-        m_log.info("Starting %s", module->name());
-        module->start();
-        started_modules++;
-      } catch (const application_error& err) {
-        m_log.err("Failed to start '%s' (reason: %s)", module->name(), err.what());
-      }
+    try {
+      m_log.info("Starting %s", module->name());
+      module->start();
+      started_modules++;
+    } catch (const application_error& err) {
+      m_log.err("Failed to start '%s' (reason: %s)", module->name(), err.what());
     }
   }
 
@@ -465,7 +463,7 @@ bool controller::process_update(bool force) {
   string margin_left(bar.module_margin.left, ' ');
   string margin_right(bar.module_margin.right, ' ');
 
-  for (const auto& block : m_modules) {
+  for (const auto& block : m_blocks) {
     string block_contents;
     bool is_left = false;
     bool is_center = false;
@@ -586,11 +584,9 @@ bool controller::on(const signals::eventqueue::exit_reload&) {
  * Process eventqueue check event
  */
 bool controller::on(const signals::eventqueue::check_state&) {
-  for (const auto& block : m_modules) {
-    for (const auto& module : block.second) {
-      if (module->running()) {
-        return true;
-      }
+  for (const auto& module : m_modules) {
+    if (module->running()) {
+      return true;
     }
   }
   m_log.warn("No running modules...");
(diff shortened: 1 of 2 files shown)


src/components/controller.cpp
```diff
@@ -76,42 +76,33 @@ controller::controller(connection& conn, signal_emitter& emitter, const logger&
...
 
-        m_modules[align].emplace_back(make_module(move(type), m_bar->settings(), module_name, m_log));
-        created_modules++;
-      } catch (const runtime_error& err) {
-        m_log.err("Disabling module \"%s\" (reason: %s)", module_name, err.what());
+  /*
+   * Check if each module name only appears once
+   * Iterate over each module and check if its name appears anywhere else
+   */
+  for (auto it = m_modules.begin(); it != m_modules.end(); it++) {
```
Lomadriel: This loop prints the error log X times if the module appears X times (with X > 2).
If you really want to really address this problem you can replace the whole loop with something like this:
```cpp
  std::sort(m_modules.begin(), m_modules.end(), [](const auto& m1, const auto& m2) { return m1->name() < m2->name(); });

  auto dup_it = m_modules.cbegin();

  do {
  	auto equal_predicate = [](const auto& m1, const auto& m2) { return m1->name() == m2->name(); };
    dup_it = std::adjacent_find(dup_it, m_modules.cend(), equal_predicate);

    if (dup_it != m_modules.cend()) {
	    auto first_mismatch = std::mismatch(dup_it, m_modules.cend(), dup_it, equal_predicate).first;
	    m_log.err(
          "The module \"%s\" appears %d times in your modules list. "
          "This should be avoided, as it can lead to inconsistent behavior. "
          "Both modules will be displayed for now.",
          (*dup_it)->name(), std::distance(dup_it, first_mismatch));

	    dup_it = first_mismatch;
    } else {
    	break;
    }
  } while (true);
```

src/components/controller.cpp
```diff
@@ -76,42 +76,33 @@ controller::controller(connection& conn, signal_emitter& emitter, const logger&
...
+    auto name = it->get()->name();
```
Lomadriel:
```suggestion
    const auto& name = it->get()->name();
```

Avoid copying the name each time.

EDIT:
The method `virtual string name() const = 0;` doesn't return a const ref but definitively should.

src/components/controller.cpp
```diff
@@ -671,15 +714,13 @@ bool controller::on(const signals::ipc::command& evt) {
...
-      }
-      auto ipc = dynamic_cast<ipc_module*>(module.get());
-      if (ipc != nullptr) {
-        ipc->on_message(hook);
-      }
+  for (const auto& module : m_modules) {
+    if (!module->running()) {
+      continue;
+    }
+    auto ipc = dynamic_cast<ipc_module*>(module.get());
```
Lomadriel:
```suggestion
    auto ipc = std::dynamic_pointer_cast<ipc_module>(module);
```

src/components/controller.cpp
```diff
@@ -127,15 +118,13 @@ controller::~controller() {
...
-        module->stop();
-        module.reset();
-      });
-      m_log.info("Deconstruction of %s took %lu ms.", module_name, cleanup_ms);
-    }
+  for (auto&& module : m_modules) {
+    auto module_name = module->name();
+    auto cleanup_ms = time_util::measure([&module] {
+      module->stop();
+      module.reset();
```
Lomadriel:
```suggestion
```
This line is now useless, since there are several instances of this `shared_ptr` it doesn't destroy anything.
Another solution is to add the following code before this loop but I don't think this is very useful since the module are deleted automatically when they objects are destroyed.
```cpp
m_blocks.clear();
```

src/components/controller.cpp
```diff
@@ -161,26 +150,24 @@ bool controller::run(bool writeback, string snapshot_dst) {
   m_sig.attach(this);
 
   size_t started_modules{0};
-  for (const auto& block : m_modules) {
-    for (const auto& module : block.second) {
-      auto inp_handler = dynamic_cast<input_handler*>(&*module);
-      auto evt_handler = dynamic_cast<event_handler_interface*>(&*module);
+  for (const auto& module : m_modules) {
+    auto inp_handler = dynamic_cast<input_handler*>(&*module);
```
Lomadriel: Maybe, `std::dynamic_pointer_cast` can be used here (https://en.cppreference.com/w/cpp/memory/shared_ptr/pointer_cast). If so `m_inputhandlers` must store `std::shared_ptr`.

src/components/controller.cpp
Lomadriel: Maybe the message should state this is deprecated, even if the error already indicates the problem.

src/components/controller.cpp
```diff
@@ -542,6 +529,64 @@ bool controller::process_update(bool force) {
...
+    try {
+      auto type = m_conf.get("module/" + module_name, "type");
+
+      if (type == "custom/ipc" && !m_ipc) {
+        throw application_error("Inter-process messaging needs to be enabled");
+      }
+
+      auto ptr = make_module(move(type), m_bar->settings(), module_name, m_log);
+      module_t module = shared_ptr<modules::module_interface>(ptr);
+      ptr = NULL;
```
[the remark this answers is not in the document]
patrick96: Why? It's illegal to use the value of `ptr` after this point so we best reset it.

src/components/controller.cpp
patrick96: This doesn't really have anything to do with this PR. We'll do this once we rework the event system which will require us to change how we handle input handlers anyway.

src/components/controller.cpp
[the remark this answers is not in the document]
Lomadriel: Oops, that a fail. I edited my first comment.
However maybe it's better to remove that variable but it would make the line quite long :/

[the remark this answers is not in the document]
patrick96: Oh I see, that makes more sense

src/components/controller.cpp
Lomadriel: Yeah you are right sorry.

src/components/controller.cpp
patrick96: This doesn't print an error message for all modules that appear multiple times though. I think you have a bug in there somewhere.

I agree, I'll change the error message.

Lomadriel: > This doesn't print an error message for all modules that appear multiple times though. I think you have a bug in there somewhere.

Could you give me an example when it doesn't work? I'll post a corrected version.

patrick96:
```
[bar/example]
modules-left = a b c
modules-center = a b
modules-right = b a
```

It says `a` appears 7 times.

Lomadriel: Thanks !

patrick96: I found a fix.

src/components/controller.cpp
```diff
@@ -80,42 +80,37 @@ controller::controller(connection& conn, signal_emitter& emitter, const logger&
...
+          "This is deprecated and should be avoided, as it can lead to inconsistent behavior. "
+          "Both modules will be displayed for now.",
+          (*dup_it)->name());
 
-        m_modules[align].emplace_back(make_module(move(type), m_bar->settings(), module_name, m_log));
-        created_modules++;
-      } catch (const runtime_error& err) {
-        m_log.err("Disabling module \"%s\" (reason: %s)", module_name, err.what());
-      }
+      dup_it = std::find_if_not(dup_it, m_modules.cend(), std::bind(equal_predicate, *dup_it, std::placeholders::_1));
```
Lomadriel: Maybe a lambda is better but I'm not really sure.

I don't know at which point the compiler is able to optimize the `std::bind` but technically the object returned by `std::bind` can use polymorphism. A lambda is a anonymous struct and so doesn't use polymorphism.

patrick96: It really doesn't matter this is only called once on startup and maybe for a dozen modules.

patrick96:
> Should we fix #412 before merging this?
Since we deprecate using two time the same module, I think we need to promote an alternative which is not copy/paste.

I don't think so. We only prohibit having the exact same module in the bar multiple times. I don't think anyone even has that in their config since they will display the exact same thing. You can still have the same module type multiple times.

Lomadriel:
> > Should we fix #412 before merging this?
> > Since we deprecate using two time the same module, I think we need to promote an alternative which is not copy/paste.
> 
> I don't think so. We only prohibit having the exact same module in the bar multiple times. I don't think anyone even has that in their config since they will display the exact same thing. You can still have the same module type multiple times.

Indeed that's true.

Issue #346, opened 2017-01-15
custom/scripts: exec-else and {if-,else-,}label-foreground


Tmplt:
I have a few custom scripts I want to create, but the current features of `custom/script` limits me a bit (unless I've missed something): could an `exec-else` key be added so something else is printed if `exec-if` returns non-zero? On top of that, adding a key to specify the label colour depending on the return value?

For example: I want a vpn module that tells me whether I'm connected, and if I'm not, the label will be printed in a colour that stands out. The implementation could look like:

```ini
[module/vpn]
type = custom/script
interval = 5

format-prefix = "vpn: "
label = %output%

exec-if = pgrep openvpn
exec = print "online\n"
exec-else = print "offline\n" ; executes if exec exits non-zero

; or perhaps {success,fail}-label-foreground?
if-label-foreground = ${colors.green} ; if exec exits zero
else-label-foreground = ${colors.red} ; if exec exits non-zero
;label-foreground = ${colors.foreground} ; for the sake of completion, used when there are no exec-{if,else}
```
At present `exec = pgrep openvpn >/dev/null && printf "online\n" || printf "offline\n"` suffices for the status, but I'm not sure how colours would be done. I'm guessing if it's printed in a colour, polybar will use it? Any case, handling colour in polybar directly would be much smoother.

Any thoughts on this?

alekcacko:
I would also like this feature. I am new with polybar, so I can't see if there are chances to achieve what you want with current features.

jaagr:
This would be a nice feature, currently you would have to print raw tags from the script to get different color, i.e: `%{F#f00}failure%{F-}`

andyrichardson:
I'm looking to do something similar to this. I have a script which returns _on_ or _off_ representing the state of the VPN.

It would be great if I could then just use `label-on = connected` and `label-off = disconnected`. This style could be used with formatting too `format-on-underline = #112233`.

GregSharpe1:
:+1:

pnotz17:
is there a way to achieve with pacman updates only to show when there is a update?and as soon as you update for the to disappear?i tried this below and it didn't work

```
type = custom/script
exec =  pacman -Qu | wc -l
;exec-if = ./.config/polybar/check-all-updates.sh or  pacman -Qu | wc -l
interval = 100
label =  %output%
format-foreground =#FF0000

```

amit-raut:
[...]

PR #1538, opened 2018-12-05
feat(script): add exec-else and format-else

Edwearth:
It executes `exec-else` instead of `exec` if `exec-if` returns false.
It also add `format-else` to use a different format if `exec-if` returns false. 
see: https://github.com/jaagr/polybar/issues/346

Example:

```
[module/wifi]
type = custom/script

exec-if = is_wifi_on.sh
exec = echo 'connected'
exec-else = echo 'disconnected'

format-foreground = #0f0

format-else-foreground = #f00
```

### DIFF
--- a/include/modules/script.hpp
+++ b/include/modules/script.hpp
@@ -7,6 +7,8 @@
 POLYBAR_NS
 
 namespace modules {
+  enum script_state { NONE = 0, IF, ELSE };
+
   class script_module : public module<script_module> {
    public:
     explicit script_module(const bar_settings&, string);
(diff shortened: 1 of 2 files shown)


patrick96:
I thought, I already commented on this, but maybe I forgot to send it.

@infokiller Yes it can be. 

That's the main reason I didn't like the idea of having `exec-else`. I'm still not quite sure if it's something I want in the codebase.

The main argument I see for `exec-else` is that, if you are using `exec-if` you are restricting what you can because you cannot control what is displayed if the condition is false. But that's also quite a good argument for removing `exec-if`, because both are basically just moving parts of your script into polybar.
The other argument would be that with this you can take advantage of easier formatting because you don't have to use format tags.

@Edwearth thank you very much for implementing this. I will dig deeper into this in the coming weeks and decide then, if we want this functionality put into polybar.

Edwearth:
I completely agree with you. I think that having an `if` statement without an `else` is a bit counter-intuitive. Furthermore  the `exec-if` is a bit useless without the `else` since it has no impact on the format and you can just add an `if` in the `exec` part.
The most important part of this PR is the `format-else`, it allows easy formatting without writing color label "manually".
Take your time and thanks for your reply :)

PR #1540, opened 2018-12-07
Documentation in repo

### LINKED ISSUE #1611: jsoncpp newest version still below minimum required and fatal: not a git repository
## Build Process
*Describe how you build polybar, list the exact commands you are using:*
`mkdir build`
`cd build`
`cmake ..`

**If you build polybar directly from this repository:**

* Output of `git describe --tags`:

**If you use some other way (like the AUR):**

*List exactly where you are building from and which version you are building.*

I am building from the github repos releases page.
`polybar-3.3.0.tar`

## Build Log
*Post everything that is output to the terminal while building polybar below. This HAS to include the output of the `cmake` and `make` commands, if you used them.*
```
thewithz@TheWithzUbuntu:~/lib/polybar/build$ cmake ..
-- Trying to enable ccache
-- Couldn't locate ccache, disabling ccache...
-- No build type specified; using RelWithDebInfo
-- Using supported compiler GNU-5.4.0
--  Build:
--    Type: RelWithDebInfo
--    CC: /usr/bin/cc  -O2 -g -DNDEBUG
--    CXX: /usr/bin/c++  -Wall -Wextra -Werror -Wno-noexcept-type -pedantic -pedantic-errors -O2 -g -DNDEBUG
--    LD: /usr/bin/ld  
--  Targets:
-- [X]   polybar-msg
-- [ ]   testsuite
--  Module support:
-- [ ]   alsa
-- [X]   curl (7.47.0)
-- [X]   i3
-- [ ]   mpd
-- [ ]   network (wireless-tools)
-- [ ]   pulseaudio
-- [ ]   xkeyboard
--  X extensions:
-- [X]   xcb-randr (1.11.1)
-- [ ]   xcb-randr (monitor support)
-- [ ]   xcb-render
-- [ ]   xcb-damage
-- [ ]   xcb-sync
-- [X]   xcb-composite (1.11.1)
-- [ ]   xcb-xkb
-- [ ]   xcb-xrm
-- [ ]   xcb-cursor
--  Log options:
-- [X]   Trace logging
-- Font not found: fixed:pixelsize=10
-- Font not found: unifont:fontformat=truetype
-- Font not found: siji:pixelsize=10
-- Found PythonInterp: /usr/bin/python2.7 (found suitable version "2.7.12", minimum required is "2.7") 
-- XCB[XCB]: Found component XCB
-- XCB[ICCCM]: Found component ICCCM
-- XCB[EWMH]: Found component EWMH
-- XCB[UTIL]: Found component UTIL
-- XCB[IMAGE]: Found component IMAGE
-- Found XCB: /usr/lib/x86_64-linux-gnu/libxcb.so;/usr/lib/x86_64-linux-gnu/libxcb-icccm.so;/usr/lib/x86_64-linux-gnu/libxcb-ewmh.so;/usr/lib/x86_64-linux-gnu/libxcb-util.so;/usr/lib/x86_64-linux-gnu/libxcb-image.so  
-- Searching for xcbgen with python2
-- Found xcbgen in /usr/lib/python2.7/dist-packages
-- XCB[RANDR]: Found component RANDR
-- Found XCB: /usr/lib/x86_64-linux-gnu/libxcb.so;/usr/lib/x86_64-linux-gnu/libxcb-icccm.so;/usr/lib/x86_64-linux-gnu/libxcb-ewmh.so;/usr/lib/x86_64-linux-gnu/libxcb-util.so;/usr/lib/x86_64-linux-gnu/libxcb-image.so;/usr/lib/x86_64-linux-gnu/libxcb-randr.so  
-- XCB[COMPOSITE]: Found component COMPOSITE
-- Found XCB: /usr/lib/x86_64-linux-gnu/libxcb.so;/usr/lib/x86_64-linux-gnu/libxcb-icccm.so;/usr/lib/x86_64-linux-gnu/libxcb-ewmh.so;/usr/lib/x86_64-linux-gnu/libxcb-util.so;/usr/lib/x86_64-linux-gnu/libxcb-image.so;/usr/lib/x86_64-linux-gnu/libxcb-randr.so;/usr/lib/x86_64-linux-gnu/libxcb-composite.so  
-- xpp: including xcb proto randr.xml
-- xpp: including xcb proto composite.xml
-- xpp: including xcb proto xproto.xml
-- Checking for module 'jsoncpp>=1.7.7'
--   
-- Could not find jsoncpp>=1.7.7 -- building static lib...
-- JsonCpp Version: 1.7.7
-- Found PythonInterp: /usr/bin/python2.7 (found suitable version "2.7.12", minimum required is "2.6") 
fatal: Not a git repository (or any of the parent directories): .git
-- Configuring done
-- Generating done
-- Build files have been written to: /home/thewithz/lib/polybar/build

```

## Environment:
* Distro (with Version):
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.5 LTS
Release:	16.04
Codename:	xenial

## Additional context
*Add any other context that you think is necessary about the problem here.*
libjsoncpp1 is already the newest version (1.7.2-1).

patrick96:
This sets up documentation inside the polybar repo. I plan to also host the html documentation with [readthedocs](https://readthedocs.org/), which offers easy deployment for sphinx documentation.

Please read `doc/Readme.md` for how to look at the documentation locally.

I have also rewritten the manpage in reStructuredText because it has a nicer syntax and we can also generate a nice looking html page from it so we can have it accessible on the web:
2018-12-07-235514_758x881_scrot

The whole idea behind having the documentation inside the repo instead of the wiki or polybar.github.io is so that you can look at the documentation for the polybar version you are using. readthedocs allows to choose the version to look at in a switcher and when you build the documentation locally you get the one matching your version anyway. This allows us to bring the documentation up-to-date together with a PR instead of adding it to the wiki afterwards and marking it with `(unreleased)` and then removing that when it makes it into a release (dealing with deprecated features is even more of a pain).

The documentation will be written in reStructuredText because it's more expressive and more standardized than markdown. Readthedocs also discourages markdown (see [here](http://www.ericholscher.com/blog/2016/mar/15/dont-use-markdown-for-technical-docs/)). And it will use `python-sphinx` to generate html documentation and the manpages.

This PR additionally moves the handling of the version string to cmake. Before `settings.hpp` determined the effective version string depending on what macros were set. Now cmake determines the effective version string and uses `configure_file` to configure all the places that need it.

EDIT:
Fixes #1611 

## Was at the top before:
@NBonaparte This is still WIP, so no need to thoroughly review this. But I want to discuss something with you. How should we get the polybar version into the documentation during building?

When building polybar the version string needs to be put in a few places:
* source code (for `polybar -v`)
* manpages
* html documentation pages

With this PR, sphinx will handle putting the version string into the documentation and the manpages, we just need to set a `version` field in `conf.py`, which is a python script, so we're quite flexible there.

Right now we use cmake to configure files and put the version string into them. This, however, doesn't work well if polybar isn't build from a git repo (e.g. release tarball), there cmake cannot gather the current git tag. Ideally we want a uniform way to get the version string into all the places it needs to be. 
I have thought about using a shell script that figures out the actual version string (using git if possible or else using hardcoded version numbers that we update each release) and then somehow passes them to cmake and sphinx.
But I'm not quite sure how to approach this. I don't want to change the build process for the user and sourcing a script within cmake doesn't work (because that's not how cmake works), so there's no way to get the version string from a bash script into cmake during configuration.

Do you have an idea how we could tackle this? This is the major thing blocking me from finishing this PR.


patrick96:
And then configuring `conf.py` with cmake? That would work. We would then have to run `cmake` before being able to build the documentation, but I guess that's fine. I'll see how I can implement this cleanly.

patrick96:
Sorry for the force-push spam, I was trying to get travis to also build the documentation but decided to do that in a separate PR because it requires us to update the distro travis uses.

@NBonaparte I have rewritten most of the version string handling and I think it's ready to review now. The version that is hardcoded into `CMakeLists.txt` is `3.3.0`, so ideally this is merged after #1543.

patrick96:
I have added the documentation to [readthedocs.org](https://readthedocs.org/). You can have a look at what this branch looks like [here](https://polybar-patrick.readthedocs.io).

If this PR is getting too big, let me know and I can split it up in multiple smaller ones.

patrick96:
Hey @NBonaparte could you review this, so that we can start moving the documentation over and start writing new documentation in this format.

patrick96:
Fortunately it's only a `makedepend`, also sphinx is used by a ton of projects, it's likely that AUR users already have it installed (or at least cached).

I was trying to create a separate travis job to build the documentation but without requiring it to get all the polybar dependencies. I decided to do that in a separate PR after this is merged. I may make the `doc` folder its own cmake project so that we can configure it separately. I also explicitly disabled building the docs in all other builds because each job should check only one thing.

CMakeLists.txt
```diff
@@ -22,6 +22,29 @@ endif()
 
 project(polybar C CXX)
 
+# Polybar version information
+# Update this on every release
+# This is used to create the version string if a git repo is not available
+set(VERSION_MAJOR "3")
+set(VERSION_MINOR "3")
+set(VERSION_PATCH "0")
```
NBonaparte: Instead of writing this out manually, we could do some parsing of `version.hpp` like
```
file(READ "include/version.hpp" ver)
string(REGEX MATCH "...")
```
similar to [this](https://stackoverflow.com/a/47084079).

patrick96: We actually remove the `version.hpp` in this PR.

Wouldn't parsing just create more problems? Or is there some benefit to your suggestion that I'm not seeing?

NBonaparte: I prefer having a separate file just for versioning to make things simpler to read/find, but it shouldn't matter too much.

patrick96: Fair enough, shouldn't be too difficult to add a `version.txt` or something like that.

patrick96: Done

PR #1546, opened 2018-12-15
Add label minlen and alignment.

infokiller:


src/drawtypes/label.cpp
```diff
@@ -8,7 +9,27 @@ POLYBAR_NS
...
+    }
+    const size_t num_fill_chars = m_minlen - len;
+    string aligned_label;
+    if (m_alignment == label_alignment::RIGHT) {
+      aligned_label = string(num_fill_chars, ' ') + m_tokenized;
+    } else if (m_alignment == label_alignment::LEFT) {
+      aligned_label = m_tokenized + string(num_fill_chars, ' ');
+    } else {
+      const string sided_fill_chars(std::ceil(num_fill_chars / 2.0), ' ');
+      aligned_label = sided_fill_chars + m_tokenized + sided_fill_chars;
```
patrick96: for an uneven `num_fill_chars`, this would create a label that has length `m_minlen + 1`

src/drawtypes/label.cpp
```diff
@@ -216,15 +237,30 @@ namespace drawtypes {
...
+      alignment = label_alignment::CENTER;
+    } else {
+      throw application_error(sstream() << "Label " << section << "." << name << " has invalid alignment "
+                                        << alignment_conf_value << ", expecting one of: right, left, center.");
+    }
 
     size_t maxlen = conf.get(section, name + "-maxlen", 0_z);
+    if (maxlen > 0 && maxlen < minlen) {
+      throw application_error(sstream() << "Label " << section << "." << name << " has maxlen " << maxlen
+                                        << " which is smaller than min_len " << minlen);
```
patrick96:
```suggestion
                                        << " which is smaller than minlen " << minlen);
```

include/drawtypes/label.hpp
```diff
@@ -31,16 +31,25 @@ namespace drawtypes {
   using icon = label;
   using icon_t = label_t;
 
+  enum class label_alignment {
```
patrick96: Please use `alignment` from `types.hpp`.

src/drawtypes/label.cpp
infokiller: Yes, I wasn't sure if I should round up or down in order to respect the center alignment.
If I round down, that means the length will be `m_minlen - 1`. Do you prefer that?

include/drawtypes/label.hpp
infokiller: Done.

infokiller:
Thanks for the good review!
Added unit tests.

src/drawtypes/label.cpp
```diff
@@ -8,7 +9,27 @@ POLYBAR_NS
...
-    return m_tokenized;
+    const size_t len = string_util::char_len(m_tokenized);
+    if (len >= m_minlen) {
+      if (m_maxlen <= 0 || len <= m_maxlen) {
+        return m_tokenized;
+      }
+      if (!m_ellipsis) {
+        return string_util::utf8_truncate(string(m_tokenized), m_maxlen);
+      }
+      return string_util::utf8_truncate(string(m_tokenized), m_maxlen - 3) + "...";
```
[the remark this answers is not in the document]
infokiller: Done.

src/drawtypes/label.cpp
patrick96: I think the following should *always* hold:
```
minlen <= len(effective label) <= maxlen
```
and if `len(m_tokenized) < minlen`, `len(effective label) == minlen`

In this case this would mean that we fill one additional space character, either on the left or on the right when `num_fill_chars` is odd. I would suggest adding it to the left side.
This could be generalized to say that the left side gets filled with `std::ceil(num_fill_chars / 2.0)` and the right side with `std::floor(num_fill_chars / 2.0)` spaces. If `num_fill_chars` is even those numbers are equal and if it's odd those numbers added together give you `num_fill_chars` again.

infokiller: > I think the following should _always_ hold:
> 
> ```
> minlen <= len(effective label) <= maxlen
> ```
I agree.
> and if `len(m_tokenized) < minlen`, `len(effective label) == minlen`
> 
I think this is more debatable. The main issue I see is that even though a user requested center alignment, they will not get it. At least in my case, I would be more surprised to see non-centered alignment than `len(m_tokenized) == minlen + 1`, assuming `minlen + 1 <= maxlen`.
Also, doing it this way means that the user can't really guarantee center alignment, since they can't restrict the labels to an even/odd length (only restrict the min/max length).
> In this case this would mean that we fill one additional space character, either on the left or on the right when `num_fill_chars` is odd. I would suggest adding it to the left side.
> This could be generalized to say that the left side gets filled with `std::ceil(num_fill_chars / 2.0)` and the right side with `std::floor(num_fill_chars / 2.0)` spaces. If `num_fill_chars` is even those numbers are equal and if it's odd those numbers added together give you `num_fill_chars` again.

patrick96: > The main issue I see is that even though a user requested center alignment, they will not get it.

Fair point. I guess as long as the case where maxlen == minlen or maxlen == minlen + 1 is handled correctly, we could do it this way.

src/drawtypes/label.cpp
```diff
@@ -216,15 +231,30 @@ namespace drawtypes {
...
+                                        << " which is smaller than minlen " << minlen);
+    }
     bool ellipsis = conf.get(section, name + "-ellipsis", true);
 
-    if(ellipsis && maxlen > 0 && maxlen < 3) {
-      throw application_error(sstream()
-          << "Label " << section << "." << name
-          << " has maxlen " << maxlen
-          << ", which is smaller than length of ellipsis (3)");
+    if (ellipsis && maxlen > 0 && maxlen < 3) {
```
patrick96: Undo this. We generally try to keep our lines under 80 characters (though 120 is the hard limit). There's no need to introduce extra changes if the code doesn't actually change.

src/drawtypes/label.cpp
patrick96: Just did some testing and minlen with center alignment causes some trouble.
In particular a label with text `abc`, `minlen = 4`, `maxlen = 4`, center alignmend, `maxlen = 4` and ellipsis enabled will produce ` ...` instead of `abc ` or ` abc`. I would suggest that in this special case the extra space is located at the end.

I would also suggest that we merge the functionality of this method and of `builder::get_label_text` to handle minlen, maxlen and ellipsis all in one function. Like this it's also easier to test this and how minlen and maxlen interact. We should probably move everything into `label::get` because that function is called in several places and should return the effective label in all those cases.

src/drawtypes/label.cpp
infokiller: Done

src/drawtypes/label.cpp
infokiller: Done.

infokiller:
Apologies for the huge delay! I promise to respond fast this time :)
I've rebased the commits on top of the up to date master since a lot of time has passed.
I've also addressed the remaining comments.

infokiller:
Is there anything I can do to get this PR merged?
Thanks!

infokiller:
Thanks for the quick response!
I formatted the files with clang-format and pushed a new commit.

include/drawtypes/label.hpp
```diff
@@ -36,26 +37,31 @@ namespace drawtypes {
      * labels in a different way.
      */
     size_t m_maxlen{0_z};
+    // The default is set to RIGHT for backwards compatibility.
+    alignment m_alignment{alignment::RIGHT};
```
patrick96: We didn't have minlen before, why do we need to be backwards compatible here?

Am I forgetting something?

infokiller:
Thanks a lot for the review and cleanup!
Looking at the comment "The default is set to RIGHT for backwards compatibility.", I actually don't remember why I wrote it, and it doesn't make any sense, because alignment is only relevant when there is a minimum length, and before this PR there was no minimum length...
So, should I set the default alignment to left?

patrick96:
The only place I could imagine this coming from is the minlen property on tokens where they get right aligned. That property was mainly intended for numbers I think where right alignment makes more sense (the ones, tens, hundreds, etc. always stay in the same place). But for labels left alignment makes more sense I think, so changing it to left would be better.

infokiller:
Yes, I agree. I changed it to left.

include/drawtypes/label.hpp
```diff
@@ -36,26 +37,30 @@ namespace drawtypes {
...
+    alignment m_alignment{alignment::LEFT};
     bool m_ellipsis{true};
 
     explicit label(string text, int font) : m_font(font), m_text(text), m_tokenized(m_text) {}
     explicit label(string text, string foreground = ""s, string background = ""s, string underline = ""s,
-        string overline = ""s, int font = 0, struct side_values padding = {0U,0U}, struct side_values margin = {0U,0U},
-        size_t maxlen = 0_z, bool ellipsis = true, vector<token>&& tokens = {})
+        string overline = ""s, int font = 0, struct side_values padding = {0U, 0U},
+        struct side_values margin = {0U, 0U}, int minlen = 0, size_t maxlen = 0_z,
+        alignment label_alignment = alignment::RIGHT, bool ellipsis = true, vector<token>&& tokens = {})
```
patrick96:
```suggestion
        alignment label_alignment = alignment::LEFT, bool ellipsis = true, vector<token>&& tokens = {})
```

src/drawtypes/label.cpp
```diff
@@ -216,15 +251,30 @@ namespace drawtypes {
         token.suffix = token_str.substr(pos + 1, token_str.size() - pos - 2);
       }
     }
+    size_t minlen = conf.get(section, name + "-minlen", 0_z);
+    string alignment_conf_value = conf.get(section, name + "-alignment", "right"s);
```
patrick96:
```suggestion
    string alignment_conf_value = conf.get(section, name + "-alignment", "left"s);
```

infokiller:
Sorry I missed that, and thanks for fixing it.

PR #1550, opened 2018-12-17
i3: add workspaces-max-{count,width} options.

infokiller:


patrick96:
Review will have to wait until after Christmas, busy right now.

Could you please briefly explain the what and why of this PR?

infokiller:
No worries!
What: be able to limit the number and width of the i3 workspaces shown in the bar.
Why: when you have many i3 workspaces, you don't always want all of them to be displayed, since it can be distracting (showing too much irrelevant info) and can push other bar elements (for example, if you have something in the center).

Issue #1462, opened 2018-10-12
Xworkspaces module creates floating point exception on 2BWM


abraxous:
Polybar can be used without the xworkspaces module on 2BWM, or with it on i3 (I have not tested other WMs). When the xworkspaces module is used in 2BWM, polybar fails with message "Floating Point Exception". I suspect this is a result of _NET_DESKTOP_NAMES not being properly handled by 2BWM. I didn't see the floating point exception mentioned in the other issue request mentioning _NET_DESKTOP_NAMES, however, so I decided it would be worth putting in an issue for documentation's sake at the very least.

Here's a sample config for testing

*A minimal but complete config with which the problem occurs:*
```dosini
[colors]
;background = ${xrdb:color0:#222}
;background = #222
;background-alt = #444
;foreground = ${xrdb:color7:#222}
;foreground = #dfdfdf
;foreground-alt = #555
;primary = #ffb52a
;secondary = #e60053
;alert = #bd2c40
background = ${xrdb:color0:#222}
background-alt = ${xrdb:color0:#222}
foreground = ${xrdb:color7:#222}
foreground-alt = ${xrdb:color7:#222}
primary = ${xrdb:color1:#222}
secondary = ${xrdb:color2:#222}
alert = ${xrdb:color3:#222}

[module/ewmh]
type = internal/xworkspaces

[bar/test]
bottom = false
dock = false
width = 100%
height = 25
offset-x = 0
offset-y = 0
override-redirect = true

background = ${colors.background}
foreground = ${colors.foreground}

line-size = 3
line-color = #f00

border-left-size = 10
border-right-size = 10
border-top-size = 10
border-bottom-size = 0
border-color = #00000000

padding-left = 2
padding-right = 2

module-margin-left = 2
module-margin-right = 2

font-0 = fixed:pixelsize=10;1
font-1 = unicode:fontformat=truetype:size=8:antialias=false;0
font-2 = "Wuncon siji:pixelsize=10;1"

modules-left =
modules-center = ewmh
modules-right =
tray-position = right
tray-padding = 2

cursor-click = pointer
```

patrick96:
Can confirm.

SIGFPE happens here: https://github.com/jaagr/polybar/blob/master/src/modules/xworkspaces.cpp#L166

[...]

PR #1579, opened 2019-01-03
xworkspace support window manager with less names than desktops

### LINKED ISSUE #248: Provide default desktop names

### LINKED ISSUE #1491: module/ewmh does not detect removed workspaces in openbox
## Describe the issue
Removed workspaces in openbox are not removed from the workspace list in the ewmh module.

**Expected behavior:**
When workspaces are removed in the WM, it should also be removed from the list.

**Actual behavior:**

When workspaces are added in the WM, they are added to the list. When they are removed in the WM, they stay in the list. 

## To Reproduce
```dosini
[bar/example]
width = 100%
height = 40
offset-x = 0
offset-y = 0
radius = 0
fixed-center = false
underline-size = 0

modules-left = ewmh 
modules-center = 
modules-right = 

[module/ewmh]
type = internal/xworkspaces

pin-workspaces = true

enable-click = true 

enable-scroll = false
```
*List any other steps needed to reproduce the issue besides starting polybar with the config you posted above.*

## Polybar Log
```

```

## Screenshots
Default workspaces:

After adding a workspace

After removing the added workspace

## Environment:
* WM: Openbox
* Output of `polybar -vvv`:
```
polybar 3.1.0-git

Features: +alsa +curl -i3 -mpd +network

X extensions: +randr (+monitors) -render -damage -sync -composite +xkb +xrm +xcursor

Build type: RelWithDebInfo
Compiler: /usr/bin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -Wall -Wextra -Werror -Wno-noexcept-type -O2 -pedantic -pedantic-errors
```
## Additional context
It doesn't matter if the workspace is added with ObConf, with the middleclick menu or using `wmctrl -n 5`

frstrikerman:
TEAMEDIT:
Fixes #1491 
Fixes #248
Closes #904

### DIFF
--- a/include/x11/ewmh.hpp
+++ b/include/x11/ewmh.hpp
@@ -24,6 +24,7 @@ namespace ewmh_util {
   vector<position> get_desktop_viewports(int screen = 0);
   vector<string> get_desktop_names(int screen = 0);
   unsigned int get_current_desktop(int screen = 0);
+  unsigned int get_number_of_desktops(int screen = 0);
   xcb_window_t get_active_window(int screen = 0);
 
   void change_current_desktop(unsigned int desktop);
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -97,7 +97,7 @@ namespace modules {
   void xworkspaces_module::handle(const evt::property_notify& evt) {
     if (evt->atom == m_ewmh->_NET_CLIENT_LIST) {
       rebuild_clientlist();
-    } else if (evt->atom == m_ewmh->_NET_DESKTOP_NAMES) {
+    } else if (evt->atom == m_ewmh->_NET_DESKTOP_NAMES || evt->atom == m_ewmh->_NET_NUMBER_OF_DESKTOPS) {
       m_desktop_names = get_desktop_names();
       rebuild_desktops();
       rebuild_desktop_states();
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -224,7 +224,7 @@ namespace modules {
     unsigned int desktops_number = ewmh_util::get_number_of_desktops();
     if(desktops_number == names.size())
       return names;
-    else if(desktops_number < names.size()){
+    else if(desktops_number < names.size()) {
       names.erase(names.begin()+desktops_number, names.end());
       return names;
     }
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -81,7 +81,7 @@ namespace modules {
     m_monitors = randr_util::get_monitors(m_connection, m_connection.root(), false);
 
     // Get desktop details
-    m_desktop_names = ewmh_util::get_desktop_names();
+    m_desktop_names = get_desktop_names();
     m_current_desktop = ewmh_util::get_current_desktop();
 
     rebuild_desktops();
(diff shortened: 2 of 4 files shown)


frstrikerman:
I made those change while this issue wasn't fixed : https://github.com/jaagr/polybar/issues/1462
It might be irrelevant now

patrick96:
Awesome, so this basically makes the xworkspaces module conformant to [this](https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472706208) section of the ewmh spec?

frstrikerman:
Yes

patrick96:
Alright it's on my todo list. Might be a while though.

frstrikerman:
I just checked and one part of the spec is missing when there is more names than dekstops. I will try to implement it during this week

patrick96:
Great. Thanks a lot for helping out!

src/x11/ewmh.cpp
```diff
@@ -77,6 +77,13 @@ namespace ewmh_util {
     return desktop;
   }
 
+  unsigned int get_number_of_desktops(int screen) {
+    auto conn = initialize().get();
+    unsigned int desktops = XCB_NONE;
```
patrick96:
```suggestion
    uint32_t desktops = 0;
```

src/x11/ewmh.cpp
```diff
@@ -77,6 +77,13 @@ namespace ewmh_util {
     return desktop;
   }
 
+  unsigned int get_number_of_desktops(int screen) {
```
patrick96: Just to make it consistent with what `xcb_ewmh.h` uses.

```suggestion
  uint32_t get_number_of_desktops(int screen) {
```

src/modules/xworkspaces.cpp
```diff
@@ -219,6 +219,21 @@ namespace modules {
     }
   }
 
+  vector<string> xworkspaces_module::get_desktop_names(){
+    vector<string> names = ewmh_util::get_desktop_names();
+    unsigned int desktops_number = ewmh_util::get_number_of_desktops();
```
patrick96:
```suggestion
    uint32_t desktops_number = ewmh_util::get_number_of_desktops();
```

include/x11/ewmh.hpp
```diff
@@ -24,6 +24,7 @@ namespace ewmh_util {
   vector<position> get_desktop_viewports(int screen = 0);
   vector<string> get_desktop_names(int screen = 0);
   unsigned int get_current_desktop(int screen = 0);
+  unsigned int get_number_of_desktops(int screen = 0);
```
patrick96:
```suggestion
  uint32_t get_number_of_desktops(int screen = 0);
```

patrick96:
Oh and I forgot to mention this, please always use curly braces for if-blocks, even id they're only a single line

frstrikerman:
I wasn't able to work on this because I had exams

src/x11/ewmh.cpp
patrick96: Fair enough, we don't really need to do this here. I added an entry in our refactoring project to use fixed width integers for everything that interacts with EWMH since the EWMH spec defines the bit-width of the numbers.

PR #1590, opened 2019-01-08
[WIP] fix(pulse): Handle daemon restart

### LINKED ISSUE #1580: Pulseaudio module interaction causes crash
## Describe the issue
If pulseaudio is killed while polybar is already running any future mouse interaction with the pulseaudio module causes a crash of polybar. 

## To Reproduce
I can reproduce with the stock config with everything but pulseaudio and the example bar removed.
- `polybar example`
- `pulseaudio -k`
- Click or scroll on the pulseaudio module

Config:
```
[colors]
background = #222
background-alt = #444
foreground = #dfdfdf
foreground-alt = #555
primary = #ffb52a
secondary = #e60053
alert = #bd2c40

[bar/example]
width = 100%
height = 27
radius = 6.0
fixed-center = false

background = ${colors.background}
foreground = ${colors.foreground}

line-size = 3
line-color = #f00

border-size = 4
border-color = #00000000

padding-left = 0
padding-right = 2

module-margin-left = 1
module-margin-right = 2

font-0 = misc fixed:pixelsize=10;1
font-1 = unifont:fontformat=truetype:size=8:antialias=false;0
font-2 = wuncon siji:pixelsize=10;1

modules-right = pulseaudio

tray-position = right
tray-padding = 2

cursor-click = pointer
cursor-scroll = ns-resize

[module/pulseaudio]
type = internal/pulseaudio

format-volume = <label-volume> <bar-volume>
label-volume = VOL %percentage%%
label-volume-foreground = ${root.foreground}

label-muted = 🔇 muted
label-muted-foreground = #666

bar-volume-width = 10
bar-volume-foreground-0 = #55aa55
bar-volume-foreground-1 = #55aa55
bar-volume-foreground-2 = #55aa55
bar-volume-foreground-3 = #55aa55
bar-volume-foreground-4 = #55aa55
bar-volume-foreground-5 = #f5a70a
bar-volume-foreground-6 = #ff5555
bar-volume-gradient = false
bar-volume-indicator = |
bar-volume-indicator-font = 2
bar-volume-fill = ─
bar-volume-fill-font = 2
bar-volume-empty = ─
bar-volume-empty-font = 2
bar-volume-empty-foreground = ${colors.foreground-alt}
```

## Polybar Log
```
warn: No monitor specified, using "DP-2"
warn: pulseaudio: using default sink alsa_output.pci-0000_00_1f.3.analog-stereo
error: tray: Failed to put tray above 0x0c00001 in the stack (XCB_MATCH (8))
Assertion 'o' failed at pulse/operation.c:133, function pa_operation_get_state(). Aborting.
Aborted (core dumped)
```

## Environment:
* WM: i3-gaps
* Output of `polybar -vvv`:
```
polybar 3.3.0

Build type: RelWithDebInfo
Compiler: /usr/lib/ccache/bin/c++
Compiler flags: -march=native -O2 -pipe -fstack-protector-strong -fno-plt -Wall -Wextra -Werror -Wno-noexcept-type -pedantic -pedantic-errors
```

patrick96:
Whenever the pulseaudio daemon dies all returned operations from `pa_*` functions are `NULL`, this causes a few failed assertions. I was able to fix this quite easily by just throwing exceptions if the operations are NULL.

However, polybar doesn't reconnect to pulse when the daemon dies. I had some problems with implementing that and could use your help @NBonaparte since you implemented this module. What I did is basically to move the constructor and destructor functionality into their own functions so that we can reconnect to pulse in `context_state_callback`. This doesn't work though and we get
```
Assertion '!in_worker(m)' failed at pulse/thread-mainloop.c:165, function pa_threaded_mainloop_stop(). Aborting.
``` 
when polybar loses connection. This assertion fails when calling `pa_threaded_mainloop_stop`. It makes sense why this is happening, since `context_state_callback` is static and likely some function in the adapter running concurrently has locked the mainloop. I'm not sure how I should handle this. Maybe we could let functions only lock the mainloop when the adapter isn't about the reconnect? Because otherwise, how can you make sure that no one has locked the mainloop before stopping it? Maybe you have some insight here.

Fixes #1580


NBonaparte:
IIRC, callbacks are called by the external (hidden) pulseaudio main loop worker thread, which is what `in_worker(m)` probably checks for. Since it is actually in it, the assertion fails.

Therefore, handling `PA_CONTEXT_THREAD` should be done in the module thread (possibly in `process_events()`, if you create a new enum type for this case).

patrick96:
Alright, so I was able to fix that with your suggestion.

But it raised a host of other issues. For example, if an event was processed after the daemon died but before the adapter handled that, the generated exception killed the entire bar. Also when connecting/reconnecting when there is no daemon around also caused the bar to lock up.

I think I got them all, but could you please be thorough when reviewing this, It's possible that I missed something and you have more experience in this area than me.

patrick96:
Fixed the compilation errors in older compilers

NBonaparte:
Do we want to handle an intermittent period where no daemon is running? This breaks when killing the daemon without a restart.

patrick96:
We probably should. I'll take a look at it later.

Lomadriel:
I fixed some concurrency issues in this adapter and tried to fix some other issues like memory leaks when some exception are thrown. See https://github.com/Lomadriel/polybar/tree/issue/1580

Lomadriel:
I just added a commit on my branch that fix the bug mentioned in my last post.
I added a delay before trying a re-connection. I took the inspiration from pavucontrol (https://github.com/pulseaudio/pavucontrol/blob/aa51a92b109d143d3d26c608f57d40d83a843ef4/src/pavucontrol.cc#L620).
However, if the connection can't be established, the module will still crash. It may be better to retry X times or indefinitely every N seconds.

hicklemon:
@NBonaparte There are three issues open that this may fix:

#1580 #2170 #2408

Any chance you could review?

PR #1596, opened 2019-01-12
Clean Up handling of raw tags

### LINKED ISSUE #1555: Tag parsing not taking multiple tags in the same block
Hi, when setting the foreground / underline colors with tags in ramp-capacity, I noticed the underline and text color goes unchanged until something else (in this case, my exit module) changes it. 
Do you have any idea of is this is normal / a bug ? 

**Normal functionment (without Tag editing):**
image
**Bugged functionment (with tag editing):**
image

```
[module/battery]
type = internal/battery
full-at = 95
battery = BAT0
adapter = AC
poll-interval = 5
format-charging = <animation-charging> <label-charging>
format-charging-foreground = ${colors.green}
format-charging-underline = ${colors.green}
format-discharging = <ramp-capacity> <label-discharging>
label-charging = "%percentage%%  "
label-discharging = "%percentage:2%%  "
label-full = 
ramp-capacity-0 ="%{F#bf616a u#bf616a +u}   %{-u}"
ramp-capacity-1 ="%{F#d08770 u#d08770 +u}   %{-u}"
ramp-capacity-2 ="%{F#ebcb8b u#ebcb8b +u}   %{-u}"
ramp-capacity-3 ="%{F#a3be8c u#a3be8c +u}   %{-u}"
ramp-capacity-4 ="%{F#a3be8c u#a3be8c +u}   %{-u}"
animation-charging-0 = "   "
animation-charging-1 = "   "
animation-charging-2 = "   "
animation-charging-3 = "   "
animation-charging-4 = "   "
animation-charging-framerate = 500
```

## Environment:
* WM: i3-gaps
* Output of `polybar -vvv`:
```polybar 3.2.0-73-gbc9eddf

Features: +alsa +curl +i3 +mpd -network(wireless-tools) +pulseaudio +xkeyboard

Build type: RelWithDebInfo
Compiler: /usr/bin/clang++
Compiler flags:  -Wall -Wextra -Werror -pedantic -pedantic-errors -Wno-error=parentheses-equality -Wno-zero-length-array
Linker flags: ```

### PR DESCRIPTION
**Pull request by patrick96**
Until now the builder parsed any raw tags that users provided (like ones coming from script modules) and used it to track how many tags were open, to be able to close it when flushing the builder.

This method didn't support multiple formatting tags in a single block (e.g. `%{F#ff0000 B-}`) and thus created some issues (see #1555). Instead of fixing this, I just completely removed the tracking of raw tags. Because we would then have basically replicated `parser.cpp` in the builder.

Because we don't track raw tags anymore they could leak out and affect the modules appearing to the right of it. For that I introduced a new formatting tag `P`, which is used for polybar control commands, `%{PR}` is the only available tag right now, and it resets all colors, activations of underline and overline and the font in the renderer (it's basically equivalent to `%{F- B- T- -u -o u- o-}` but less verbose).

This will probably change the behavior of some configs that use raw tags outside of the script module. 
This PR only causes raw tags to be reset later and never earlier. So colors leaking out could happen more, but tags won't be reset earlier than they currently are.
Some drawtypes would call flush at (for the user) arbitrary places, users could have relied on those flushes to reset their raw tags they used inside the labels and because now this doesn't happen anymore those raw tags could leak over the subsequent labels in the module. I don't think this is a big deal as we never gave any guarantees when polybar would reset tags within a module (except that labels won't reset tags that they didn't set at the beginning).

Fixes #1555

**Changelog Note:** Possibly breaks behavior.

### CODE REVIEW

File: src/components/renderer.cpp [hunk 1]
```diff
@@ -827,4 +828,24 @@ bool renderer::on(const signals::parser::text& evt) {
   return true;
 }
 
+bool renderer::on(const signals::parser::control& evt) {
+  auto ctrl = evt.cast();
+
+  switch(ctrl) {
```
Lomadriel:
```suggestion
  switch (ctrl) {
```

patrick96:
> Maybe we should add something into the parser to forbid the usage of the %{PX} tags to the users?

The point of this PR kind of is that we don't have to do tag parsing before sending the entire string to `parser.cpp`. I think if we don't add user documentation for it, it should be fine.

PR #1619, opened 2019-01-23
Networkclick

schlerp:
First time submitting a PR for a non-python project, hopefully its not too sloppy.

Basically I wanted to be able to launch wicd-curses when I click on the network so I took the code from text.cpp and added it to network.cpp to allow configuration of click events in the config file. I then wanted the same thing for the cpu.cpp module so I could launch htop. Both appear to work well from my local testing and have added a ton of useful functionality to my polybar setup.

Ideally I wanted this to add the click events to the prefix (which I use mainly for icons in my polybar setup), however I couldnt get that to work. Happy to have another crack at it though if someone could point me in the right direction.

### DIFF
--- a/include/modules/cpu.hpp
+++ b/include/modules/cpu.hpp
@@ -2,6 +2,7 @@
 
 #include "settings.hpp"
 #include "modules/meta/timer_module.hpp"
+#include "utils/command.hpp"
 
 POLYBAR_NS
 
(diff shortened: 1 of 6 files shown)


schlerp:
Hey Patrick, 

I have added the changes you requested except for the part about get_output.

I have done some more local testing and it works well in this state, let me know if there is anything else i should chage

patrick96:
Closing due to inactivity

Feel free to reopen this if you still want to work on it.

Issue #1628, opened 2019-01-25
Linking problem undef reference 'xcb_prefetch_maximum_request_length'


jubalh:
It seems when I compile polybar with `CMAKE_EXE_LINKER_FLAGS="-Wl,--as-needed -Wl,--no-undefined -Wl,-z,now"`
The build aborts with:

```
libpoly.a(bar.cpp.o): undefined reference to symbol 'xcb_prefetch_maximum_request_length'
```

## Build Process

Using 3.3.0 [tarball](https://github.com/jaagr/polybar/releases/download/3.3.0/polybar-3.3.0.tar).

We have a general cmake macro which expands to:
```
    find . -name CMakeLists.txt \
        -exec sed -i -re '/^[[:blank:]]*[sS][eE][tT][[:blank:]]*\([[:blank:]]*(CMAKE_BUILD_TYPE|CMAKE_COLOR_MAKEFILE|CMAKE_INSTALL_PREFIX|CMAKE_VERBOSE_MAKEFILE).*\)/{s/^/#IGNORE /}' {} + 
    mkdir -p build 
    cd build 
    /usr/bin/cmake $OLDPWD/. \
%if "/usr/bin/make" == "/usr/bin/make" 
        -G"Unix Makefiles" \
%else 
        -GNinja \
%endif 
        -DCMAKE_INSTALL_PREFIX:PATH=/usr \
        -DINCLUDE_INSTALL_DIR:PATH=/usr/include \
        -DLIB_INSTALL_DIR:PATH=/usr/lib64 \
        -DSYSCONF_INSTALL_DIR:PATH=/etc \
        -DSHARE_INSTALL_PREFIX:PATH=/usr/share \
        -DCMAKE_INSTALL_LIBDIR:PATH=/usr/lib64 \
        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
        -DCMAKE_C_FLAGS="${CFLAGS:--O2 -g -m64 -fmessage-length=0 -D_FORTIFY_SOURCE=2 -fstack-protector -funwind-tables -fasynchronous-unwind-tables} -DNDEBUG" \
        -DCMAKE_CXX_FLAGS="${CXXFLAGS:--O2 -g -m64 -fmessage-length=0 -D_FORTIFY_SOURCE=2 -fstack-protector -funwind-tables -fasynchronous-unwind-tables} -DNDEBUG" \
[...]

PR #1633, opened 2019-01-28
Contrib: Add spec file

jubalh:


patrick96:
We don't take build files into the repository if we don't maintain them ourselves. Because having to keep the build files up to date with wherever they are actually maintained is quite a hassle.

If you want this repo to be the place where the build files are stored and updated and you push them to OBS whenever they are changed here (like we do with the AUR), then we could do that. Just be aware that any changes to the build files will go through the PR process and that may be a little slower than when you have full control over the repo where the spec file is stored.

patrick96:
Hmm, what would be the advantage of that? For me it seems like we would only add the spec file for the sake of having it, or am I missing something?

jubalh:
Yes, that's true. Because when I created the PR I didn't know that you only have the aur added because you maintain it. I thought you try to have the recipes for the various package managers in one place.

Since rpm can be used on more systems than just openSUSE, I thought when I contribute the spec file here, people of other rpm based distro (Fedora,..) can easily see it in your repo and can easier create their packages.

If however this is not the intended purpose we can also close this PR.

patrick96:
Ah, I didn't know that the same spec file could be used across multiple distros. In this case, I think it would fit into the repo.

Is there some distro independent way to build spec files that we could add to the wiki as documentation?

contrib/rpm/polybar.spec
```diff
@@ -0,0 +1,70 @@
...
+
+%description
+A fast and easy-to-use status bar for tilling WM
+
+%prep
+%setup -q
+
+%build
+%cmake
+make userconfig
```
patrick96: This puts a file into the user's home directory (and overwrites existing ones). I don't know if that belongs into a package build file.

jubalh: Thanks for pointing this out. In our case this actually does nothing then, because the OBS is a clean build environment. Changes to a users home are just ignored in this scenario. It seems sysek was not aware of this, I'll update this PR and also the internal package.

jubalh:
> Ah, I didn't know that the same spec file could be used across multiple distros. In this case, I think it would fit into the repo.

It depends if they use the same names for the packages so that dependencies are resolved correctly. Above we try to use pkgconfig most of the time so this is good.

> Is there some distro independent way to build spec files that we could add to the wiki as documentation?

Unfortunately I never built an rpm on another system than openSUSE. So I can't answer this. But I have some friends packaging for Fedora and we often exchange spec files. So even if the build process might be different having the spec file reduces the task greatly.
In the case of OBS the spec file is actually all thats needed and we can create all the rest from this.

patrick96:
Alright, I think we'll merge this anyway. Just one more questions, does this spec file work without the patch you mentioned in #1628?

jubalh:
Whether it works like this or not depends on the definition of the %cmake macro on the particular distro/build environment. So I can't really answer that for all cases.

But I think like this it already helps a lot. Personally I would add it like this.

patrick96:
Ah, I see.

patrick96:
Hey, @jubalh Are you still planning on maintaining this spec file? If not, I will remove it from the repo because it has become quite outdated.

jubalh:
@patrick96 I can update it next week :-)

patrick96:
Thanks :)

Just so you know, this wasn't meant to pressure you into updating the file. polybar is now in all the major rpm based distros, so we can also just remove the spec file.

jubalh:
@patrick96 sorry I forgot about this.
Today I updated the openSUSE package to 3.5.5 and then remembered it. Only then saw that you removed the spec file. Which is fine with me too :)

patrick96:
Since polybar is already in the official opensuse repos, I believe it doesn't make that much sense to keep updating the spec file here as well.

PR #1639, opened 2019-02-02
contrib: Add rpm fedora spec file

Vladius25:
Working spec for fedora with all modules. Everyone can build rpm from it.
[Check my copr](https://copr.fedorainfracloud.org/coprs/vladius/polybar/build/853364/).

### DIFF
--- a/contrib/rpm/polybar-fedora.spec
+++ b/contrib/rpm/polybar-fedora.spec
@@ -1,0 +1,71 @@
+Name: polybar	
+Version: 3.3.0
+Release: 1%{?dist}
+Summary: helps users build beautiful and highly customizable status bars
+
+License: MIT
+URL: https://github.com/jaagr/%{name}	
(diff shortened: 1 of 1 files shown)


patrick96:
Just a few days ago @jubalh added an openSUSE spec file (#1633). Is it possible to combine the two so that they work on both platforms? Looking at them, they share most of the content. I'd rather not have a spec file for every distro that supports them here and have to update them whenever changes are made. The main reason why I merged the openSUSE spec file was because it could also be useful to fedora users.

jubalh:
Yes that should be possible. I'm also for this. Maybe @Vladius25 can create another PR instead of this one that updates the existing spec file to be compatible with fedora. There are ifdefs in rpm spec to check for certain distros in case something needs to differ.
And OBS for example respects those ifdefs when it builds for various distros.

jubalh:
Sorry quite busy lately.

First glance looks good. I would take it.
If you want more thorough review you  need to take a couple of days.

PR #1644, opened 2019-02-06
Test compilation fails again

patrick96:
The `INSTANTIATE_TEST_CASE_P` macro when used with zero variadic arguments required a comma after the last argument to not generate any warnings, now it generates errors when there is a comma.


patrick96:
It was the only way how I knew to avoid the warning :smile: 

But kind of annoying that there is no way to have zero variadic arguments that works both now and before the "fix" (that I know of).

PR #1645, opened 2019-02-06
contrib: update rpm spec file to fedora support

Vladius25:
TEAMEDIT: Closes #1639

### DIFF
--- a/contrib/rpm/polybar.spec
+++ b/contrib/rpm/polybar.spec
@@ -1,70 +1,96 @@
 #
 # spec file for package polybar
-# Initially created for openSUSE
+# Created for openSUSE and Fedora
 #
+Name:          polybar	
+Version:       3.3.0
+Release:       1%{?dist}
+Summary:       helps users build beautiful and highly customizable status bars
 
-Name:           polybar
-Version:        3.3.0
-Release:        0
-Summary:        A fast and easy-to-use status bar
-License:        MIT
-Group:          System/GUI/Other
-URL:            https://github.com/jaagr/polybar
-Source:         https://github.com/jaagr/polybar/archive/%{version}.tar.gz
-BuildRequires:  clang >= 3.4
-BuildRequires:  cmake >= 3.1
-BuildRequires:  pkgconfig
-BuildRequires:  python-xml
-BuildRequires:  xcb-util-image-devel
-BuildRequires:  xcb-util-wm-devel
-# optional dependency
-BuildRequires:  pkgconfig(alsa)
-# main dependency
-BuildRequires:  pkgconfig(cairo)
-BuildRequires:  pkgconfig(jsoncpp)
-BuildRequires:  pkgconfig(libcurl)
-BuildRequires:  pkgconfig(libpulse)
-BuildRequires:  pkgconfig(python)
-BuildRequires:  pkgconfig(xcb)
-BuildRequires:  pkgconfig(xcb-proto)
-BuildRequires:  pkgconfig(xcb-util)
-%if 0%{?suse_version} <= 1315
-BuildRequires:  i3-devel
+Group:         System/GUI/Other
+License:       MIT
+URL:           https://github.com/jaagr/%{name}	
+Source0:       %{url}/releases/download/%{version}/%{name}-%{version}.tar
+
+BuildRequires: clang >= 3.4
+BuildRequires: cmake >= 3.1
+BuildRequires: git
+BuildRequires: pkgconfig(cairo)
+BuildRequires: pkgconfig(xcb)
+BuildRequires: pkgconfig(xcb-util)
+BuildRequires: pkgconfig(xcb-proto)
+BuildRequires: pkgconfig(xcb-image)
+BuildRequires: pkgconfig(xcb-ewmh)
+BuildRequires: pkgconfig(xcb-xrm)
+BuildRequires: pkgconfig(xcb-cursor)
+BuildRequires: pkgconfig(alsa)
+BuildRequires: pkgconfig(libpulse)
+BuildRequires: pkgconfig(jsoncpp)
+BuildRequires: pkgconfig(libcurl)
+BuildRequires: pkgconfig(libmpdclient)
+BuildRequires: pkgconfig(libnl-genl-3.0)
+
+%if 0%{?fedora}
+BuildRequires: ninja-build
+BuildRequires: python2
+BuildRequires: wireless-tools-devel
+BuildRequires: i3-ipc
 %else
-BuildRequires:  i3-gaps-devel
+BuildRequires: ninja
+BuildRequires: pkgconfig
+BuildRequires: python-xml
+BuildRequires: pkgconfig(python)
+%if 0%{?suse_version} <= 1315
+BuildRequires: i3-devel
+%else
+BuildRequires: i3-gaps-devel
 %endif
 %if 0%{?suse_version}
-BuildRequires:  libiw-devel
+BuildRequires: libiw-devel
 %else
-BuildRequires:  wireless-tools-devel
+BuildRequires: wireless-tools-devel
+%endif
 %endif
 
 %description
-A fast and easy-to-use status bar for tilling WM
+Polybar aims to help users build beautiful and highly customizable status bars for their desktop environment, without the need of having a black belt in shell scripting.
+
+The main purpose of Polybar is to help users create awesome status bars. It has built-in functionality to display information about the most commonly used services.
 
 %prep
-%setup -q
+%setup -q -n %{name}
+git config --global http.sslVerify false
+mkdir -p %{_target_platform}
 
 %build
-%cmake
-make
+pushd %{_target_platform}
+    cmake -G Ninja \
+    -DCMAKE_CXX_COMPILER=clang++ \
+    -DCMAKE_INSTALL_PREFIX:PATH=%{_prefix} \
+    -DBUILD_TESTS:BOOL=ON \
+    -DCMAKE_BUILD_TYPE=Debug \
+    ..
+popd
+%ninja_build -C %{_target_platform}
 
 %install
-%cmake_install
+%ninja_install -C %{_target_platform}
+
+%check
+pushd %{_target_platform}
+    ctest --output-on-failure
+popd
 
 %files
-%dir %{_datadir}/bash-completion/
-%dir %{_datadir}/bash-completion/completions
-%dir %{_datadir}/doc/%{name}
-%dir %{_datadir}/zsh/
-%dir %{_datadir}/zsh/site-functions
+%doc README.md
+%license LICENSE
+
 %{_bindir}/%{name}
 %{_bindir}/%{name}-msg
-%{_datadir}/doc/%{name}/config
-%{_mandir}/man1/%{name}.1%{?ext_man}
+%{_mandir}/man1/%{name}.1.gz
 %{_datadir}/bash-completion/completions/%{name}
 %{_datadir}/zsh/site-functions/_%{name}
 %{_datadir}/zsh/site-functions/_%{name}_msg
+%config(noreplace) %{_datadir}/doc/%{name}/config
 
 %changelog
-


patrick96:
Thanks! Let's wait for @jubalh to review it.

contrib/rpm/polybar.spec
```diff
@@ -1,70 +1,96 @@
...
-%setup -q
+%setup -q -n %{name}
+git config --global http.sslVerify false
+mkdir -p %{_target_platform}
 
 %build
-%cmake
-make
+pushd %{_target_platform}
+    cmake -G Ninja \
```
jubalh: does fedora not have a %cmake macro for such default flags?

contrib/rpm/polybar.spec
```diff
@@ -1,70 +1,96 @@
...
 %description
-A fast and easy-to-use status bar for tilling WM
+Polybar aims to help users build beautiful and highly customizable status bars for their desktop environment, without the need of having a black belt in shell scripting.
+
+The main purpose of Polybar is to help users create awesome status bars. It has built-in functionality to display information about the most commonly used services.
 
 %prep
-%setup -q
+%setup -q -n %{name}
+git config --global http.sslVerify false
```
jubalh: at which point in the build do you use git and why?

contrib/rpm/polybar.spec
```diff
@@ -1,70 +1,96 @@
...
-%dir %{_datadir}/zsh/
-%dir %{_datadir}/zsh/site-functions
+%doc README.md
+%license LICENSE
+
 %{_bindir}/%{name}
 %{_bindir}/%{name}-msg
-%{_datadir}/doc/%{name}/config
-%{_mandir}/man1/%{name}.1%{?ext_man}
+%{_mandir}/man1/%{name}.1.gz
```
jubalh: no man macro on fedora?

contrib/rpm/polybar.spec
Vladius25: %cmake has specific options, with which clang doesn't work at all, and gcc fails.

contrib/rpm/polybar.spec
Vladius25: Git is used in project for download google tests during build. OpenSuse leap doesn't clone git repo without ` http.sslVerify false`.
```
%check
pushd %{_target_platform}
    ctest --output-on-failure
popd
```

contrib/rpm/polybar.spec
Vladius25: It is an openSuse's macro.

contrib/rpm/polybar.spec
jubalh: OBS doesnt download anything during build because network connection is not allowed :-)

contrib/rpm/polybar.spec
jubalh: Which options do fail? Usually distro macros should be used for such cases because they guarantee that the binary is compiled with the flags the distro wants. It could be that they prefer position independent binaries etc.

Maybe you are referring to this issue: https://github.com/jaagr/polybar/issues/1628 ?

contrib/rpm/polybar.spec
Vladius25: It also means that build machine must use network. It is possibe in copr, but not in Fedora koji (dont't know about openSuse). Tests can be disabled. Another opportunity is `Source1:`, but CMakeLists have to be edited in this case.

contrib/rpm/polybar.spec
Vladius25: Yes, the same with gcc. But clang can't compile cmake's test program because of key, which it hasn't got. Maybe, i should use gcc instead.

contrib/rpm/polybar.spec
jubalh: To have reproducible builds one needs pristine sources

contrib/rpm/polybar.spec
jubalh: key?

Vladius25: I mean `-fstack-clash-protection`
```
CMake Error at /usr/share/cmake/Modules/CMakeTestCXXCompiler.cmake:45 (message):
  The C++ compiler

    "/usr/bin/clang++"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /builddir/build/BUILD/polybar-3.3.0/x86_64-redhat-linux-gnu/CMakeFiles/CMakeTmp
    
    Run Build Command:"/usr/bin/gmake" "cmTC_1f3b6/fast"
    /usr/bin/gmake -f CMakeFiles/cmTC_1f3b6.dir/build.make CMakeFiles/cmTC_1f3b6.dir/build
    gmake[1]: Entering directory '/builddir/build/BUILD/polybar-3.3.0/x86_64-redhat-linux-gnu/CMakeFiles/CMakeTmp'
    Building CXX object CMakeFiles/cmTC_1f3b6.dir/testCXXCompiler.cxx.o
    /usr/bin/clang++    -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection    -o CMakeFiles/cmTC_1f3b6.dir/testCXXCompiler.cxx.o -c /builddir/build/BUILD/polybar-3.3.0/x86_64-redhat-linux-gnu/CMakeFiles/CMakeTmp/testCXXCompiler.cxx
    clang-7: error: unknown argument: '-fstack-clash-protection'
    gmake[1]: *** [CMakeFiles/cmTC_1f3b6.dir/build.make:66: CMakeFiles/cmTC_1f3b6.dir/testCXXCompiler.cxx.o] Error 1
    gmake[1]: Leaving directory '/builddir/build/BUILD/polybar-3.3.0/x86_64-redhat-linux-gnu/CMakeFiles/CMakeTmp'
    gmake: *** [Makefile:121: cmTC_1f3b6/fast] Error 2

```

contrib/rpm/polybar.spec
patrick96: What is the rationale behind running tests when building the package?

contrib/rpm/polybar.spec
patrick96: According to the [Fedora Packaging Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/#compiler), gcc should be used if possible.

contrib/rpm/polybar.spec
Vladius25: Maybe useful sometimes, but not really needed, i think.

contrib/rpm/polybar.spec
Vladius25: @jubalh How did you fix #1628 ? I don't see `Patch` section in your spec.

Vladius25: > According to the [Fedora Packaging Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/#compiler), gcc should be used if possible.

Yes, but when F30 decides to use gcc9 with lots of regressions, it's hard to use gcc ;)

patrick96: > @jubalh How did you fix #1628 ? I don't see `Patch` section in your spec.

It's used [here](https://build.opensuse.org/package/show/X11:Utilities/polybar). The next release should contain the fix for it.

F30 is already using an unreleased gcc version? :open_mouth:

contrib/rpm/polybar.spec
jubalh: > What is the rationale behind running tests when building the package?

If one packages a python package for example it makes sense to run the tests to see if runtime behaviour also works. In the case of polybar I decided not to run tests as I think they are mostly used in the development cycle.

contrib/rpm/polybar.spec
jubalh: If Fedora doesn't have this macro then let's use your way here, since it works for both.

contrib/rpm/polybar.spec
Vladius25: > Can you force gcc8?
No, it's impossible for rawhide(F30). Many packages was broken after mass rebuild with gcc 9. It's not such important, because rawhide is rawhide.

contrib/rpm/polybar.spec
jubalh: I would disable tests (removing this section) here and don't run git. For the reason that build systems usually try to run in a clean environment and without internet access.

contrib/rpm/polybar.spec
jubalh: I would suggest to use gcc here, and use the %cmake macro.
The maintainer of the openSUSE package decided to use clang, when I revisited the spec file I also suggested to use gcc there. But now I see that it's again clang. So I'll open another request for this at openSUSE.

jubalh:
BTW. + 1 for @patrick96 for doing very good maintainer work and really caring about this project.

contrib/rpm/polybar.spec
Vladius25: `%cmake` in OpenSuse is 
```
/usr/bin/cmake /builddir/build/BUILD/polybar/. (args)
```
 whilest in Fedora it's just
```
/usr/bin/cmake (args)
```
What do you think about?
```
%if 0%{?fedora}
%define cmakedir .
[Fedora build requires]
%else
[OpenSuse build requires]
%endif

%build
%cmake %{?cmakedir} 
%make_build

%install
%make_install
```

contrib/rpm/polybar.spec
patrick96: Yeah, for polybar, tests are mainly a development tool. Also we try not to add code that breaks tests ;)

contrib/rpm/polybar.spec
patrick96: > Many packages was broken after mass rebuild with gcc 9.

I'll try to fix the gcc9 breakage once it is available on arch, we'll probably need to do a patch release for that anyway

patrick96:
Thanks @jubalh! I appreciate that :smiley:

contrib/rpm/polybar.spec
jubalh: Sounds alright. Or having ifdefs in the build/install sections.

jubalh: > `%cmake` in OpenSuse is
```
/usr/bin/cmake /builddir/build/BUILD/polybar/. (args)
```

Actually %cmake in openSUSE is:
```
rpm --eval %cmake

    find . -name CMakeLists.txt \
        -exec sed -i -re '/^[[:blank:]]*[sS][eE][tT][[:blank:]]*\([[:blank:]]*(CMAKE_BUILD_TYPE|CMAKE_COLOR_MAKEFILE|CMAKE_INSTALL_PREFIX|CMAKE_VERBOSE_MAKEFILE).*\)/{s/^/#IGNORE /}' {} + 
    mkdir -p build 
    cd build 
    /usr/bin/cmake $OLDPWD/. \
%if "/usr/bin/make" == "/usr/bin/make" 
        -G"Unix Makefiles" \
%else 
        -GNinja \
%endif 
        -DCMAKE_INSTALL_PREFIX:PATH=/usr \
        -DINCLUDE_INSTALL_DIR:PATH=/usr/include \
        -DLIB_INSTALL_DIR:PATH=/usr/lib64 \
        -DSYSCONF_INSTALL_DIR:PATH=/etc \
        -DSHARE_INSTALL_PREFIX:PATH=/usr/share \
        -DCMAKE_INSTALL_LIBDIR:PATH=/usr/lib64 \
        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
        -DCMAKE_C_FLAGS="${CFLAGS:--O2 -g -m64 -fmessage-length=0 -D_FORTIFY_SOURCE=2 -fstack-protector -funwind-tables -fasynchronous-unwind-tables} -DNDEBUG" \
        -DCMAKE_CXX_FLAGS="${CXXFLAGS:--O2 -g -m64 -fmessage-length=0 -D_FORTIFY_SOURCE=2 -fstack-protector -funwind-tables -fasynchronous-unwind-tables} -DNDEBUG" \
        -DCMAKE_Fortran_FLAGS="${FFLAGS:--O2 -g -m64 -fmessage-length=0 -D_FORTIFY_SOURCE=2 -fstack-protector -funwind-tables -fasynchronous-unwind-tables} -DNDEBUG" \
        -DCMAKE_EXE_LINKER_FLAGS="-Wl,--as-needed -Wl,--no-undefined -Wl,-z,now" \
        -DCMAKE_MODULE_LINKER_FLAGS="-Wl,--as-needed" \
        -DCMAKE_SHARED_LINKER_FLAGS="-Wl,--as-needed -Wl,--no-undefined -Wl,-z,now" \
%if "lib64" == "lib64" 
        -DLIB_SUFFIX=64 \
%endif 
        -DCMAKE_SKIP_RPATH:BOOL=ON \
        -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
        -DBUILD_SHARED_LIBS:BOOL=ON \
        -DBUILD_STATIC_LIBS:BOOL=OFF \
        -DCMAKE_COLOR_MAKEFILE:BOOL=OFF \
        -DCMAKE_INSTALL_DO_STRIP:BOOL=OFF \
        -DCMAKE_MODULES_INSTALL_DIR=/usr/share/cmake/Modules
```

PR #1653, opened 2019-02-16
fix(build): Make text coloration more portable

softmoth:
Using echo "\033[" is not portable, as some shells require "echo -e" and
some don't accept "-e". Using printf is better, as its behavior is
consistent. But in this case, calling out to tput is best since it also
handles all terminal types. https://unix.stackexchange.com/a/274542/167151

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,25 +11,25 @@ help:
 	@awk -F':' '/^\w*:/ {print "  "$$1;}' Makefile
 
 configure:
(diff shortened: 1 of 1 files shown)


patrick96:
Hi. Thanks a lot for taking time to contribute :smiley: 

However, there are already changes in the pipeline that remove this `Makefile` completely since it is superseded by the `build.sh` script (see #1553). Thanks anyway ;)

PR #1656, opened 2019-02-18
Fixed minor menu item-spacing issue

jstamant:
There was missing some spacing between item+separator in my menu using `expand-right=false`.
Thought this would be a great opportunity for first-time pull-request.

Searching for 'menu' in the issues-tracker reveals no pertinent issues to me.

In this image, you can see the missing spacing between my the 'reboot' label and its separator in the top-right. I've fixed this behaviour, and it works regardless of `expand-right=true` or `expand-right=false`
2019-02-17-172128-scrot

### DIFF
--- a/src/modules/menu.cpp
+++ b/src/modules/menu.cpp
@@ -89,12 +89,10 @@ namespace modules {
             builder->node(m_labelseparator);
             builder->space(spacing);
           }
-        } else if (item == m_levels[m_level]->items.back()) {
-          //Insert separator after last menu-item and before menu-toggle for expand-right=false
-          if (!m_expand_right && *m_labelseparator) {
-            builder->space(spacing);
-            builder->node(m_labelseparator);
-          }
+        //Insert separator after last menu-item and before menu-toggle for expand-right=false
+        } else if (!m_expand_right && *m_labelseparator) {
+          builder->space(spacing);
+          builder->node(m_labelseparator);
         }
       }
     } else {
--- a/src/modules/menu.cpp
+++ b/src/modules/menu.cpp
@@ -74,10 +74,9 @@ namespace modules {
       builder->cmd_close();
     } else if (tag == TAG_MENU && m_level > -1) {
       auto spacing = m_formatter->get(get_format())->spacing;
-      //Insert spacing after menu-items and before menu-toggle for expand-right=true
-      if (m_expand_right) {
-        if (*m_labelseparator)
-          builder->node(m_labelseparator);
+      //Insert separator after menu-toggle and before menu-items for expand-right=true
+      if (m_expand_right && *m_labelseparator) {
+        builder->node(m_labelseparator);
         builder->space(spacing);
       }
       for (auto&& item : m_levels[m_level]->items) {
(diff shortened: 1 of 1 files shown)


patrick96:
Thanks a lot. Will review when I have some time

src/modules/menu.cpp
```diff
@@ -74,38 +74,29 @@ namespace modules {
       builder->cmd_close();
     } else if (tag == TAG_MENU && m_level > -1) {
       auto spacing = m_formatter->get(get_format())->spacing;
+      //Insert spacing after menu-items and before menu-toggle for expand-right=true
+      if (m_expand_right) {
+        if (*m_labelseparator)
+          builder->node(m_labelseparator);
+        builder->space(spacing);
```
Lomadriel: If I am not wrong this space should only be printed if there is a separator.
As you can see in the following screenshot, it seems that there is two spaces between the "x" and "Browsers"
poly

src/modules/menu.cpp
```diff
@@ -74,38 +74,29 @@ namespace modules {
...
-         * <label-toggle> x | y |
-         */
-        if(m_expand_right) {
-          builder->cmd(mousebtn::LEFT, item->exec);
-          builder->node(item->label);
-          builder->cmd_close();
+        } else if (item == m_levels[m_level]->items.back()) {
+        //Insert spacing after menu-toggle and before menu-items for expand-right=false
+          if (!m_expand_right) {
+            builder->space(spacing);
```
Lomadriel: Same as above:
poly

src/modules/menu.cpp
jstamant: That's correct: The spacing should have only been printed if the user specified a separator. I've made a commit to address this. I didn't test this code without a separator, originally. I've tested this new commit thoroughly with with a monospace font and all variations of `expand-right` and `label-separator`.

Screenshot_2019-05-17_17-59-45
Screenshot_2019-05-17_17-58-59


The correct `format-spacing` in the `menu` module is already printed between the menu-toggle and menu-items. I'm actually unsure where it comes from, which is why this separator+spacing issue came up. I didn't account for it. I don't see any signs of the `menu` module inserting the spacing, so I suspect that the 'builder' class or some formatter or something is magically inserting the right amount of spacing, making my life easier.

Lomadriel: It's not the builder nor the a formatter, the space is inserted here ;)
https://github.com/polybar/polybar/blob/207d148bfb3a88e7979703929f9f8a4532a2c7f9/src/modules/menu.cpp#L87

src/modules/menu.cpp
```diff
@@ -74,38 +74,27 @@ namespace modules {
...
-        /*
-         * If the menu expands to the right, the separator should be drawn on the left side because otherwise the menu
-         * would look like this:
-         * <label-toggle> x | y |
-         */
-        if(m_expand_right) {
-          builder->cmd(mousebtn::LEFT, item->exec);
-          builder->node(item->label);
-          builder->cmd_close();
+        } else if (item == m_levels[m_level]->items.back()) {
```
patrick96: The condition in this if is always true when the else branch is taken (it's the opposite of the if condition above).

I will quickly fix this myself and push to this repo if you have that enabled.

patrick96:
Alright, travis went through, merging this now.

Thanks for contributing :)

jstamant:
Thanks everyone: that was a positive experience for my first time contributing code!

patrick96:
@jstamant I'm glad you feel that way. Sorry it took so long.

PR #1661, opened 2019-02-25
Fix xkeyboard crashing when indicators where enabled

### LINKED ISSUE #1660: xkeyboard displaying incorrect information and crashing
## Describe the issue
I recompiled polybar from source today, and I noticed that the xkeyboard module is totally misbehaving. When I reloaded my polybar, I noticed that it showed I had both my Capslock and Scroll lock keys on, which I definitely did not (now that I think about it, k don't even have a scroll lock key). I tried pressing Capslock a few times, and polybar crashed. I tried again a few times, and it seemed that pressing Capslock a few times causes it to crash repeatedly.

**Expected behavior:**
Polybar's xkeyboard module should only show the Capslock and Scroll lock indicators when the respective keys are on. In addition, polybar should not crash when the caps lock key is hit a few times.

**Actual behavior:**
Polybar always shows the Capslock and Scroll lock indicators. In addition, it crashes when Capslock is hit a few times.

**Was it working before?**
I last compiled polybar from commit ca4426a9, which worked fine. Now, it fails on commit db9a83a8. (Curiously, this commit is about xkeyboard related stuff.)

## To Reproduce
My configuration is basically the stock example configuration. I haven't made any changes to xkeyboard. I will test out a sample MVE config tomorrow.

Launch polybar. Note Capslock and Scroll lock indicators. Press Capslock a few times. Note crash.

## Polybar Log
I will try to add this tomorrow.

## Screenshots
I will try to add this tomorrow.

## Environment:
* WM: i3-gaps
* Output of `polybar -vvv`: I will try to add this tomorrow.

patrick96:
Fixes #1660

### DIFF
--- a/src/modules/xkeyboard.cpp
+++ b/src/modules/xkeyboard.cpp
@@ -63,7 +63,7 @@ namespace modules {
     }
 
     if (m_formatter->has(TAG_LABEL_INDICATOR)) {
-      m_conf.warn_deprecated(name(), TAG_LABEL_INDICATOR, "label-indicator-on");
+      m_conf.warn_deprecated(name(), "label-indicator", "label-indicator-on");
       // load an empty label if 'label-indicator-off' is not explicitly specified so
       // no existing user configs are broken (who expect nothing to be shown when indicator is off)
       m_indicator_state_off = load_optional_label(m_conf, name(), "label-indicator-off"s, ""s);
--- a/src/modules/xkeyboard.cpp
+++ b/src/modules/xkeyboard.cpp
@@ -130,32 +130,34 @@ namespace modules {
       m_layout->replace_token("%number%", to_string(m_keyboard->current()));
     }
 
-    m_indicators.clear();
+    if (m_formatter->has(TAG_LABEL_INDICATOR)) {
+      m_indicators.clear();
 
-    for (auto it : INDICATOR_TYPES) {
-      const auto& indicator_str = m_keyboard->indicator_name(it);
+      for (auto it : INDICATOR_TYPES) {
+        const auto& indicator_str = m_keyboard->indicator_name(it);
 
-      if (blacklisted(indicator_str)) {
-        continue;
+        if (blacklisted(indicator_str)) {
+          continue;
+        }
+
+        auto indicator_on = m_keyboard->on(it);
+        auto &indicator_labels = indicator_on ? m_indicator_on_labels : m_indicator_off_labels;
+        auto &indicator_icons = indicator_on ? m_indicator_icons_on : m_indicator_icons_off;
+        auto &indicator_state = indicator_on ? m_indicator_state_on : m_indicator_state_off;
+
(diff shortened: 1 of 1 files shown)


arjvik:
I just tested the PR out, and it works great! Thanks @patrick96 for fixing this so quickly!

ddffnn:
Even when using `patrick96:issue/1660` polybar generates a segmentation fault when I use `format = <label-layout>` in module/xkeyboard. If I use `format = <label-name>` polybar runs without issue.

ddffnn:
Just chiming in to say that [16a714a](https://github.com/jaagr/polybar/pull/1661/commits/16a714a9edbef3a73071269913f176efc6e284e8) fixed the seg fault for me. I use `format = <label-layout>`.

patrick96:
@ddffnn This only occured to me because of the comment you posted (and then deleted?) :wink:

ddffnn:
Well, I didn't get the crash when I used `format = <label-name>`, and I put that in my comment. Then I realized that `<label-name>` isn't even valid so the module didn't show up at all. I assumed that the seg fault didn't occur because I had essentially turned off the module. I deleted my comment because I thought it was just noise. I'm glad it wasn't!

arjvik:
Wait I'm confused - with the new system, what should the format be? I compiled 2432a7e and tested that it works fine with (basically) the default config from a while back that I slightly tweaked. This is the same config that crashed db9a83a but worked on ca4426a.

Here is the relevant section of mine in case it is needed:
```dosini
[module/xkeyboard]
type = internal/xkeyboard
blacklist-0 = num lock

format-prefix = " "
format-prefix-foreground = ${colors.foreground-alt}
format-prefix-underline = ${colors.secondary}

label-layout = %layout%
label-layout-underline = ${colors.secondary}

label-indicator-padding = 2
label-indicator-margin = 1
label-indicator-background = ${colors.secondary}
label-indicator-underline = ${colors.secondary}
```

ddffnn:
The new feature that allows setting the icon based on the current layout doesn't seem to work for layout variants. I run `setxkbmap -layout dvorak,us` and have my xkeyboard module configured with `format = <label-layout>` and `label-layout = %name%`. Both dvorak and us have the same `%layout%` value (us) so I don't display it. This means that I can't have different icons selected for each variant. Maybe the icons should match against the `%name%` value?

patrick96:
@arjvik The format should work as it always has, we didn't change anything there. This PR is only here to fix crashes that occured with valid formats. The supported values for `format` are listed on the [Wiki](https://github.com/jaagr/polybar/wiki/Module:-xkeyboard#additional-formatting)

@ddffnn This a PR to fix a bug. Please open feature requests as a new issue. You should tag Gilnaa who originally implemented this, maybe he can add this feature as well.

arjvik:
When will this be merged?

NBonaparte:
Now.

PR #1662, opened 2019-02-26
Fix data race in get_output/update

DanielChabrowski:
Thread sanitizer reported a data race for cpu_module and memory_module, my guess is that they should show up eventually for all modules that update their content frequently - battery as well?

Thread sanitizer output for cpu_module:
https://pastebin.com/Raj6hLbh

Since we would end up with always locking buildlock and updatelock, buildlock was removed and the other one renamed.

### DIFF
--- a/include/modules/meta/base.hpp
+++ b/include/modules/meta/base.hpp
@@ -144,8 +144,7 @@ namespace modules {
     const logger& m_log;
     const config& m_conf;
 
-    mutex m_buildlock;
-    mutex m_updatelock;
+    mutex m_contentlock;
     mutex m_sleeplock;
     std::condition_variable m_sleephandler;
 
--- a/include/modules/meta/base.inl
+++ b/include/modules/meta/base.inl
@@ -54,9 +54,7 @@ namespace modules {
     m_log.info("%s: Stopping", name());
     m_enabled = false;
 
-    std::lock(m_buildlock, m_updatelock);
-    std::lock_guard<std::mutex> guard_a(m_buildlock, std::adopt_lock);
-    std::lock_guard<std::mutex> guard_b(m_updatelock, std::adopt_lock);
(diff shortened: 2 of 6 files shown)


patrick96:
Hey, thanks a lot!

Seems completely logical that if we have different locks for building (reading data) and for updating that we have a read-write data race.

Unfortunately this PR makes my battery module (and possibly others) deadlock. [Here](https://gist.github.com/patrick96/75b250241ccb6eb2e9c4037266964952) is where it is stuck (I used `thread apply all where` in `gdb` after letting it run for a few seconds).

I used the following config:
```dosini
[bar/example]
modules-right = battery 

[module/battery]
type = internal/battery
battery = BAT0
adapter = ADP1
```

I think the actual deadlock happens in Thread 5 where it wants to acquire `m_contentlock` but somehow can't. I suspect this is because Thread 4 has already acquired this lock. I don't really have time to dig into this further. Maybe you can figure out what's wrong.

DanielChabrowski:
Interesting, I'll take a look. Thank you for checking.

Thread 4 uses `m_sleeplock` so that shouldn't be an issue.

DanielChabrowski:
While I couldn't check the battery module on my vm, i3 module seems to be stuck as well. For the i3 the issue is that `i3_module::has_event` is blocking and contentlock is not released until there is some kind of event, like switching to other workspace. While the ipc is waiting for the event, `get_output` is called from another thread and it tries to lock the mutex again.

I'm guessing that battery_module has a similar issue with its `inotify_module::poll_events` function but I would have to confirm that first.

It really is threading madness right now.
https://gist.github.com/DanielChabrowski/c5aa668cb2e1631aa2b5877ce2613596

I'm closing this for now, this commit alone won't fix the issues.

patrick96:
Yeah, I noticed that too that polybar has tons of data races :disappointed:

Issue #1666, opened 2019-03-02
Detached systray percentage x offset not working


memeplex:
When I set a detached tray at the right side and configure a x offset of, say, -100 it works as expected. But if the offset is, say, -20% it doesn't seem to be taken into account.

**Expected behavior:**

Offset should work no matter whether it is specified in pixels or percentage.

**Actual behavior:**

Offset is only honoured when specified in pixels.

**Was it working before?**

I don't know. This was the first time I enabled (and disabled again) the systray feature.

## To Reproduce

```dosini
tray-position = right
tray-detached = true
tray-offset-x = -20%
```

## Environment:

* WM:
```
~:: i3 --version
i3 version 4.16.1 (2019-01-27) © 2009 Michael Stapelberg and contributors
```

* Output of `polybar -vvv`:

```
polybar 3.3.0

Features: +alsa +curl +i3 -mpd +network(libnl) +pulseaudio +xkeyboard

X extensions: +randr (+monitors) -render -damage -sync +composite +xkb +xrm +xcursor

Build type: RelWithDebInfo
Compiler: /usr/bin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -Wall -Wextra -Werror -Wno-noexcept-type -pedantic -pedantic-errors
Linker flags: -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now
```
## Additional context

```
~:: uname -a
Linux neu 4.20.12-arch1-1-ARCH #1 SMP PREEMPT Sat Feb 23 15:11:34 UTC 2019 x86_64 GNU/Linux
```

PR #1669, opened 2019-03-03
Fix tray offset #1666

Lomadriel:
As reported in #1666, negative percentage in tray are ignored.

It seems that the function (```percentage_to_value```) used to compute the value from the percentage clamps the value between ```0``` and ```max_value```.
To fix this problem, I added a ```signed_percentage_to_value``` function which clamps the value between ```-max_value``` and ```max_value``` and so it allows percentage between -100% and 100%.

However, when I fixed this bug, I noticed something strange, is it okay that the bar stops where the tray is supposed to be without any offset (see image below)?
image

In the wiki it is written that only the content should be shifted, so I wonder if this is a bug.

Minimal config:
```
[bar/example]
width = 100%
height = 27

font-0 = fixed:pixelsize=10;1
font-1 = unifont:fontformat=truetype:size=8:antialias=false;0
font-2 = siji:pixelsize=10;1

modules-center = test

tray-position = right
tray-padding = 2
tray-detached = false
tray-offset-x = -10%
tray-background = #0063ff
border-size = 4
border-color = #00000000

background = #222
foreground = #dfdfdf

[module/test]
type = custom/text
content = Default

```

With ```tray-detached = true```, the behavior is as expected since the content of the bar is not shifted.
image

### DIFF
--- a/include/utils/math.hpp
+++ b/include/utils/math.hpp
@@ -56,7 +56,18 @@ namespace math_util {
   }
 
   /**
-   * Get value for percentage of `max_value`
+   * Get value for signed percentage of `max_value` (cap between -max_value and max_value)
+   */
+  template <typename ValueType, typename ReturnType = int>
+  ReturnType signed_percentage_to_value(ValueType signed_percentage, ValueType max_value) {
+    if (std::is_integral<ReturnType>())
+      return cap<ReturnType>(signed_percentage * max_value / 100.0f + 0.5f, -max_value, max_value);
+    else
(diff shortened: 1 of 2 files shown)


patrick96:
> However, when I fixed this bug, I noticed something strange, is it okay that the bar stops where the tray is supposed to be without any offset (see image below)?

Not sure if this is intended behavior but this also happens when using `tray-offset-y`, so it has been around for a long time. Since these offsets are meant to be used with a detached tray, it shouldn't be too bad.

Issue #1527, opened 2018-11-25
Offset doesn't work


zandimna:
Describe the issue
*A clear and concise description of what the bug is.*
offset-x and offset-y options are does not work
**Expected behavior:**
the bar should shift by {} pixels to the right and {} pixels down
**Actual behavior:**
the bar itself didn't move, but the tray icons did
## To Reproduce
default config, with offset-x and offset-y uncommented

## Polybar Log
```
warn: No monitor specified, using "DP-1"
warn: pulseaudio: using default sink alsa_output.pci-0000_00_1f.3.analog-stereo
error: tray: Failed to put tray above 0x3000002 in the stack (XCB_MATCH (8))

```

## Screenshots
https://imgur.com/a/MJfQ7rF
## Environment:
WM: i3
Output of `polybar -vvv`:
```
polybar 3.2.1
Features: +alsa +curl +i3 -mpd +network +pulseaudio +xkeyboard
X extensions: +randr (+monitors) -render -damage -sync -composite +xkb +xrm +xcursor
Build type: RelWithDebInfo
Compiler: /usr/bin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -Wall -Wextra -Werror -Wno-noexcept-type -O2 -pedantic -pedantic-errors
Linker flags: -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now
```

zandimna:
> See #1101, you need to set `override-redirect` to `true`.

Okay thanks. I encountered the same issues as described there. I'll just use border-size-top instead.

patrick96:
i3 is the worst when it comes to bar positioning :sweat_smile:

PR #1671, opened 2019-03-04
Add units support for polybar #1651 #951

### LINKED ISSUE #951: Support DPI-dependent units
Currently, properties such as `width` and `height` only support pixel values or percentages. However, this is pretty limiting for users who wish to use the same bar across different monitors. To give a specific use-case, I have three setups I switch between regularly:

 - Bar is on a high-DPI (144) monitor with resolution 2560x1440
 - Bar is on a regular-DPI (96) monitor with resolution 1920x1080
 - Bar is on a rotated regular-DPI (96) monitor with resolution 1200x1920

Picking a single value for `height` here becomes really difficult: a pixel value will cause the bar to either be way too small when the DPI is high, or way too tall when the DPI is low, but a percentage value will cause the bar to be really strangely sized on the rotated monitor. Even if the monitor isn't rotated, different aspect ratios can cause a height of 3% to look very different (compare 1920x1080 and 1920x1200 for example).

It would be very useful to have a DPI-dependent unit for these properties to complement the existing two attributes. We can probably draw some inspiration from [CSS length units](https://developer.mozilla.org/en-US/docs/Web/CSS/length) here, and add a unit like `em`, which is scaled relative to the current font size, which in turn is affected by the DPI (or should be if it isn't already). Alternatively, we could go with `in`/`mm`, though this would break with CSS, in which [`1in == 96px`](https://developer.mozilla.org/en-US/docs/Web/CSS/length#CSS_units_and_dots-per-inch).

### LINKED ISSUE #1265: "module-margin-left = -1" causes total system halt, Arch AUR package for Polybar 3.1
I'm just working on my polybar config, experimenting with settings just to see what happens, and I set this setting in the default polybar config:

`module-margin-left = -1`

I'm running Arch, with linux-lts 4.14.44, and xorg-server 1.20.0-2, with Intel hd3000 graphics using the kernel kms intel driver, without compton running (normally I would, but haven't set it up yet).  I'm running a vanilla i3-gaps config (pulled via the i3 group in Arch), and am launching polybar via dmenu running `polybar example`, and starting my session with sddm.  If there's any other info that can help, I'll share.

Just tested, this also happens on git the latest commit/version b70d5b5

### LINKED ISSUE #1651: Allow `px` units when specified

For example:  
`label-unfocused-padding = 0`  
`label-unfocused-padding = 1`

### LINKED ISSUE #1700: Offset tag doesn't work as indicated in the documentation
## Describe the issue

When I use the `%{O}` tag it moves the whole block and not only the text to its right.

**Expected behavior:**

The `%{O}` tag only move the text to its right as the documentation said:

> ## Offset `%{O}`
> %{O10} pushed 10 pixels to the right %{O-5} 5 pixels to the left

**Actual behavior:**

The whole block is moved. By example if I use `"%{O10} move right %{O-5} move left"`, the whole block is moved 5 pixels to the right (10-5).

**Was it working before?**
* Did you also experience this bug in an earlier version of polybar? Don't know
* If no, what was the last version where this worked correctly? Don't know

## To Reproduce
```dosini
[bar/example]
width = 100%
height = 27

font-1 = fixed:pixelsize=10;1
font-0 = Inconsolata:style=Bold:pixelsize=13;2
font-2 = siji:pixelsize=10;1

modules-left = label_test
modules-center =
modules-right =

[module/label_test]
type = custom/text
content = %{O10} pushed 10 pixels to the right %{O-5} 5 pixels to the left

[module/xwindow]
type = internal/xwindow
label = %title:0:30:...%

```
## Polybar Log
```
* Bar geometry: 1920x27+0+0; Borders: 0,0,0,0
- bar: Attach X event sink
- bar: Attach signal receiver
* Starting application
* Starting module/label_test
- module/label_test: Thread id = 2
- bar: Create renderer
- renderer: Allocate colormap
- renderer: Load fonts
* Configured DPI = 96x96
* Loaded font "Inconsolata:style=Bold:pixelsize=13" (name=Inconsolata, offset=2, file=/usr/share/fonts/TTF/Inconsolata-Bold.ttf)
* Loaded font "fixed:pixelsize=10" (name=DejaVu Sans, offset=1, file=/usr/share/fonts/TTF/DejaVuSans.ttf)
* Loaded font "siji:pixelsize=10" (name=DejaVu Sans, offset=1, file=/usr/share/fonts/TTF/DejaVuSans.ttf)
* Bar window: 0x3a00001
- bar: Reconfigure window
- bar: Set window WM_NAME
- bar: Set window _NET_WM_PID
- bar: Map window
- bar: Draw empty bar
- bar: Setup tray manager
* module/label_test: Rebuilding cache
- bar: Force update
* Redrawing bar window
- renderer: flush(1 geom=473x27+0+0, falloff=0)
- bar: Received expose event
```

## Screenshots
polybar

## Environment:
* Output of `polybar -vvv`:
```
polybar 3.3.0

Build type: RelWithDebInfo
Compiler: /usr/bin/c++
```
or (current master)
```
polybar 3.3.0-44-gded7a4f-git

Build type: Debug
Compiler: /usr/bin/c++
Compiler flags:  -Wall -Wextra -Werror -Wno-noexcept-type -pedantic -pedantic-errors
Linker flags: 
```
 
## Additional context
While this problem isn't resolves #1671 is blocked. If this is really a bug and not the standard behavior, I can try to fix this bug in #1671 (since I need to modify the `%{O}` tag to add pt support)

Lomadriel:
Implement feature requested in #1651 and #951.
This pr add supports for `px` (pixel) and `pt` (point) for polybar.

I followed the advices given by @patrick96 in #1651, so I add a `size_with_unit` structure to store the value and the unit. Unlike the `size` definition, the value is signed to handle negative offset.

To handle the `width` and `height` problem, I added a `geometry_format_values` structure which contains a percentage and the offset (a `size_with_unit` structure).

`builder::space` now uses the `%{O}` formatting tag but this is handled by `size_with_unit_to_string` helper in the new header `"utils/unit.hpp"` to avoid duplicate code (this function is also used in `controller.cpp` since I can't use `builder::space` here).

DPI initialization is now in the constructor of the bar. However this initialization seems strange, the DPI are only computed if `dpi < 0` is explicitly specified in the configuration (see below). If the DPI isn't specified in the configuration, the default value is 96. Is it the correct behavior?

https://github.com/Lomadriel/polybar/blob/cc4e09edc41cfbd5f2f4bf7235470aa1c21bc59a/src/components/bar.cpp#L155-L178

The rest of the PR is just adapting the code to use these structures. I tried to find all keys that could uses units but I am not sure if I added units support for all the key that could benefit of this feature.

EDIT: Fixes #1700

TEAMEDIT:
Closes #1651 
Closes #951
Fixes #1265

### DIFF
--- a/include/components/types.hpp
+++ b/include/components/types.hpp
@@ -83,20 +83,12 @@ enum class space_type { SPACE, POINT, PIXEL };
 
 enum class size_type { POINT, PIXEL };
 
-namespace details {
-  template <typename T>
-  struct size_with_unit_impl {
-    space_type type{space_type::SPACE};
-    T value{0U};
-  };
-}  // namespace details
-
 struct space_size {
   space_type type{space_type::SPACE};
   float value{0.f};
 };
 
-struct ssize_with_unit {
+struct geometry {
   size_type type{size_type::PIXEL};
   float value{0.f};
 };
@@ -108,7 +100,7 @@ struct side_values {
 
 struct geometry_format_values {
   double percentage{0.};
-  ssize_with_unit offset{size_type::PIXEL, 0};
+  geometry offset{size_type::PIXEL, 0};
 };
 
 struct edge_values {
--- a/include/components/types.hpp
+++ b/include/components/types.hpp
@@ -80,10 +80,16 @@ struct size {
 
 enum class unit_type { SPACE, POINT, PIXEL };
 
-struct size_with_unit {
-  unit_type type{unit_type::SPACE};
-  int value{0U};
-};
+namespace details {
+  template <typename T>
+  struct size_with_unit_impl {
+    unit_type type{unit_type::SPACE};
+    T value{0U};
+  };
+}  // namespace details
+
+using size_with_unit = details::size_with_unit_impl<unsigned long int>;
+using ssize_with_unit = details::size_with_unit_impl<long int>;
 
 struct side_values {
   size_with_unit left{unit_type::SPACE, 0U};
@@ -91,8 +97,8 @@ struct side_values {
 };
 
 struct geometry_format_values {
-    double percentage{0.};
-    size_with_unit offset{unit_type::PIXEL, 0U};
+  double percentage{0.};
+  ssize_with_unit offset{unit_type::PIXEL, 0U};
 };
 
 struct edge_values {
--- a/include/components/types.hpp
+++ b/include/components/types.hpp
@@ -88,19 +88,23 @@ struct space_size {
   float value{0.f};
 };
 
+static constexpr space_size ZERO_SPACE = {space_type::SPACE, 0.};
+
 struct geometry {
   size_type type{size_type::PIXEL};
   float value{0.f};
 };
 
+static constexpr geometry GEOMETRY_ZERO_PIXEL = {size_type::PIXEL, 0.f};
+
 struct side_values {
-  space_size left{space_type::SPACE, 0_z};
-  space_size right{space_type::SPACE, 0_z};
+  space_size left{ZERO_SPACE};
+  space_size right{ZERO_SPACE};
 };
 
 struct geometry_format_values {
   double percentage{0.};
-  geometry offset{size_type::PIXEL, 0};
+  geometry offset{GEOMETRY_ZERO_PIXEL};
 };
 
 struct edge_values {
--- a/include/components/types.hpp
+++ b/include/components/types.hpp
@@ -78,9 +78,21 @@ struct size {
   unsigned int h{1U};
 };
 
+enum class unit_type { SPACE, POINT, PIXEL };
+
+struct size_with_unit {
+  unit_type type{unit_type::SPACE};
+  int value{0U};
+};
+
 struct side_values {
-  unsigned int left{0U};
-  unsigned int right{0U};
+  size_with_unit left{unit_type::SPACE, 0U};
+  size_with_unit right{unit_type::SPACE, 0U};
+};
+
+struct geometry_format_values {
+    double percentage{0.};
+    size_with_unit offset{unit_type::PIXEL, 0U};
 };
 
 struct edge_values {
@@ -141,11 +153,15 @@ struct bar_settings {
   struct size size {
     1U, 1U
   };
+
+  double dpi_x{0.};
+  double dpi_y{0.};
+
   position pos{0, 0};
   position offset{0, 0};
-  side_values padding{0U, 0U};
-  side_values margin{0U, 0U};
-  side_values module_margin{0U, 0U};
+  side_values padding{{unit_type::SPACE, 0U}, {unit_type::SPACE, 0U}};
+  side_values margin{{unit_type::SPACE, 0U}, {unit_type::SPACE, 0U}};
+  side_values module_margin{{unit_type::SPACE, 0U}, {unit_type::SPACE, 0U}};
   edge_values strut{0U, 0U, 0U, 0U};
 
   unsigned int background{0xFF000000};
@@ -158,7 +174,7 @@ struct bar_settings {
   std::unordered_map<edge, border_settings, enum_hash> borders{};
 
   struct radius radius {};
-  int spacing{0};
+  size_with_unit spacing{unit_type::SPACE, 0};
   string separator{};
 
   string wmname{};
--- a/include/components/types.hpp
+++ b/include/components/types.hpp
@@ -2,6 +2,7 @@
 
 #include <xcb/xcb.h>
 
+#include <cassert>
 #include <string>
 #include <unordered_map>
 
@@ -78,27 +79,36 @@ struct size {
   unsigned int h{1U};
 };
 
-enum class unit_type { SPACE, POINT, PIXEL };
+enum class space_type { SPACE, POINT, PIXEL };
+
+enum class size_type { POINT, PIXEL };
 
 namespace details {
   template <typename T>
   struct size_with_unit_impl {
-    unit_type type{unit_type::SPACE};
+    space_type type{space_type::SPACE};
     T value{0U};
   };
 }  // namespace details
 
-using size_with_unit = details::size_with_unit_impl<unsigned long int>;
-using ssize_with_unit = details::size_with_unit_impl<long int>;
+struct space_size {
+  space_type type{space_type::SPACE};
+  float value{0.f};
+};
+
+struct ssize_with_unit {
+  size_type type{size_type::PIXEL};
+  float value{0.f};
+};
 
 struct side_values {
-  size_with_unit left{unit_type::SPACE, 0U};
-  size_with_unit right{unit_type::SPACE, 0U};
+  space_size left{space_type::SPACE, 0_z};
+  space_size right{space_type::SPACE, 0_z};
 };
 
 struct geometry_format_values {
   double percentage{0.};
-  ssize_with_unit offset{unit_type::PIXEL, 0U};
+  ssize_with_unit offset{size_type::PIXEL, 0};
 };
 
 struct edge_values {
@@ -165,9 +175,9 @@ struct bar_settings {
 
   position pos{0, 0};
   position offset{0, 0};
-  side_values padding{{unit_type::SPACE, 0U}, {unit_type::SPACE, 0U}};
-  side_values margin{{unit_type::SPACE, 0U}, {unit_type::SPACE, 0U}};
-  side_values module_margin{{unit_type::SPACE, 0U}, {unit_type::SPACE, 0U}};
+  side_values padding{{space_type::SPACE, 0_z}, {space_type::SPACE, 0_z}};
+  side_values margin{{space_type::SPACE, 0_z}, {space_type::SPACE, 0_z}};
+  side_values module_margin{{space_type::SPACE, 0_z}, {space_type::SPACE, 0_z}};
   edge_values strut{0U, 0U, 0U, 0U};
 
   unsigned int background{0xFF000000};
@@ -180,7 +190,7 @@ struct bar_settings {
   std::unordered_map<edge, border_settings, enum_hash> borders{};
 
   struct radius radius {};
-  size_with_unit spacing{unit_type::SPACE, 0};
+  space_size spacing{space_type::SPACE, 0_z};
   string separator{};
 
   string wmname{};
--- a/include/modules/fs.hpp
+++ b/include/modules/fs.hpp
@@ -1,8 +1,10 @@
 #pragma once
 
+#include <utility>
+
 #include "components/config.hpp"
-#include "settings.hpp"
 #include "modules/meta/timer_module.hpp"
+#include "settings.hpp"
 
 POLYBAR_NS
 
@@ -25,7 +27,7 @@ namespace modules {
     int percentage_free{0};
     int percentage_used{0};
 
-    explicit fs_mount(const string& mountpoint, bool mounted = false) : mountpoint(mountpoint), mounted(mounted) {}
+    explicit fs_mount(string mountpoint, bool mounted = false) : mountpoint(std::move(mountpoint)), mounted(mounted) {}
   };
 
   using fs_mount_t = unique_ptr<fs_mount>;
@@ -61,11 +63,11 @@ namespace modules {
     vector<fs_mount_t> m_mounts;
     bool m_fixed{false};
     bool m_remove_unmounted{false};
-    size_with_unit m_spacing{unit_type::SPACE, 2U};
+    space_size m_spacing{space_type::SPACE, 2U};
 
     // used while formatting output
     size_t m_index{0_z};
   };
-}
+}  // namespace modules
 
 POLYBAR_NS_END
--- a/include/modules/fs.hpp
+++ b/include/modules/fs.hpp
@@ -61,7 +61,7 @@ namespace modules {
     vector<fs_mount_t> m_mounts;
     bool m_fixed{false};
     bool m_remove_unmounted{false};
-    int m_spacing{2};
+    size_with_unit m_spacing{unit_type::SPACE, 2U};
 
     // used while formatting output
     size_t m_index{0_z};
--- a/include/modules/meta/base.inl
+++ b/include/modules/meta/base.inl
@@ -133,7 +133,7 @@ namespace modules {
     bool no_tag_built{true};
     bool fake_no_tag_built{false};
     bool tag_built{false};
-    auto mingap = std::max(1_z, format->spacing);
+    auto mingap = std::max(1, format->spacing.value);
     size_t start, end;
     string value{format->value};
     while ((start = value.find('<')) != string::npos && (end = value.find('>', start)) != string::npos) {
--- a/include/modules/meta/base.inl
+++ b/include/modules/meta/base.inl
@@ -133,7 +133,7 @@ namespace modules {
     bool no_tag_built{true};
     bool fake_no_tag_built{false};
     bool tag_built{false};
-    auto mingap = std::max(1, format->spacing.value);
+    auto mingap = std::max(1UL, format->spacing.value);
     size_t start, end;
     string value{format->value};
     while ((start = value.find('<')) != string::npos && (end = value.find('>', start)) != string::npos) {
--- a/include/modules/meta/base.inl
+++ b/include/modules/meta/base.inl
@@ -133,7 +133,7 @@ namespace modules {
     bool no_tag_built{true};
     bool fake_no_tag_built{false};
     bool tag_built{false};
-    auto mingap = std::max(1UL, format->spacing.value);
+    auto mingap = std::max(1.f, format->spacing.value);
     size_t start, end;
     string value{format->value};
     while ((start = value.find('<')) != string::npos && (end = value.find('>', start)) != string::npos) {
--- a/include/utils/unit.hpp
+++ b/include/utils/unit.hpp
@@ -1,0 +1,40 @@
+#pragma once
+
+#include <string>
+
+#include "components/types.hpp"
+
+POLYBAR_NS
+
+namespace unit_utils {
+  template <typename ValueType, typename ReturnType = int>
+  ReturnType point_to_pixel(ValueType point, ValueType dpi) {
+    return static_cast<ReturnType>(dpi * point / ValueType(72));
+  }
+
+  template <typename ReturnType = int>
+  ReturnType size_with_unit_to_pixel(size_with_unit size, double dpi) {
+    if (size.type == unit_type::PIXEL) {
+      return size.value;
+    }
+
+    return point_to_pixel<double, ReturnType>(size.value, dpi);
+  }
+
+  inline string size_with_unit_to_string(size_with_unit size, double dpi) {
+    if (size.value > 0) {
+      if (size.type == unit_type::SPACE) {
+        return string(static_cast<string::size_type>(size.value), ' ');
+      } else {
+        if (size.type == unit_type::POINT) {
+          size.value = point_to_pixel(static_cast<double>(size.value), dpi);
+        }
+
+        return "%{O" + to_string(size.value) + "}";
+      }
+    }
+    return {};
+  }
+}  // namespace unit_utils
+
+POLYBAR_NS_END
--- a/include/utils/unit.hpp
+++ b/include/utils/unit.hpp
@@ -7,13 +7,13 @@
 POLYBAR_NS
 
 namespace unit_utils {
-  template <typename ValueType, typename ReturnType = int>
+  template <typename ValueType, typename ReturnType = long unsigned int>
   ReturnType point_to_pixel(ValueType point, ValueType dpi) {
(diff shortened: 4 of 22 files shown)


Lomadriel:
I fixed the unit tests. Should I add a test for the function `template <> geometry_format_values config::convert(string&& value) const`? (which does the string -> geometry_format_values conversion that isn't checked anymore)

Lomadriel:
I just remove some bugs (negative margin, padding...).
Normally this should also fix #1265.

Lomadriel:
> > DPI initialization is now in the constructor of the bar. However this initialization seems strange, the DPI are only computed if dpi < 0 is explicitly specified in the configuration (see below). If the DPI isn't specified in the configuration, the default value is 96. Is it the correct behavior?
> 
> Moving it into `bar.cpp` is fine. I don't know what the original intention was. But since that part of the code is over 2 years old, I'd say it's the expected behavior now ;)

If that's the expected behavior, maybe the documentation/wiki should be modified to state that any value <= 0 cause the DPI to be auto-detected. Moreover, this detection is bugged, since if I change the resolution of my screen (like when I duplicate my laptop screen and the projector is in 4/3), I have to relaunch polybar to update this value. (Maybe an x event should cause the value to be updated)

> I think we need to separate how we specify sizes and spacing.
> 
> Spacing can be in spaces, pixel, or points while size can be percentages, pixel, points or those combined in the offset format.
> 
> I first thought we can keep this together because it has lots of things in common (POINT and PIXEL), but I don't think this is worth it.

I agree, I tried to do my best but this is really confusing.

> The more I think about it, the less I am a fan of converting points to pixels only when the value in pixels is needed and passing around point values. I know this is actually what I suggested we do originally because the builder would be the only place where this conversion happens. But this would make things quite inflexible because if we ever extend our code we would need to do the conversion in all places that use those pixel values.

The following paragraph supposes that the DPI might become dynamic otherwise its content should be ignored.

For that point, I think this might be a problem if the DPI changes during the execution.
Since the "point" is here to have a DPI sensitive unit, it is quite strange to convert point sizes into pixel sizes when charging the configuration. If the DPI changes, we would need to update every sizes and it is quite expensive.
Maybe introducing a sort of `settings` singleton that will hold some important values (like the DPI) could solve the problem even if I don't like introducing singleton.
It will allow to have a space_size struct like that:
```c++
struct space_size {
  space_type type;
  // If type == space_type::pixel converts the value in pixel if necessary and returns it.
  // Otherwise returns the size in number of spaces.
  // No need to use float in the return type since pixel or space doesn't need them.
  ssize_t get_value() const {
    if (type == space_type::point)     
      return convert_to_pixel(settings::get_instance().get_dpi_x(), value);
    else
      return value;
  }
private:
  // to be sure that nobody uses it directly
  float value; // could be an union if you want since the get_value is the only place
                    // were the union will be used.
};
```
The geometry_size struct could be defined in a similar way with an enum type like that:
```c++
enum class geometry_size_type { PERCENTAGE, POINT, PIXEL }
```

> ```c++
> enum size_type { SPACE, PIXEL };
> ```
Do you want me to use a plain C enum or may I use an enum class? (I prefer using enum class to avoid global namespace pollution)

What do you think of this?

patrick96:
Everything within the bar except for the modules themselves is quite static. Polybar cannot dynamically change its size or position when changes to the monitor occur. All of this is handled through `screenchange-reload` which just reloads polybar whenever the screen changes.

It is conceivable that we want to make things more dynamic in the future. But even then, I would not convert to pixels in the builder but rather extend the `%{O}` tag to also work with points and then let the renderer handle it. Either way seems fine to me, thinking about it, doing this in the renderer may even be a little cleaner. This is up to you how you want to do it, I just don't want it to happen somewhere in the middle where we potentially have to do the conversion in multiple places.

Adding `get_value` might look very clean in the code that uses it, but it introduces a singleton *and* couples the concept of units for spacing purposes to the bar class. I want to avoid that.

>The geometry_size struct could be defined in a similar way with an enum type like that:
>```cpp
>enum class geometry_size_type { PERCENTAGE, POINT, PIXEL }
>```

Wouldn't `geometry_size_type` only have point and pixel? As far as I know, percentages cannot be used as offsets.

> Do you want me to use a plain C enum or may I use an enum class? (I prefer using enum class to avoid global namespace pollution)

You can use enum class. I just used enum because I'm more familiar with C ;)

Lomadriel:
> Everything within the bar except for the modules themselves is quite static. Polybar cannot dynamically change its size or position when changes to the monitor occur. All of this is handled through `screenchange-reload` which just reloads polybar whenever the screen changes.

I didn't know it worked that way, but that's seems logical with the fact that some of my modules are reset when I change my resolution.



> It is conceivable that we want to make things more dynamic in the future. But even then, I would not convert to pixels in the builder but rather extend the `%{O}` tag to also work with points and then let the renderer handle it. Either way seems fine to me, thinking about it, doing this in the renderer may even be a little cleaner. This is up to you how you want to do it, I just don't want it to happen somewhere in the middle where we potentially have to do the conversion in multiple places.

Maybe I can tried to do that to see if it is more clean. 

> > The geometry_size struct could be defined in a similar way with an enum type like that:
> > ```c++
> > enum class geometry_size_type { PERCENTAGE, POINT, PIXEL }
> > ```
> 
> Wouldn't `geometry_size_type` only have point and pixel? As far as I know, percentages cannot be used as offsets.

Yeah my bad, it is just point and pixel.

Lomadriel:
By the way, do you have an idea for naming the enums and structs associated with that:

> * spacing:
>   
>   * `POINT`: `float >= 0`
>   * `PIXEL`: `ssize_t`
>   * `SPACE`: `ssize_t`
> 
> * size:
>   
>   * `POINT`: `float`
>   * `PIXEL`: `size_t`

Each time I found a name it is either bad or already taken (like `size`)

patrick96:
Since we probably put these inside a namespace, naming is not such a big issue, but I think the following names are reasonable. Feel free to improve them.

```
enum class size_type { POINT, PIXEL }
enum class space_type { POINT, PIXEL, SPACE }

struct spacing {
  space_type type;
  ...
}

struct geometry {
  float percentage;
  ...
}
```

Lomadriel:
I just finished some modifications to take your review in account. Just as expected, there are two structures, one for spacing and the other for size.
The `%{O}` tag now supports units (`%{O100px}` and `%{O100pt}`).
There is also a commit that fixes the problem mentioned in #1527.

include/utils/unit.hpp
```diff
@@ -0,0 +1,40 @@
...
+  inline string size_with_unit_to_string(size_with_unit size, double dpi) {
+    if (size.value > 0) {
+      if (size.type == unit_type::SPACE) {
+        return string(size.value, ' ');
+      } else {
+        if (size.type == unit_type::POINT) {
+          size.value = point_to_pixel(static_cast<double>(size.value), dpi);
+        }
+
+        return "%{O" + to_string(size.value) + "}";
```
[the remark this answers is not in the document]
Lomadriel: I removed the tag from this function, however, in had to add `%{O}` tag in the controller. :/

src/components/controller.cpp
```diff
@@ -460,10 +460,30 @@ bool controller::process_update(bool force) {
   const bar_settings& bar{m_bar->settings()};
   string contents;
   string separator{bar.separator};
-  string padding_left(bar.padding.left, ' ');
-  string padding_right(bar.padding.right, ' ');
-  string margin_left(bar.module_margin.left, ' ');
-  string margin_right(bar.module_margin.right, ' ');
+
+  auto add_surrounding_tag = [](const space_size& space_size) -> std::string {
```
patrick96: This functionality is now implemented twice, once here and once in the builder. It probably makes sense to create a single function that generates either a number of spaces or an apropriate `%{O` tag.

I'm not sure about the apropriate place for this function. I generally would like to the builder be responsible, so a static function there may make sense. But this would couple the builder and the controller together but that may not be too bad since the controller also *builds* the module output.

src/components/parser.cpp
```diff
@@ -133,7 +133,7 @@ void parser::codeblock(string&& data, const bar_settings& bar) {
         break;
 
       case 'O':
-        m_sig.emit(offset_pixel{static_cast<int>(std::strtol(value.c_str(), nullptr, 10))});
+        m_sig.emit(offset{string{value}});
```
patrick96: I think the parser should parse the value of the offset tag and only send the renderer either a direct pixel value or a `struct geometry`. The latter probably makes more sense since the parser shouldn't have to bother with dpi.

src/components/renderer.cpp
```diff
@@ -300,13 +276,12 @@ void renderer::end() {
...
   }
 
-
   // For pseudo-transparency, capture the contents of the rendered bar and
   // composite it against the desktop wallpaper. This way transparent parts of
   // the bar will be filled by the wallpaper creating illusion of transparency.
   if (m_pseudo_transparency) {
     cairo_pattern_t* barcontents{};
-    m_context->pop(&barcontents); // corresponding push is in renderer::begin
+    m_context->pop(&barcontents);  // corresponding push is in renderer::begin
```
patrick96: Why? :thinking:

src/components/renderer.cpp
```diff
@@ -767,9 +755,14 @@ bool renderer::on(const signals::parser::reverse_colors&) {
...
-  m_log.trace_x("renderer: offset_pixel(%f)", evt.cast());
-  m_blocks[m_align].x += evt.cast();
+bool renderer::on(const signals::parser::offset& evt) {
+  m_log.trace_x("renderer: offset(%f)", evt.cast());
+
+  auto& x_offset = m_blocks[m_align].x;
+  auto offset_width = unit_utils::geometry_to_pixel(unit_utils::geometry_from_string(evt.cast()), m_bar.dpi_x);
+
+  draw_offset(x_offset, offset_width);
+  x_offset += offset_width;
```
patrick96: Using references like this can be a bit confusing. I think it is clearer if we access and modify `m_blocks[m_align].x` directly.

src/drawtypes/label.cpp
```diff
@@ -148,14 +148,23 @@ namespace drawtypes {
...
-      auto value = conf.get(section, key, 0U);
-      auto left = conf.get(section, key + "-left", value);
-      auto right = conf.get(section, key + "-right", value);
-      return side_values{static_cast<unsigned short int>(left), static_cast<unsigned short int>(right)};
+    const auto get_left_right = [&](string&& key) {
+      const auto parse_or_throw = [&](const string& key, space_size default_value) {
+        try {
+          return conf.get(section, key, default_value);
+        } catch (const std::exception& err) {
+          throw application_error(
```
patrick96: I really don't like that we have to handle malformed values in the places where they are accessed when we even use a `convert` function to parse them. There's not much we can do about that right now though :disappointed:. I'll probably takle this when we rework the convert functions in the new config parser

patrick96:
It would probably be useful if we also allow the use of these units in `format-padding`, `format-offset` etc. Those are set in `modules/meta/base.cpp`

include/modules/fs.hpp
```diff
@@ -25,7 +27,7 @@ namespace modules {
     int percentage_free{0};
     int percentage_used{0};
 
-    explicit fs_mount(const string& mountpoint, bool mounted = false) : mountpoint(mountpoint), mounted(mounted) {}
+    explicit fs_mount(string mountpoint, bool mounted = false) : mountpoint(std::move(mountpoint)), mounted(mounted) {}
```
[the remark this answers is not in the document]
Lomadriel: If the parameter is copied, it is better to let the copy to the caller. This allow the caller to move the object, if the parameter is a const ref the caller can't move the object.
It is recommended for a constructor (but not for a function or a method).

See https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html#modernize-pass-by-value for more information.

src/components/renderer.cpp
Lomadriel: Sorry, my IDE made this modification because of the clang format.
I forgot to remove that from the commit.
However this is the current setting in the clang format.
The current value is 2 (Google default).
Maybe the clang format should be modified as follow:
```
SpacesBeforeTrailingComments: 1
```

src/components/renderer.cpp
Lomadriel: I agree I don't know why I did that. This is ugly :scream:

Lomadriel: I agree I don't know why I did that. This is ugly :scream:

Lomadriel: Fixed by https://github.com/jaagr/polybar/pull/1671/commits/18dfa8ddaf802b1c62646d405cadccf59d6da13d

src/drawtypes/label.cpp
Lomadriel: I agree, a verification phase would be welcome

include/components/types.hpp
```diff
@@ -129,11 +149,15 @@ struct bar_settings {
...
+
+  double dpi_x{0.};
+  double dpi_y{0.};
+
   position pos{0, 0};
   position offset{0, 0};
-  side_values padding{0U, 0U};
-  side_values margin{0U, 0U};
-  side_values module_margin{0U, 0U};
+  side_values padding{{space_type::SPACE, 0_z}, {space_type::SPACE, 0_z}};
```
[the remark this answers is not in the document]
Lomadriel: Fixed by https://github.com/jaagr/polybar/pull/1671/commits/0e63878dafbac216b05708e9f314264e7b61f910

src/components/renderer.cpp
Lomadriel: I can remove that and add this line to the clang format if you want.

Lomadriel:
> The issue in #1700 is still not totally fixed. Offsets do produce spaces that have the proper background color, but the string I used in #1700 still does not look the same as with lemonbar:
> 
> ```
> %{O100}Filler%{B#ff0000 O100}First%{O200}Second%{O-100}
> ```
> 
> With this PR:
> 
> bg_after_fix
> 
> With lemonbar:
> 
> bg_lemonbar
> 
> You see, `Second` is not displayed with this PR:

Fixed by https://github.com/jaagr/polybar/pull/1671/commits/fa1719ad08a770ef3b25d65bd497b3aecd0663b4

Lomadriel:
> It would probably be useful if we also allow the use of these units in `format-padding`, `format-offset` etc. Those are set in `modules/meta/base.cpp`

If I am not wrong `format-padding` already support units.
https://github.com/jaagr/polybar/pull/1671/commits/ab19b3be526945857ed25b9c74935e3f71373d74 adds unit support for `format-offset`

src/components/parser.cpp
Lomadriel: Fixed by https://github.com/jaagr/polybar/pull/1671/commits/f2c170c6c69522461f808ba0a762430366dc192e

src/components/controller.cpp
Lomadriel: Fixed by https://github.com/jaagr/polybar/pull/1671/commits/f2c170c6c69522461f808ba0a762430366dc192e

Lomadriel:
I just added a commit that fixes #1814.
I can't make a separate PR since I modified the renderer :/
Sorry

patrick96:
I think you should make a separate PR and resolve the merge conflict once one of the two is merged.

Lomadriel:
> I think you should make a separate PR and resolve the merge conflict once one of the two is merged.

Commit removed. See #1817.

src/components/renderer.cpp
patrick96: Yes, please do :)

include/modules/meta/base.inl
```diff
@@ -133,7 +133,7 @@ namespace modules {
     bool no_tag_built{true};
     bool fake_no_tag_built{false};
     bool tag_built{false};
-    auto mingap = std::max(1_z, format->spacing);
+    auto mingap = std::max(1.f, format->spacing.value);
```
patrick96: I think this is the only area of concern. This mingap is used down below to determine how many whitespaces to remove if a tag wasn't built.

This creates problems when the spacing isn't actually whitespace but an offset.

I would solve this by just deleting this variable and down below fix the logic that makes sure that spacing is only added before a tag, if the tag was actually built.

The function could look like this:

```cpp
  template <typename Impl>
  string module<Impl>::get_output() {
    std::lock_guard<std::mutex> guard(m_buildlock);
    auto format_name = CONST_MOD(Impl).get_format();
    auto format = m_formatter->get(format_name);
    bool no_tag_built{true};
    auto tag_builder = make_unique<builder>(m_bar);
    size_t start, end;
    string value{format->value};
    while ((start = value.find('<')) != string::npos && (end = value.find('>', start)) != string::npos) {
      if (start > 0) {
        if (no_tag_built) {
          // If no module tag has been built we do not want to add
          // whitespace defined between the format tags, but we do still
          // want to output other non-tag content
          auto trimmed = string_util::ltrim(value.substr(0, start), ' ');
          if (!trimmed.empty()) {
            m_builder->node(move(trimmed));
          }
        } else {
          m_builder->node(value.substr(0, start));
        }
        value.erase(0, start);
        end -= start;
        start = 0;
      }
      string tag{value.substr(start, end + 1)};
      if (tag.empty()) {
        continue;
      } else if (tag[0] == '<' && tag[tag.size() - 1] == '>') {
        if (CONST_MOD(Impl).build(tag_builder.get(), tag)) {
          m_builder->space(format->spacing);
          m_builder->append(tag_builder->flush());
          no_tag_built = false;
        }
        else {
          tag_builder->flush();
        }
      }
      value.erase(0, tag.size());
    }

    if (!value.empty()) {
      m_builder->append(value);
    }

    return format->decorate(&*m_builder, m_builder->flush());
  }
```

This isn't the best solution, but the best one where we don't need to massively modify the builder.

Now we can also completely remove the `remove_trailing_space` methods from the builder.

src/modules/meta/base.cpp
```diff
@@ -14,10 +14,10 @@ namespace modules {
       builder->flush();
       return "";
     }
-    if (offset != 0) {
+    if (offset.value != 0) {
       builder->offset(offset);
     }
-    if (margin > 0) {
+    if (margin.value > 0) {
```
patrick96: Same here, a negative margin should be allowed (unless it is of type `SPACE`)

src/components/builder.cpp
```diff
@@ -614,5 +618,23 @@ void builder::tag_close(attribute attr) {
       break;
   }
 }
+string builder::add_surrounding_tag(const space_size& space) {
```
patrick96: `get_format_string` may be a more descriptive name since this function doesn't actually add a tag if we use spaces.

frebib:
Can this be rebased against master please

frebib:
I squashed and rebased this against master (I tried to rebase without squashing but kept screwing up the history, so figured I didn't need to keep it anyway just to build and run it)
https://github.com/frebib/polybar/commit/7385b894a795058cd4e0db14c9320b3e8fda5c66

I rely on this fork to have polybar even remotely usable when changing DPI. Please consider reviving it @Lomadriel

fabianschuiki:
Any way to help get this PR merged? E.g. implement requested code changes? I'd really love to see these units to improve HiDPI support :+1:

frebib:
Until now I've been rebasing this PR onto master, but https://github.com/polybar/polybar/commit/fd556525a892d62d8a35595e0fb5cb648253d31c was a big change that will not be a trivial change to rebase onto. I may or may not attempt to fix this up again, depending on how hard it seems to be

patrick96:
@frebib If you do attempt to resolve the conflicts and want to help get this merged, a PR would be very much appreciated :)

And of course, ask if you need help.

I think most conflicts should be quite straighforward (just tedious to get the changes from both sides in), but in the renderer and parser, it might be more difficult because there we actually changed the places where the offset handling happens.

frebib:
@patrick96 I can certainly give it a go, but I'm not at all familiar with C++ or much of the syntax/language features.
If I were to rebase this, it'd be squashed into a single commit- hopefully that's ok.
There were a few unaddressed comments from what I remember about the implementation which I should probably address at the same time.
In use I also noticed that there are still some glaring issues with unit support particularly in the tray area. I managed to hack in support for most of it, but the padding between the icons is still not correctly scaled. Hopefully that shouldn't be too tricky to solve.

patrick96:
>  If I were to rebase this, it'd be squashed into a single commit- hopefully that's ok.

Yeah, that's fine. This would have happened with this PR anyway, I think.

> I can certainly give it a go, but I'm not at all familiar with C++ or much of the syntax/language features.

In that case, this might get tricky. Let me know, if you need any help.

> In use I also noticed that there are still some glaring issues with unit support particularly in the tray area. I managed to hack in support for most of it, but the padding between the icons is still not correctly scaled. Hopefully that shouldn't be too tricky to solve.

Yeah, I don't think this PR touches the tray at all. Feel free to leave out units in the tray if that's easier. First priority should be to get the overall framework working.

frebib:
I gave it a go. Seems to compile & run the same as the old builds did for me.
See #2394

Issue #1630, opened 2019-01-27
prevent module/bspwm to create separators for empty workspaces


dannycolin:
## Describe the issue

**Actual behavior:**
*A clear and concise description of what the bug is.*

module/bspwm still create a separator for every workspaces even for those that have an empty label. Here's an excerpt of my config file. This behavior create an unwanted gap between workspaces' labels.

**Expected behavior:**
*A clear and concise description of what you expected to happen.*

module/bspwm shouldn't create a separator for empty workspaces when `label-empty` is set to nothing. 

**Was it working before?**
* Did you also experience this bug in an earlier version of polybar?
* If no, what was the last version where this worked correctly?

I wasn't using bspwm prior to the version of polybar I'm actually using.

## To Reproduce
*A minimal but complete config with which the problem occurs:*

* add those settings:

```dosini
label-empty =
label-focused = %name%
label-occupied = %name%
label-separator = " "
```

* Open an application on workspace 1
* Move to workspace 3 or greater

*List any other steps needed to reproduce the issue besides starting polybar with the config you posted above.*

## Polybar Log
*Post everything polybar outputs to the terminal when you run it and the issue occurs below*

There's no error in the output.

## Screenshots
*If applicable, add screenshots to help explain your problem.*

[screenshot link](https://www.dannycolin.com/screenshots/polybar-bspwm-issue.png)

## Environment:
* WM: bspwm
* Output of `polybar -vvv`:

polybar 3.2.1

Features: -alsa -curl +i3 -mpd +network +pulseaudio +xkeyboard

X extensions: +randr (+monitors) -render -damage -sync -composite +xkb +xrm +xcursor

Build type: RelWithDebInfo
Compiler: /usr/bin/clang++
Compiler flags:  -Wall -Wextra -Werror -O2 -pedantic -pedantic-errors -Wno-error=parentheses-equality -Wno-zero-length-array
Linker flags: 

## Additional context
*Add any other context that you think is necessary about the problem here.*

patrick96:
Oh, that's gonna be a pain to implement. But you're right, we shouldn't show separators for empty labels. The main difficulty will come from the logic behind when to display separators because separators are displayed in between workspaces so it isn't as easy as to just say don't draw separators for empty labels.

dannycolin:
@patrick96 would it be easier to add `label-{focused|occupied|urgent|empty}-padding` like in module/i3?

I mean with that there's no more needs to have a separator and verify if a workspace label is empty.
[...]

PR #1673, opened 2019-03-04
Add custom ellipsis string #1195

Lomadriel:
Add feature requested in #1195.

- add `label-*-ellipsis-string` to specify a custom string. 
- add `deprecated_type_changed` helper in `config.hpp` (display a warning if old key is used and recommends the new key which has a different type (here bool -> string)
- update unit test and add one to check UTF-8 chars.

I also added a check that throw an exception if `label-*-ellipsis = true` and the string is empty. However I'm not sure it respects your way of doing things.


patrick96:
Oh wow, you have been busy :smiley: 

I have actually implemented this [before](https://github.com/patrick96/polybar/commits/feat/ellipsis) (059c69e6256e6e2bb99f648dd6d89f320c117a26), but I never opened a PR because I was conflicted about how to handle failure modes. So I think you can check there to see if you are doing more or less the same. But it looks good at first glance.

I don't have a lot of time right now and I'm also going away for a few days, so you'll have to wait a bit for me to review this and all your other PRs as well. 

Thank you very much for helping out, I'm looking forward to getting these features merged.

Lomadriel:
> I have actually implemented this before (059c69e), but I never opened a PR because I was conflicted about how to handle failure modes.

Yeah I was also conflicted about how to handle failure modes. In my code I used exception like the old check you removed, you used `logger::make().err(…)`.

 
> I don't have a lot of time right now and I'm also going away for a few days, so you'll have to wait a bit for me to review this and all your other PRs as well.

Okay no problem. Are there feats/bugs that you want me to try to implement/fix (my exam week is over)? I am still new to polybar, but I'll do my best. ^^

patrick96:
There are a few bugs that you could work on:

* #568, there has been some refactoring in #1213, maybe you could also incorporate that.
* #428
* #702, you would need to reproduce this first.
* #916, Ignore the jgmenu part, the underlying issue is that click commands are not executed if they produce output. See also [Known Issues](https://github.com/jaagr/polybar/wiki/Known-Issues#click-commands-are-not-executed)
* #1282. Here we need to make sure that we carry over the token properties of `%percentage-cores%` to `%percentage-core[0-9]+%`
* #1180, this is also something you need to reproduce first.
* #1081 
* #1630 


Other things you could work on:
* #1127 
* #1164 

I think right now we should focus on bug fixes and refactoring to make the code more maintainable. Once you are somewhat familiar with the codebase, you could also try your luck on refactoring.

Thanks a ton for helping out!!

src/components/builder.cpp
```diff
@@ -550,10 +550,10 @@ string builder::get_label_text(const label_t& label) {
   size_t maxlen = label->m_maxlen;
 
   if (maxlen > 0 && string_util::char_len(text) > maxlen) {
-    if (label->m_ellipsis) {
-      text = string_util::utf8_truncate(std::move(text), maxlen - 3) + "...";
-    }
-    else {
+    if (!label->m_ellipsis.empty()) {
+      text = string_util::utf8_truncate(std::move(text), maxlen - string_util::char_len(label->m_ellipsis)) +
```
patrick96: This code also works if ellipsis is empty (we truncate 0 characters). We don't need the extra check

include/components/config.hpp
```diff
@@ -162,6 +162,23 @@ class config {
     return default_value;
   }
 
+  /**
+   * Attempt to load value using the deprecated key name. If successful show a
+   * warning message. If it fails, the default value is returned.
+   * This version is to by used when the new key doesn't have the same type as the old one.
```
patrick96:
```suggestion
   * This version is to be used when the new key doesn't have the same type as the old one.
```

src/drawtypes/label.cpp
```diff
@@ -218,13 +218,21 @@ namespace drawtypes {
...
 
     size_t maxlen = conf.get(section, name + "-maxlen", 0_z);
-    bool ellipsis = conf.get(section, name + "-ellipsis", true);
+    bool has_ellipsis = conf.has(section, name + "-ellipsis");
+    bool ellipsis = conf.deprecated_type_changed(section, name + "-ellipsis", name + "-ellipsis-string", true);
+    string ellipsis_str = conf.get(section, name + "-ellipsis-string", "..."s);
 
-    if(ellipsis && maxlen > 0 && maxlen < 3) {
+    if (has_ellipsis && ellipsis && ellipsis_str.empty()) {
+      throw application_error(
```
patrick96: If `-ellipsis-string` is set, `-ellipsis` should probably be ignored.

Lomadriel:
Fixed.
At the moment, the behavior is the following:
 - If `label-ellipsis-string` exists then `label-ellipsis` is ignored.
 - The new way to disable ellipsis is `label-ellipsis-string = ""` or equivalent.
 - The default value of the key `label-ellipsis-string` is `...`.

Examples:

```ini
label-maxlen = 5
label-ellipsis = false // ignored since `label-ellipsis-string` exists + depreciation warning
label-ellipsis-string = "…"
```

```ini
label-maxlen = 5
label-ellipsis-string = "" // same as `label-ellipsis = false`
```

```ini
label-maxlen = 5
label-ellipsis-string = "…" // ellipsis activated
```

src/components/builder.cpp
Lomadriel: Fixed by https://github.com/polybar/polybar/pull/1673/commits/d2a6acbdfafdb6472993573a921f04deb8f261c7

src/drawtypes/label.cpp
Lomadriel: Fixed by https://github.com/polybar/polybar/pull/1673/commits/0b005696146b6f5d722cf7c7527ef5b3aa65a038

PR #1675, opened 2019-03-06
Fix compilation with clang8 #1674

### LINKED ISSUE #1674: Fail to build with LLVM8/Clang8.
## Build Process

FreeBSD [Port](https://svnweb.freebsd.org/ports/head/x11/polybar/).

[Complete build log](http://package18.nyi.freebsd.org/data/headamd64PR236062-default/2019-03-01_06h42m53s/logs/errors/polybar-3.3.0.log).

```
--- Description ---
$ poudriere jail -cj clang8 -v projects/clang800-import -m svn+https
$ poudriere testport -j clang8 x11/polybar
[...]
In file included from src/components/bar.cpp:22:
In file included from include/x11/tray_manager.hpp:15:
include/x11/tray_client.hpp:18:16: error: explicitly defaulted copy assignment
operator is implicitly deleted [-Werror,-Wdefaulted-function-deleted]
  tray_client& operator=(tray_client& c) = default;
               ^
include/x11/tray_client.hpp:38:15: note: copy assignment operator of
'tray_client' is implicitly deleted because field 'm_connection' is of
reference type 'polybar::v3_3_0::connection &'
  connection& m_connection;
              ^
1 error generated.
```

IDK C++ but I've asked a friend:

> To fix it would either need to write the assignment operator (explicitly *also* copying the reference), or change the type of connection from reference to something that is implicitly copyable (e.g. pointer, some kind of reference wrapper, ...)

Thank you.

Lomadriel:
This PR fixes #1674 by removing the copy ctor and assignment operator of `tray_client`.

Copy assignment was marked as defaulted however as stated in the C++ standard:
> A defaulted copy/move assignment operator for class X is defined as deleted if X has:
> - […]
> -  non-static data member of reference type, or
> - […]
>
> § 12.8

Since a reference can't be reassigned, I just marked this operator as explicitly deleted.

The copy ctor was also defaulted and it is well-formed but the dtor unembed the window so copying that class may cause bugs, so I also marked it as explicitly deleted. Another solution if copying this class is needed is to do the "unembeding" in a custom deleter of `m_xembed`.

### DIFF
--- a/include/x11/tray_client.hpp
+++ b/include/x11/tray_client.hpp
@@ -14,8 +14,8 @@ struct xembed_data;
 class tray_client {
  public:
   explicit tray_client(connection& conn, xcb_window_t win, unsigned int w, unsigned int h);
-  tray_client(const tray_client& c) = default;
-  tray_client& operator=(tray_client& c) = default;
+  tray_client(const tray_client& c) = delete;
+  tray_client& operator=(tray_client& c) = delete;
 
   ~tray_client();
 


patrick96:
Oh, so because `m_connection` is of reference type, the move assignment operator is implicitly deleted?

Lomadriel:
> Oh, so because `m_connection` is of reference type, the move assignment operator is implicitly deleted?

Yeap, the move assignment operator is also implicitly deleted. Moreover there is no move ctor because  there is a destructor declared. So this class is non-movable and non-copyable.
The move constructor can be declared and implemented without causing any problem, but it is needed to add a condition to not free the resource in the dtor.

PR #1680, opened 2019-03-07
Fix broken pipe in command

### LINKED ISSUE #916: Click command not executed when it produces output.
When creating a polybar "button" for [jgmenu](https://github.com/johanmalm/jgmenu), I came upon the issue described below. I'd be grateful for a steer on if it is the intended behaviour and if there is anything I could do to avoid it.

I use
  * polybar v3.1.0 (build from Arch AUR two days ago)
  * i3 v4.14.1
  * Arch Linux

Using the default polybar config file (/usr/share/doc/polybar/config), I added:
`modules-left = jgmenu bspwm i3`
and

```
[module/jgmenu]
type = custom/script
label-padding = 2
interval = 10
exec = echo "menu"
click-left = "jgmenu_run &"
```

I found that "jgmenu_run" and "jgmenu" work in terms of launching jgmenu. However, when running jgmenu as a long-running application (default), I don't get polybar back (as described on wiki, so nothing strange about that).

However, "jgmenu_run &" and "jgmenu &" do not launch jgmenu. This is what's peculiar to me. These commands work from the terminal with the "&".

I found that the following works:
`click-left = "jgmenu_run >/dev/null 2>&1 &"`

The issue can be re-created without jgmenu, by setting "click-left = ~/test.sh &" and then doing the following:

```
cat <<EOF >~/test.sh
#!/bin/sh
echo "foo"
dmenu_run
EOF
```

If the 'echo "foo"' line is removed, it works.

Lomadriel:
Fix #916.
As I said in #916, the bug was:
1.    The child replaces STDIN, STDERR and STDOUT by the pipe.
2.    The child closes the unused pipes
3.    Since the command (`cmd &`) does a fork the `wait` function returns immediately while the process is still alive.
4.    The destructor of the command closes the pipe (parent side)
5.    The new process writes into STDOUT (alias the closed pipe)
6.    The new process get a SIGPIPE and is killed.

I modified the class command, to create "two classes", one that redirects the output with pipes, and one that doesn't redirect output.
I refactored the rest of the code in consequence.

EDIT: At the moment, users still need to use `cmd &` to avoid blocking polybar. It may be interesting to add a parameter to the constructor to detach the child.

EDIT: Ref #1717

### DIFF
--- a/src/utils/command.cpp
+++ b/src/utils/command.cpp
@@ -37,6 +37,7 @@ int command<output_policy::IGNORED>::exec(bool wait_for_completion) {
 
   if (process_util::in_forked_process(m_forkpid)) {
     setpgid(m_forkpid, 0);
+    process_util::redirect_process_output_to_dev_null();
     process_util::exec_sh(m_cmd.c_str());
   } else {
     if (wait_for_completion) {
--- a/src/utils/command.cpp
+++ b/src/utils/command.cpp
@@ -3,6 +3,7 @@
 #include <csignal>
 #include <cstdlib>
 #include <utility>
+#include <utils/string.hpp>
 
 #include "errors.hpp"
 #include "utils/command.hpp"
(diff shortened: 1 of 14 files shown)


src/utils/command.cpp
```diff
@@ -18,7 +19,95 @@
...
+ * Execute the command
+ */
+int command<output_policy::IGNORED>::exec(bool wait_for_completion) {
+  if ((m_forkpid = fork()) == -1) {
+    throw system_error("Failed to fork process");
+  }
+
+  if (process_util::in_forked_process(m_forkpid)) {
+    setpgid(m_forkpid, 0);
+    process_util::exec_sh(m_cmd.c_str());
```
patrick96, line 40: Here, the forked process inherits polybar's stdin, stdout, and stderr. So any output will become part of polybar's output. Is this intended?

For one, it makes sense that people can see the output of their script for debugging, but the polybar output may get spammed with command outputs. I'm not sure what the best approach is here.

tests/unit_tests/utils/command.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+using namespace polybar;
+
+TEST(Command, status) {
+  // Test for command<output_policy::IGNORED>::exec(bool);
+  {
+    auto cmd = command_util::make_command<output_policy::IGNORED>("echo polybar > /dev/null");
+    int status = cmd->exec();
+
+    ASSERT_EQ(status, EXIT_SUCCESS);
```
patrick96, line 14: Please use `EXPECT_*` everywhere so that we don't see only a part of the test fail. Use `ASSERT_*` if the rest of the code would otherwise have undefined behavior (for example for NULL checks).

patrick96:
> EDIT: At the moment, users still need to use cmd & to avoid blocking polybar. It may be interesting to add a parameter to the constructor to detach the child.

It may be worth reconsidering the interface we expose for executing commands.
For many use-cases we don't need all the functionality of a `command` instances. In most cases we want to just execute a command and either forget about it (ignore output, fork into background, don't call `wait`) or execute it synchronously and get its output and exit code. Here having to create a `command` object is a bit overkill.

The only place `command` is really needed is in the script module and there the interface it provides is not wide enough, it still has to manually deal with pipe file descriptors.

I believe that we could do without the `command` class. All the use cases except the script module could be implemented in two functions in `process.cpp` and the script module can either do the forking and redirecting pipes itself or we create a well defined interface that covers everything the script module needs.

For spawning commands in the background without blocking polybar, we can take an approach similar to what jgmenu does:

https://github.com/johanmalm/jgmenu/blob/e34a804799a4dcafe4346cd7a9eec213013c5fae/src/spawn.c#L41-L69

Lomadriel:
I agree, the only place where we should use the command object is in indeed in the script module.

Lomadriel:
Rebased on master

src/utils/command.cpp
patrick96: I think this is the only point that still needs to be addressed.
I have thought about this a bit and I believe processes spawned like this should have their stdout and stderr redirected to `/dev/null`. If they don't do that, they will forever output in the shell running polybar (even after polybar is closed). This is especially annoying for long-running programs.

I think voiding the output like this is enough:

```c
int fd_null = open("/dev/null", O_WRONLY);
dup2(fd_null, STDOUT_FILENO);
close(fd_null);
```

And the same for stderr (with error checks of course).

patrick96:
> I agree, the only place where we should use the command object is in indeed in the script module.

We should maybe plan a refactor after this PR is merged. It will rip out a lot of redundant code and should make running commands a lot cleaner.

src/utils/process.cpp
```diff
@@ -23,6 +25,18 @@ namespace process_util {
     return pid == 0;
   }
 
+  void redirect_process_output_to_dev_null() {
+    auto redirect = [](int fd_to_redirect) {
+      int fd = open("/dev/null", O_WRONLY);
+      if (fd < 0 || dup2(fd, fd_to_redirect) < 0)
+        throw system_error("Failed to redirect process output");
```
patrick96, line 32: Please use curly brackets here

Lomadriel: Done

Lomadriel:
:tada:

Issue #1046, opened 2018-02-22
Battery not updating unless I refresh i3


smaakmaker:
I sometimes get the feeling that my battery module isn't printing the correct value for my battery. Then, when I refresh i3, it changes dramatically and displays it correctly again. Has anyone else had this issue? I'll admit it's a minor one, but it's still something I'd fix instead of running out of juice while polybar says it's full.

Thanks

patrick96:
Are you running fedora?

Try the following, while polybar is running and you suspect the battery value is not updated, run
```bash
cat /sys/class/power_supply/BAT0/charge_now || cat /sys/class/power_supply/BAT0/energy_now
```
where you replace `BAT0` with whatever you have put as the `battery` parameter in your config.
And see if the value in polybar the immediately updates without reloading i3.

patrick96:
Closing due to inactivity

f0s3:
I am facing this exact issue being on arch linux with i3wm.

Issue #469, opened 2017-02-28
battery module won't update after suspend


reedts:
The battery module reports 0% charge and won't update (even on plug/unplug of AC) after resuming from suspend. The log output shows following error:
````
error: module/battery: Error while creating inotify watch (what: Failed to attach inotify watch (reason: No such file or directory))
````
I'm using default polling interval.

Polybar version is 3.0.4 on my Arch Linux (4.9.11-1).

**Comment by jaagr** [comment]
Do the files exist after resuming?

~~~ sh
$ ls -1 /sys/class/power_supply/*/
~~~

**Comment by reedts** [comment]
Yes, everything as expected:
````
ls -1 /sys/class/power_supply/
AC@
BAT0@
````
Both symlinks are valid.

**Comment by jaagr** [comment]
Run the command I wrote to list the contents of those two directories.

**Comment by reedts** [comment]
Here you go:
````
% ls -1 /sys/class/power_supply/*/
/sys/class/power_supply/AC/:
device@
online
power/
subsystem@
type
uevent

/sys/class/power_supply/BAT0/:
alarm
capacity
capacity_level
cycle_count
device@
energy_full
energy_full_design
energy_now
manufacturer
model_name
power/
power_now
present
serial_number
status
subsystem@
technology
type
uevent
voltage_min_design
voltage_now
````
If I can do anything else to help you, please let me know!

**Comment by jaagr** [comment]
Does it work after restarting the bar?

**Comment by reedts** [comment]
Yes, after restarting the bar, everythings back to normal.

**Comment by l4ngl3y** [comment]
Yes It old problem.

**Comment by MrAnno** [comment]
I had the same problem.
While trying to attach gdb to the problematic process, I accidentally started a second instance from polybar, which automagically fixed the battery module in the first process as well.

The only thing I was able to capture was the lack of inotify watches in the first process:

```
❯ ls -al /proc/5218/fd

0 -> /dev/null
1 -> /dev/null
10 -> socket:[55080]
11 -> pipe:[57476]
12 -> pipe:[57476]
13 -> pipe:[57477]
14 -> pipe:[57477]
15 -> socket:[57479]
16 -> anon_inode:[eventfd]
17 -> anon_inode:[eventfd]
2 -> /home/anno/.xsession-errors
3 -> /dev/urandom
4 -> socket:[56570]
5 -> socket:[56587]
6 -> pipe:[55072]
7 -> pipe:[55072]
8 -> socket:[55075]
9 -> socket:[55078]
```

PR #1681, opened 2019-03-08
Fix difference between real and displayed value in backlight module

### LINKED ISSUE #1180: Backlight not updating correctly
When I change my brightness, it looks like polybar is updating the stored value, but not actually printing it until I change the brightness again. This results in a value that is behind by an update.

Say, if my brightness is 45%. I increment it by 5%, and I see my screen get brighter. However, the value stays at 45%. 
brightness(+5%) = Actual: 55%, Polybar: 50%
brightness(+5%) = Actual: 60%, Polybar: 55%
brightness(-20%) = Actual: 40%, Polybar: 60%
brightness(+10%) = Actual: 50%, Polybar 40%

Restarting polybar prints the correct value.

I am running a Lenovo Yoga 2 Pro on Fedora 27. 
I'm using internal/backlight, because xbacklight is not supported on my system. 
`card = intel_backlight`
I have a bash script that updates `/sys/class/backlight/intel_backlight/brightness` with an x% change, bound to function keys via i3

This is pretty unimportant, but it's a silly bug and I figured I'd file an issue. I may try digging through the source to resolve this after I'm finished setting up this laptop.

Lomadriel:
Fix #1180.

WARNING :warning: :warning: : This PR is not only a bugfix, it also add a new feature in polybar :arrow_right: udev meta module.

As explained in #1180. Inotify event aren't guaranteed to work on sysfs (/sys/ /proc/ /dev/ …). So we can't use inotify to watch in theses directories. (citation of the man page below)

> Inotify reports only events that a user-space program triggers through the filesystem API. As a result, it does not catch remote events that occur on network filesystems. (Applications must fall back to polling the filesystem to catch such events.) Furthermore, various pseudo-filesystems such as /proc, /sys, and /dev/pts are not monitorable with inotify.

So as the man said, the filesystem can be polled to catch such events, but there is an API to monitor device :arrow_right: udev.
~~Since udev is quite "standard" among linux distributions and freebsd, I don't think this dependency will cause any problem.~~

~~If polybar uses this API, it could be easy to solve #469 #839 #1046 #1181 that may be linked to the aforementioned problem with inotify.~~

EDIT: To fix the power problems, it is more UPower that should be used (with dbus) sorry my bad.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ addons:
         - libxcb-util0-dev
         - python-xcbgen
         - xcb-proto
+        - libudev-dev
       - &optional_deps
         - libxcb-xkb-dev
         - libxcb-cursor-dev
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ A compiler with C++14 support ([clang-3.4+](http://llvm.org/releases/download.ht
 - `xcb-proto`
 - `xcb-util-image`
 - `xcb-util-wm`
+- `libudev`
 
 **Optional dependencies:**
 - `xcb-util-cursor` *required for the `cursor-click` and `cursor-scroll` settings*
--- a/cmake/03-libs.cmake
+++ b/cmake/03-libs.cmake
@@ -6,6 +6,7 @@ find_package(Threads REQUIRED)
 list(APPEND libs ${CMAKE_THREAD_LIBS_INIT})
 
 querylib(TRUE "pkg-config" cairo-fc libs dirs)
+querylib(TRUE "pkg-config" libudev libs dirs)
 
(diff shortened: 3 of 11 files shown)


Lomadriel:
I will fix this without using UDev.

patrick96:
We may be able to use this udev implementation in other places in the future ;)

Lomadriel:
> We may be able to use this udev implementation in other places in the future ;)

Yeap, but we will need to separate OS implementation since FreeBSD doesn't support UDev (we must use devd).
But using UDev and devd may should several problems in polybar.

patrick96:
Once we have a solid interface for separating data collection from the part that formats it (#1529), we could tackle this.

PR #1683, opened 2019-03-09
Fix animation framerate

### LINKED ISSUE #568: animation-charging: frame time unstable and constantly differs.
```
[module/battery]
type = internal/battery

; This is useful in case the battery never reports 100% charge
full-at = 99

; Use the following command to list batteries and adapters:
; $ ls -1 /sys/class/power_supply/
battery = BAT0
adapter = AC

format-charging = <animation-charging> <label-charging>
format-discharging = <ramp-capacity> <label-discharging>

label-charging = %percentage%%
label-discharging = %percentage%%
; Only applies if <ramp-capacity> is used
ramp-capacity-0 = ï‰„
ramp-capacity-1 = ï‰ƒ
ramp-capacity-2 = ï‰‚
ramp-capacity-3 = ï‰
ramp-capacity-4 = ï‰€

; Only applies if <animation-charging> is used
animation-charging-0 = ï‰„
animation-charging-1 = ï‰ƒ
animation-charging-2 = ï‰‚
animation-charging-3 = ï‰
animation-charging-4 = ï‰€
; Framerate in milliseconds
animation-charging-framerate = 500
```

Title speaks for itself: animation is weird, it can wait more that delay for the first frame, and then run trough remaining frames faster than the blink of an eye, or be almost normal.
In other words, time that 1 frame stays drawn is different every time.

My version of polybar is:

```
$ polybar --version
polybar 3.0.5

Features: +alsa +curl -i3 -mpd +network
```

My OS is Arch Linux, with Intel Graphics (only).
I can't currently provide any screencast, but will do so if I have possibility.

Lomadriel:
Fix #568.

As it may be seen in the code, the animation need to be synchronized with the thread that calls the `broadcast` function in order to keep a stable framerate.
To synchronize the thread and the animation, I moved the "animation update" in the `subthread` method.
Although this fix the bug, I don't like to move the "animation update" responsibility to another class. The modules should not have to update the animation. Moreover it could be easy to forget the call to `increment` that updates the animation. (when the code is refactored or simply when a new module will need this class).

This is why I also implemented a version with an `animation_manager`, but since this is a big refactor, I put this version in another branch.
=> https://github.com/Lomadriel/polybar/tree/refactor/animation
This commit (https://github.com/Lomadriel/polybar/commit/1dc8949bf53389143c757aaa63753c42ad501a2e) is the diff between this PR and the branch `refactor/animation`

In this version a module only store an `animation_manager`, that manages the animations. Once launched the `animation_manager` will process as follow:

1. Call a functor to let the module choose the current animation.
2. Sleep if no animation has been chosen.
3. Otherwise, increments the animation
4. Call a callback to let the module do whatever needs to be done (mostly call `broadcast`)
5. Sleep until the remaining time between two steps of the animation is passed.

As it really changes the way animations work, I let you choose which version you want.

TEAM EDIT: Closes #1213

### DIFF
--- a/include/drawtypes/animation.hpp
+++ b/include/drawtypes/animation.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <atomic>
 #include <chrono>
 
 #include "common.hpp"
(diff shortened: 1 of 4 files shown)


Lomadriel:
I fixed a bug and replaced `system_clock` by `steady_clock`

Lomadriel:
Conflict fixed

patrick96:
Thanks a lot :rocket:

PR #1684, opened 2019-03-10
Improves date module (label for alternative and timezone support)

### LINKED ISSUE #596: Select time zone for date module

### LINKED ISSUE #1293: [module/date] Zone and UTC feature support request

Lomadriel:
Fix #596.

A new attempt to implement this feature. This PR fixes the concurrency problem in #1353 by using a static mutex in the date class. Since the date class is the only one to support timezone, this shouldn't be a problem. However if a global clock is wanted I can add an utility function/class that will do the job.

TEAM EDIT: Closes #1293

### DIFF
--- a/include/common.hpp
+++ b/include/common.hpp
@@ -44,4 +44,6 @@ constexpr size_t operator"" _z(unsigned long long n) {
   return n;
 }
 
+[[deprecated("Use env_util::env_guard (thread safe version)")]] int setenv(const char*, const char*, int);
+
 POLYBAR_NS_END
--- a/include/modules/date.hpp
+++ b/include/modules/date.hpp
@@ -4,9 +4,7 @@
 #include "modules/meta/timer_module.hpp"
 
 #include <atomic>
-#include <ctime>
-#include <iomanip>
-#include <iostream>
+#include <sstream>
 
 POLYBAR_NS
 
@@ -39,9 +37,6 @@ namespace modules {
 
     string m_date;
     string m_time;
-
-    static std::atomic_bool s_timezone_activated;
-    static mutex s_timezone_mutex;
 
     // Single stringstream to be used to gather the results of std::put_time
     std::stringstream datetime_stream;
--- a/include/modules/date.hpp
+++ b/include/modules/date.hpp
@@ -15,18 +15,23 @@ namespace modules {
 
     bool update();
     bool build(builder* builder, const string& tag) const;
+    string get_format() const;
 
    protected:
     bool input(string&& cmd);
 
    private:
+    static constexpr auto FORMAT_ALT = "format-alt";
+
     static constexpr auto TAG_LABEL = "<label>";
+    static constexpr auto TAG_LABEL_ALT = "<label-alt>";
     static constexpr auto EVENT_TOGGLE = "datetoggle";
 
     // \deprecated: Use <label>
     static constexpr auto TAG_DATE = "<date>";
 
     label_t m_label;
+    label_t m_label_alt;
 
     string m_dateformat;
     string m_dateformat_alt;
--- a/include/modules/date.hpp
+++ b/include/modules/date.hpp
@@ -3,9 +3,10 @@
 #include "modules/meta/input_handler.hpp"
 #include "modules/meta/timer_module.hpp"
 
-#include <iostream>
-#include <iomanip>
+#include <atomic>
 #include <ctime>
+#include <iomanip>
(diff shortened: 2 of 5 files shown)


Lomadriel:
I just realized that `setenv` is not thread safe at all (I thought that `setenv` was thread safe if the modified environment variables were different but this is not the case).
I will add an utility function tomorrow.

Lomadriel:
Okay so I added a `env_guard` class that behave like a `std::lock_guard`, in the constructor the environment variable is set and the original value is restored in the destructor. Moreover since this clock inherit of `std::lock_guard` a mutex is locked for the whole life of this `env_guard` and so it avoids concurrent call to `set_env`.

Lomadriel:
Add `label-alt` and `format-alt`:

Documentation:
```ini

[module/date]
type = internal/date

; Seconds to sleep between updates
interval = 1.0

; See "http://en.cppreference.com/w/cpp/io/manip/put_time" for details on how to format the date string
; NOTE: if you want to use syntax tags here you need to use %%{...}
date = %Y-%m-%d%

; Optional time format
time = %H:%M

; if `date-alt` or `time-alt` is defined, clicking
; the module will toggle between formats
date-alt = %A, %d %B %Y
time-alt = %H:%M:%S

; Use the following command to list timezone
; ls /usr/share/zoneinfo/
; (Defaults: system timezone)
;timezone = 
timezone-alt = UTC

; Available tags:
; <label> (default)
; format = <label>

; Available tags:
; <label-alt> (default)
; format-alt = <label-alt>

; Available tokens:
;   %date%
;   %time%
; Default: %date%
label = %date% %time%
label-font = 3
label-foreground = #9A32DB

; Available tokens:
;   %date-alt%
;   %time-alt%
; Default: %date-alt%
label-alt = %date-alt% %time-alt%
label-alt-font = 3
label-alt-foreground = #9A32DB

```

adkillick:
I would love to be able to have a UTC clock because I'm lazy with time conversion math when looking at server logs. Any chance this will be merged at some point?

fosskers:
Until there is official support, we can get around this with a custom script that calls `date` directly. For example, here's what I use to show Japanese time alongside my machine's locale time:
```
[module/japan]
type = custom/script
interval = 5
exec = TZ=Asia/Tokyo date "+%H:%M %Z"
```

roomcays:
I imagine using `time` and `time-alt` to switch between local and UTC time with a single click.

patrick96:
Closing this because `setenv` is inherently unsafe in multi-threaded environments (no matter how many mutexes we use). See [this comment](https://github.com/polybar/polybar/pull/3160#issuecomment-2323348167) on a recent PR for more info

Issue #999, opened 2018-02-02
add hook state token for ipc


as3mbus:
hey i think it would be nice to have an active hook token for IPC module, which enable user to iterate through multiple click instead of different click button. 
so basically it'll be possible to do this
click-left = polybar-msg -p %pid% hook demo %activehook%+1

patrick96:
Closing in favor of #2464

PR #1686, opened 2019-03-12
feat(ipc_module): interval for hooks.

### LINKED ISSUE #1127: IPC module interval support
Is there a way to periodically refresh an IPC module? So that the initial hook gets executed every `interval` seconds, similar to a custom script.

What I'm trying to achieve is an update checker that periodically retrieves the number of updates available and resets this number once the system is updated using IPC. Simply using a custom script with a very short interval is not ideal, since I don't want to query the package database too frequently.

Lomadriel:
Fix #1127.

Add an `interval` for each hook (new key `hook-*-interval`, those keys are not required. If they are not present they are defaulted to 0 and it signifies that the hook should not be recalled. Otherwise the hook will be called every `hook-*-interval` seconds.


Lomadriel:
I also added a second commit which resolves #999 (%active-hook% token for module actions)

abrouwers:
I have a use for this, as I'm using a simple weather module that I'd like to update (a) Every 10 minutes, or (b) if I click ("what is the weather right now? after opening my laptop quickly).  I tried applying this PR to test, but it doesn't seem to apply to 3.5.5.  Is there any update available?  I'd be happy to test if so.

Basically, what I'm hoping will work is something like:

[module/weather]
type = custom/ipc
hook-0 = python3 ~/.config/polybar/scripts/weather.py
initial = 1
interval = 600
click-left = polybar-msg hook weather 1

############################################################

Here's what I see when I try to apply:

$ patch -Np1 -i ../1852.patch
patching file include/modules/ipc.hpp
Hunk #1 FAILED at 1.
Hunk #2 succeeded at 12 (offset -2 lines).
Hunk #3 FAILED at 23.
2 out of 3 hunks FAILED -- saving rejects to file include/modules/ipc.hpp.rej
patching file src/modules/ipc.cpp
Hunk #1 FAILED at 1.
Hunk #6 FAILED at 144.
2 out of 6 hunks FAILED -- saving rejects to file src/modules/ipc.cpp.rej
patching file include/modules/ipc.hpp
Hunk #1 FAILED at 43.
1 out of 1 hunk FAILED -- saving rejects to file include/modules/ipc.hpp.rej
patching file include/utils/string.hpp
patching file src/modules/ipc.cpp
Hunk #1 FAILED at 96.
Hunk #2 succeeded at 148 with fuzz 2 (offset -45 lines).
1 out of 2 hunks FAILED -- saving rejects to file src/modules/ipc.cpp.rej

PR #1687, opened 2019-03-12
font: Draw red rectangle for unmatched charachters

patrick96:
I think we need a visual indicator when characters are dropped because most people just think that the module doesn't show at all.

I thought about drawing a hollow rectangle, similar to what terminals do, but I found it easier to just draw a red rectangle.

This can probably be polished a bit more.

@NBonaparte thoughts?


patrick96:
Alright, I found out that fonts provide a `defaultChar` which is a glyph for characters that are not in the font.

I have only figured out how to draw all characters as those glyphs and it looks like this:

dropped

This is probably a nice solution to the problem of there being no visual indicator for dropped characters. But I think we need to rewrite some font rendering stuff for this.

PR #1692, opened 2019-03-16
Apply clang-tidy modernize fixes

DanielChabrowski:
Hey,
I quickly applied clang-tidy modernize-* fixes with clang-tidy 7.0. It would be nice to also call clang-format but there were multiple changes that would obfuscate the PR.

### DIFF
--- a/include/adapters/mpd.hpp
+++ b/include/adapters/mpd.hpp
@@ -1,9 +1,9 @@
 #pragma once
 
 #include <mpd/client.h>
-#include <stdlib.h>
 #include <chrono>
 #include <csignal>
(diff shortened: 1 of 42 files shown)


patrick96:
Thanks. Unfortunately, this will create merge conflicts for lots of PR and then make those more difficult to review once those branches are rebased. We had the same concerns in #1079.

I don't think it is possible to use those clang tools all at once without putting a huge burden on contributors that have open PRs. We should develop a plan on how to do this with as few conflicts as possible. Maybe something like that for every PR, we should run clang-tidy and clang-format on all changed files.

PR #1693, opened 2019-03-17
feat(xworkspaces): add urgent hint cache.

### LINKED ISSUE #874: xworkspace Improvements

### LINKED ISSUE #1081: Problem with xworkspaces label-urgent
I'm running polybar on Arch Linux with Xmonad.

The xworkspaces module correctly identifies and labels the active workspace, but labels all other workspaces as "empty". I haven't been able to get the "occupied" or "urgent" labels to show under an circumstances.

My xworkspaces config:
```dosini
[module/workspaces]
type = internal/xworkspaces
format = <label-state>
pin-workspaces = false
label-active-foreground = ${colors.foreground-alt}
label-empty-foreground = ${colors.foreground-alt}
label-active = "  "
label-empty = "  "
label-occupied = "  "
label-urgent = "  "
```

Is this a known bug? Is there a clear problem with that section of my config?

Lomadriel:
Fixes #1081.

Also fix a segfault when a i3-sticky window exists.
These windows are on a "magic" desktop and the returned desktop id is garbage so it must be ignored.

This PR includes all the changes from #882, applies the changes requested by @patrick96 and adds a cache for the urgent hint to prevent it from disappearing when the desktop is changed.

I also rebased the branch from @kronn onto master

### DIFF
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -164,17 +164,17 @@ namespace modules {
     bounds.erase(std::unique(bounds.begin(), bounds.end(), [](auto& a, auto& b) { return a == b; }), bounds.end());
 
     unsigned int step;
-    if (bounds.size() > 0) {
+    if (!bounds.empty()) {
       step = m_desktop_names.size() / bounds.size();
     } else {
       step = 1;
     }
     unsigned int offset = 0;
-    for (unsigned int i = 0; i < bounds.size(); i++) {
-      if (!m_pinworkspaces || m_bar.monitor->match(bounds[i])) {
+    for (const auto& bound : bounds) {
+      if (!m_pinworkspaces || m_bar.monitor->match(bound)) {
         auto viewport = make_unique<struct viewport>();
         viewport->state = viewport_state::UNFOCUSED;
-        viewport->pos = bounds[i];
+        viewport->pos = bound;
 
         for (auto&& m : m_monitors) {
           if (m->match(viewport->pos)) {
@@ -216,6 +216,9 @@ namespace modules {
       for (auto&& d : v->desktops) {
         if (m_desktop_names[d->index] == m_current_desktop_name) {
           d->state = desktop_state::ACTIVE;
+          m_urgent_desktops.erase(m_current_desktop_name);
+        } else if (m_urgent_desktops[m_desktop_names[d->index]]) {
+          d->state = desktop_state::URGENT;
         } else if (occupied_desks.count(d->index) > 0) {
           d->state = desktop_state::OCCUPIED;
         } else {
@@ -258,6 +261,7 @@ namespace modules {
       for (auto&& d : v->desktops) {
         if (d->index == desk && d->state != desktop_state::URGENT) {
           d->state = desktop_state::URGENT;
+          m_urgent_desktops[m_desktop_names[d->index]] = true;
 
           d->label = m_labels.at(d->state)->clone();
           d->label->reset_tokens();
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -3,6 +3,7 @@
 #include <bitset>
 #include <mutex>
 #include <set>
+#include <unordered_map>
 
 #include "components/config.hpp"
 #include "components/types.hpp"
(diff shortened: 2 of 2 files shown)


src/modules/xworkspaces.cpp
```diff
@@ -201,6 +217,9 @@ namespace modules {
       }
       offset += step;
     }
+
+    m_desktop_occupied.resize(m_desktop_names.size(), 0);
```
kronn: Is is correct to initialize new elements with `0`. My understanding is that it should be `false`, since the vector is declared as `vector<bool> m_desktop_occupied{};`. 

```suggestion
    m_desktop_occupied.resize(m_desktop_names.size(), false);
```

I might be wrong about this, though. The content of this tally-object change during implementation of the initial PR quite a bit back and forth, so there might be other areas that need adaptation.

kronn:
Thanks for picking up my slack, @Lomadriel :-) I have only one comment, otherwise I think that this PR could replace mine. What do you think, @patrick96 ?

patrick96:
@kronn I actually wanted to comment and say I'd merge your PR first. But if you'd rather @Lomadriel works on it, we can of course do that. Doesn't really matter to me :)

src/modules/xworkspaces.cpp
Lomadriel: Yeah that's correct it's just a rest of a cancelled modition.

kronn:
@patrick96 I don't really care either way, but I see benefits in first finishing my PR, because it's already reviewed. Then this one would become a lot smaller and could be reviewed an merged faster as well. 

I will update my PR this evening, borrowing the requested changes from this PR, so a rebase will be cleanly possible.

kronn:
I will not do this, of course, if this PR gets merged sooner :wink:

patrick96:
> I will not do this, of course, if this PR gets merged sooner 😉

I don't think I'll be that fast ;)

Lomadriel:
> I will update my PR this evening, borrowing the requested changes from this PR, so a rebase will be cleanly possible.

I had to rebase your branch on top of master to resolve a conflict. The only commit that contains a new feature is the last one. I will update my branch according to yours.

kronn:
I rebased my PR and addressed the issues @patrick96 had. your commits add more to this than just the cleanup Patrick requested. I hope that my PR is now in a good enough state to no longer block your contribution.

Pascal3366:
I am still waiting for years to get support for occupied and urgent states in herbstluftwm (EWMH).
Why did this never gor merged??

Lomadriel:
> @Lomadriel Is there anything in this PR that is not in #882?

Yeah, it fixes urgent hint. I'll update it soon.

patrick96:
All right. Great!

dkasak:
So now that #882 is merged, is this PR obsolete apart from the urgency fixes? Should a new PR be made with just the urgency stuff?

patrick96:
@dkasak It's up to lomadriel, how he wants to continue this. But as far as I can tell we now only need the urgency fixes from this PR.

Lomadriel:
Rebased on master.

patrick96:
Just did a bunch of testing around the urgent hint.
There are some cases that this PR doesn't cover:

* The user selects the workspace with the urgent hint but now the window with the urgent hint. In that case this PR clears the urgent hint even though it would also be reasonable to leave it since the urgent hint should be cleared when the corresponding window is selected.
* The urgent hint is not cleared when the window with the hint disappears. For example if you run `sleep 2 && echo -e "\a" && sleep 2 && exit` in a terminal and switch away from the workspace, the urgent hint persists even though the window no longer exists.
* I haven't tested this but moving windows between workspaces (programmatically without actually selecting workspaces) and adding/removing workspaces before the workspace with the hint will likely also do weird things because the cache is index based.

We already know how to query whether a window has the urgent hint set: `icccm_util::get_wm_urgency`. With this we could populate the cache in `rebuild_clientlist` and let the WM handle clearing the urgent hint (which is its job anyway)

PR #1704, opened 2019-03-21
Add a whitelist for tokens

### LINKED ISSUE #1701: Windows title is not santized (xtitle)
## Describe the issue

While I was debugging polybar, I had a notepad opened to note some strings and I noticed that my main bar (not the one I was debugging) was freezing. After investigation I found that the culprit was the title of my notepad  `"%{l}%{A1:xkeyboard/switch:}fr  %{u#e60053 +u B#e60053"`

**Expected behavior:**

The title of my window should be displayed without having my bar freezing and the tags must be ignored.

**Actual behavior:**

Polybar freeze or if the tag formatting is correct, the tag actually change the xtitle module.

## To Reproduce
```dosini
[module/title]
type = internal/xwindow

format = <label>
label = %title%

[bar/example]
monitor = eDP-1
monitor-fallback =
monitor-strict = false
override-redirect = false
bottom = true
fixed-center = true
width = 100%
height = 20

font-0 = "Ubuntu Condensed:size=10;2"
font-1 = "icomoon\-feather:size=10;2"
font-2 = "xos4 Terminus:size=12;1"
font-3 = "HanaMinA:style=Regular:size=13;3"

modules-left =
modules-center = title
modules-right =

```
+ Having a window title which have a tag in it

## Polybar Log
```
* Bar geometry: 1920x20+0+1060; Borders: 0,0,0,0
- bar: Attach X event sink
- bar: Attach signal receiver
* Starting application
* Starting module/title
- module/title: Thread id = 2
- bar: Create renderer
- renderer: Allocate colormap
- renderer: Load fonts
* Configured DPI = 96x96
* Loaded font "Ubuntu Condensed:size=10" (name=Ubuntu Condensed, offset=2, file=/home/loma/.local/share/fonts/Ubuntu-C.ttf)
* Loaded font "icomoon\-feather:size=10" (name=icomoon-feather, offset=2, file=/home/loma/.local/share/fonts/icomoon-feather.ttf)
* Loaded font "xos4 Terminus:size=12" (name=DejaVu Sans, offset=1, file=/usr/share/fonts/TTF/DejaVuSans.ttf)
* Loaded font "HanaMinA:style=Regular:size=13" (name=HanaMinA, offset=3, file=/usr/share/fonts/TTF/HanaMinA.ttf)
* Bar window: 0x1800001
- bar: Reconfigure window
- bar: Set window WM_NAME
- bar: Set window _NET_WM_PID
- bar: Map window
- bar: Draw empty bar
- bar: Setup tray manager
* module/title: Rebuilding cache
- bar: Force update
* Redrawing bar window
- renderer: flush(2 geom=289x20+816+0, falloff=0)
- bar: Received expose event
* module/title: Rebuilding cache
* Redrawing bar window
```

## Screenshots
Look it is fun I can inject colors in polybar with the title of my window by having a window with this name `"%{c}%{u#e60053 +u B#e60053}polybar [~/Code/repositories/polybar] - .../src/components/parser.cpp - CLion%{B- -u u-}"` :smile: 
polybar

## Environment:
* Output of `polybar -vvv`:
```
polybar 3.3.0

Build type: RelWithDebInfo
Compiler: /usr/bin/c++
```

Lomadriel:
Resolves #1701.
With this PR, tag are ignored in every token except %output% from the script module.

I think it would be better to add unit tests before merging (on the label and on the parser), but I would like your opinion on how the problem is solved.


include/drawtypes/label.hpp
```diff
@@ -51,27 +53,28 @@ namespace drawtypes {
...
+        , m_overline(move(overline))
         , m_font(font)
         , m_padding(padding)
         , m_margin(margin)
         , m_maxlen(maxlen)
         , m_ellipsis(ellipsis)
-        , m_text(text)
+        , m_text(move(text))
         , m_tokenized(m_text)
-        , m_tokens(forward<vector<token>>(tokens)) {
```
Lomadriel: Forward should only be used on universal references, here it is a rvalue so move should be used.
(See https://en.cppreference.com/w/cpp/utility/forward or if you want to look deeper into that https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers)

include/drawtypes/label.hpp
```diff
@@ -51,27 +53,28 @@ namespace drawtypes {
...
-          assert(!m_ellipsis || (m_maxlen == 0 || m_maxlen >= 3));
-        }
+        , m_tokens(move(tokens))
+        , m_token_whitelist{algo_util::sort_in_place(move(token_whitelist))} {
+      assert(!m_ellipsis || (m_maxlen == 0 || m_maxlen >= 3));
+    }
 
     string get() const;
-    operator bool();
+    explicit operator bool();
```
Lomadriel: I added an explicit to avoid implicit conversion of label to bool (it is to be noted that it is an explicit call to this operator in an if).

src/drawtypes/label.cpp
```diff
@@ -55,10 +55,14 @@ namespace drawtypes {
...
 
+        if (!std::binary_search(m_token_whitelist.cbegin(), m_token_whitelist.cend(), token)) {
+          repl = string_util::replace_all(repl, "%", "%%");
+        }
+
         /*
          * Only replace first occurence, so that the proper token objects can be used
          */
-        m_tokenized = string_util::replace(m_tokenized, token, move(repl));
+        m_tokenized = string_util::replace(m_tokenized, token, repl);
```
Lomadriel: Removed `move` since the parameter is a `const std::string&` so the `move` does nothing.

src/components/parser.cpp
```diff
@@ -33,15 +33,29 @@ parser::parser(signal_emitter& emitter) : m_sig(emitter) {}
...
 void parser::parse(const bar_settings& bar, string data) {
   while (!data.empty()) {
-    size_t pos{string::npos};
-
-    if (data.compare(0, 2, "%{") == 0 && (pos = data.find('}')) != string::npos) {
-      codeblock(data.substr(2, pos - 2), bar);
-      data.erase(0, pos + 1);
-    } else if ((pos = data.find("%{")) != string::npos) {
-      data.erase(0, text(data.substr(0, pos)));
+    size_t beg_pos{string::npos};
```
patrick96: Why was it necessary to rewrite the parser? This feature should really only affect the string produced by the builder.

Also, we use `%` to escape `%{` blocks because that's the way lemonbar does it. This would also break existing configs relying on this escaping behavior.

include/utils/algorithm.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include <algorithm>
+
+#include "common.hpp"
+
+POLYBAR_NS
+
+namespace algo_util {
```
patrick96: Did you add this, only so you could use binary search instead of linear seach when replacing tokens?

If so, please remove this.

In the common case the token whitelist will contain 0-2 items, here binary search is really not any faster than linear search.

Lomadriel: Yeah, I agree, I will remove that

src/components/parser.cpp
Lomadriel: Hum, maybe I missed something but the actual parser does not escape when there is a '%'.
I used '\' because I could not differentiate if the following string was tag escaping or just a percentage and a tag: '97%%{O100}'. But maybe I can double each percentage when a label is created.

patrick96: > the actual parser does not escape when there is a '%'.

Oh, that seems like an oversight, it really should. I was kind of dreading this moment. It is entirely possible that the whole escaping business inside polybar breaks in various places if someone is taking a closer look. I was taking a look at the parser a few months back but couldn't make heads or tails of what polybar does with escaping and I just wrote down all my questions and put them away for later. Seems my question "How is escaping of formatting tags handled?" was answered: "Not at all".

I'm not really sure if it is even possible to escape to properly detect `97%%{O100}` without completely parsing all user supplied raw formatting tags. If you come up with a solution, let me know. Maybe wait a bit with implementing it so we can discuss it first.

Lomadriel: I think that's indeed impossible. As we discussed on the IRC, maybe the best is just breaking the backward compatibility and using the comportment of Lemonbar.

I will rewrite the parser to take in account that '%' is a special character. And as you said, this PR should not be merged before the 4.0 of polybar.

However, maybe we can use the solution proposed in #1065 (with random IDs) as a temporary solution while waiting the 4.0.

patrick96: I think we have to use those X tags. But maybe we can build those on top of https://github.com/jaagr/polybar/pull/1596 which introduces the `%{P` tag for polybar control commands, so we could use `%{PX`

patrick96: Looking at this again, and man this is a difficult problem.

I think the most elegant way to solve this (and also the most time-consuming) is to stop using lemonbar formatting tags internally (at least in string form). So instead of passing a large formatting string from the controller to the parser and renderer, we pass an array of formatting elements. This array is basically the formatting string represented using C++ data structures, so each element is either just plain text or describes a formatting tag.

Formatting tags are parsed when text is read from labels, formats, ramps, tokens, etc. From then on out we never deal with formatting tags as pure strings anymore.

I think that can solve this issue without breaking too many configs. The configs that will still break are the ones that literally write something like `97%%{O100}` in one of their labels or as part of the script output. I think this is acceptable even without a major release.

PR #1709, opened 2019-03-25
Release 3.3.1

patrick96:
A patch release with a few bug fixes is way overdue.

This adds PRs #1576, #1579, #1591, #1612, #1699, #1514, #1608, #1571, #1644, #1675, #1669, #1647, #1678 to 3.3.1

I decided to not update PKGBUILDs in the release branches because it would be a hassle to add the sha256 hashes to the PKGBUILDs in the release branch *after* the release. So we are just always going to update the PKGBUILDs in `master`

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,13 +129,18 @@ Find a more complete list on the [dedicated wiki page](https://github.com/jaagr/
 
 Please [report any problems](https://github.com/jaagr/polybar/issues/new/choose) you run into when building the project.
 
-  ~~~ sh
-  $ git clone --branch 3.3 --recursive https://github.com/jaagr/polybar
-  $ mkdir polybar/build
-  $ cd polybar/build
-  $ cmake ..
-  $ sudo make install
-  ~~~
+Download the `polybar-<version>.tar` for the version you want to build from the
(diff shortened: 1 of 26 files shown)


NBonaparte:
What's going on with the commits?

patrick96:
What do you mean?

patrick96:
Yep, that's the idea. See our [Release Guidelines](https://github.com/jaagr/polybar/wiki/Release-Guidelines#patch-release). I didn't see any other way to get specific fixes into releases without having to release all the features as well.

Will merge this today or tomorrow

patrick96:
I'm a moron :man_facepalming: I accidentally pushed the changes for this PR to the 3.3 branch on this repo instead of my fork. The 3.3.1 changes have been live from the start.

Next time, I'll need to take better care.

NBonaparte:
That makes more sense.
Why are the commit ids different than the ones in master though? If you just cherry picked them, wouldn't they be the same?

patrick96:
I think cherry picking reapplies the diff and thus creates a new commit.

The system is not ideal, but I haven't found a better way. Do you know how other projects approach this?

PR #1712, opened 2019-03-26
Update README.md

RMPR:
Install instructions for fedora.


patrick96:
There are 21 repositories on copr that provide polybar. Why should we link to this particular one?

RMPR:
No particular reason, I just used it to have i3-gaps and I noticed that polybar was there as well.

patrick96:
Then I'd rather not add it to the README, if it's not an official package or at least regularly updated.

Thanks anyway :)

PR #1722, opened 2019-04-01
Create CODE_OF_CONDUCT.md

patrick96:
Github recommends we add a CoC to our project. For now it's just standard Contributor Covenant Code of Conduct but we can also adapt it in the future if we want to address specific issues.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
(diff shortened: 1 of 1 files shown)


patrick96:
> given its rather colourful past.

What do you mean by that? Maybe you know more than me ;)

I thought it was a good idea to add it to show that if there was inappropriate conduct, there would be someone to handle it. Even if we never had problems before.

I also wanted to write our own CoC but when I saw that there are templates available, I just used one of those because I didn't really have time. But I agree, writing our own would be more relevant and effective.

PR #1725, opened 2019-04-03
Replace condition_variable from STL by a monotonic implementation.

### LINKED ISSUE #857: Modules stop updating when system clock moves backwards
I've noticed that some modules (such as the cpu, memory and date) stop updating after the system clock has moved backwards.

I dual boot with Windows which is constantly messing up my hardware time, however NTP fixes it when the system boots. This usually happens after Polybar has been launched though, and it ends up in this broken state.

I'm running 3.0.5 on Arch. If I get time this weekend I'll look into fixing it, but if you know where to start looking that would help :)

### LINKED ISSUE #1932: incorrect time after unhibernating
## Describe the issue
After unhibernating the computer, the display time is wrong.

## To Reproduce
- hibernate
- wait some time
- unhibernate

## Environment:
* WM: i3
* Distro: Arch Linux
* Output of `polybar -vvv`:
```
polybar 3.4.1

Build type: Release
Compiler: /usr/bin/c++
Linker flags: -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now```

Lomadriel:
Fixes #857.

Add a re-implementation of `std::condition_variable` that uses the monotonic clock of POSIX.
Also replaces the call to `system_clock::now()` by `steady_clock::now()`.

EDIT:

Fixes #1932

### DIFF
--- a/cmake/03-libs.cmake
+++ b/cmake/03-libs.cmake
@@ -31,3 +31,13 @@ querylib(WITH_XCURSOR "pkg-config" xcb-cursor libs dirs)
 if(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
   querylib(TRUE "pkg-config" libinotify libs dirs)
 endif()
+
+include(CheckLibraryExists)
+check_symbol_exists(clock_gettime "time.h" HAVE_CLOCK_GETTIME)
+
+if (NOT HAVE_CLOCK_GETTIME)
+  # Before glibc 2.17 clock_gettime is in rt library
+  CHECK_LIBRARY_EXISTS(rt clock_gettime "time.h" WITH_RT)
+  set(HAVE_CLOCK_GETTIME ${WITH_RT})
+  find_package(rt REQUIRED)
+endif()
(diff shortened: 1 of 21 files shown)


Lomadriel:
A little update concerning this PR, this bug has been fixed in GCC10 if I'm not wrong.

patrick96:
I assume you are refering to [this](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=41861) bug. So that would mean we could scrap the use of our custom condition_variable but keep using `steady_clock`.

It's probably going to be a long time until gcc 10 is installed on most users computers but I'm not sure if maintaining our own version of `condition_variable` is appropriate, since the issue is easily fixed by restarting polybar in place.

Also I'm wondering whether or not clang is affected, I would assume not

Lomadriel:
> I assume you are refering to [this](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=41861) bug. So that would mean we could scrap the use of our custom condition_variable but keep using `steady_clock`.

Yes it's this bug.

It's probably going to be a long time until gcc 10 is installed on most users computers but I'm not sure if maintaining our own version of condition_variable is appropriate, since the issue is easily fixed by restarting polybar in place.

Indeed that might be overkill for DST but I remember that someone had a problem with a clock that wasn't working properly but I can't find that issue anymore.

patrick96:
You mean #857?

Issue #1753, opened 2019-04-25
[internal/battery]: %percentage% reports 100% if real percentage is greater than full-at


roobre:
## Describe the issue

In some cases is useful to get the `format-full` output when the battery is above a threshold, like with thinkpads and other laptops which allow the user to stop the charging process above certain value. This is what (I think) `full-at` property is meant to do. However, when the value is below `full-at`, the value of the `%percentage%` used in `label-full` is hardcoded to 100%. I don't think this is expected behavior, `full-at` should only control which format is used, not which value is reported.

**Expected behavior:**

Real `%percentage%` is reported, regardless of the `full-at` property.

**Actual behavior:**

Real `%percentage%` is reported, regardless of the `full-at` property.

**Was it working before?**
No idea.

## To Reproduce

```
[module/battery]
type = internal/battery

battery = BAT0
adapter = AC

full-at = 55

label-full = %percentage%%
format-full =  <label-full>

label-charging = %percentage%% %time%
format-charging =  <label-charging>

label-discharging = %percentage%% %time%
format-discharging = <label-discharging>
```

## Polybar Log
(*Not relevant*)
```

```

## Screenshots
![DeepinScreenshot_select-area_20190425123352](https://user-images.githubusercontent.com/969721/56729799-67554280-6756-11e9-8173-b24aade86c73.png)

## Environment:
* WM: i3
* Distro: Arch
* Output of `polybar -vvv`:
```polybar 3.3.1

Features: +alsa +curl +i3 -mpd +network(libnl) +pulseaudio +xkeyboard

X extensions: +randr (+monitors) -render -damage -sync +composite +xkb +xrm +xcursor

Build type: RelWithDebInfo
Compiler: /usr/lib/distcc/bin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -Wall -Wextra -Werror -Wno-noexcept-type -pedantic -pedantic-errors
Linker flags: -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now```

patrick96:
This is intended behavior. The intention behind `full-at` is that some batteries never report a 100% charge and if the battery is over the `full-at` threshold it will be considered full with 100% charge. This has been the behavior since the first commit.

[...]

PR #1756, opened 2019-05-02
battery: add `%percentage_raw%` token, which is not forced to `100` when real percentage is above `full-at`

roobre:
Following the discussion on #1753, this PR adds `%percentage_raw%` to the battery module, which represent the real percentage reported by the kernel regardless of the value set in the `full-at` property by the user. This property still sets the state as *Full* when the percentage level goes above its value. The behavior of the `%percentage%` token has not been altered.

Since in the next major (breaking) release the plan is to bring this behavior to the `%percentage%` token, the code has been refactored towards this deprecation of this old behavior, storing the real percentage instead of the clamped one and adding a clamping function for that purpose.

Feedback and comments are welcome.

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -230,6 +230,7 @@ namespace modules {
     if (label) {
       label->reset_tokens();
       label->replace_token("%percentage%", to_string(m_percentage));
+      label->replace_token("%percentage_raw%", to_string(current_percentage_raw()));
       label->replace_token("%consumption%", current_consumption());
 
       if (m_state != battery_module::state::FULL && !m_timeformat.empty()) {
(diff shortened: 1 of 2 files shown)


roobre:
> To not break existing configs, we unfortunately need to also clamp the percentage when passing it to the ramp and bar in the `build` function.

You're correct! I did not notice this because my particular config does not use `<ramp_capacity>` for `format-full`, I just hardcoded it to a fixed "battery full" icon instead. I'll fix it as soon as I can.

roobre:
Please tell me if you see something else off.
Thanks!

src/modules/battery.cpp
```diff
@@ -294,10 +295,14 @@ namespace modules {
   /**
    * Get the current capacity level
    */
-  int battery_module::current_percentage(state state) {
+  int battery_module::current_percentage() {
     int percentage{read(*m_capacity_reader)};
```
patrick96: This can be a single return statement

roobre: Fixed

PR #1767, opened 2019-05-14
Remove warnings in cmake

Lomadriel:
As stated in the documentation of set (https://cmake.org/cmake/help/latest/command/set.html), `INTEGER` type doesn't exist.
`STRING` type is now used and a check has been added.

This PR removes the following warnings:
```
CMake Warning (dev) at cmake/02-opts.cmake:90 (set):
  implicitly converting 'INTEGER' to 'STRING' type.
CMake Warning (dev) at cmake/02-opts.cmake:91 (set):
  implicitly converting 'INTEGER' to 'STRING' type.
```
CMake version: 3.14.3


Lomadriel:
Done

PR #1769, opened 2019-05-15
Remove undefined behaviors and clean code

Lomadriel:
This PR goal is to remove some "errors" in the code base.
See my review explaining each problem.


src/x11/background_manager.cpp
```diff
@@ -189,7 +189,7 @@ void bg_slice::allocate_resources(const logger& log, xcb_visualtype_t* visual) {
 }
 
 void bg_slice::free_resources() {
-  m_surface.release();
```
Lomadriel: Release doesn't free. This function releases the ownership of the managed object. So there is a memory leak.
`smart_ptr.reset()` or `smart_ptr = nullptr` must be used to delete the object.

Lomadriel:
I just found some strange code parts, but I don't really know if there is a reason behind this or if the code need to be corrected. I will try to investigate.
https://github.com/polybar/polybar/blob/cb75857ff66543e9aedac3f87df130ba03132471/include/utils/math.hpp#L53-L56
Is `max_value - max_value` really intended? I think this should be `0` in this case no?

https://github.com/polybar/polybar/blob/cb75857ff66543e9aedac3f87df130ba03132471/include/x11/connection.hpp#L108-L110
This copy assignment operator is also strange.

https://github.com/polybar/polybar/blob/cb75857ff66543e9aedac3f87df130ba03132471/src/x11/connection.cpp#L151-L166
The inner loop also seems to be buggy because of the unconditional return.

Lomadriel:
I also have a remark about theses `int i = my_float + 0.5f` that can be found in the code, like:
https://github.com/polybar/polybar/blob/cb75857ff66543e9aedac3f87df130ba03132471/include/utils/math.hpp#L72
Maybe using a function from the STL (see https://en.cppreference.com/w/cpp/numeric/math/round) should be used to do the rounding since in some cases, the current method doesn't work.
Should I replace every `+ 0.5` with `std::round` or `std::lround`?

include/x11/xresources.hpp
```diff
@@ -48,7 +48,7 @@ class xresource_manager {
   T get(const char* name, const T& fallback) const {
     try {
       return convert<T>(require<T>(name));
-    } catch (const xresource_error) {
+    } catch (const xresource_error&) {
```
Lomadriel: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Re-exception-ref

patrick96 (approved):
LGTM

PR #1770, opened 2019-05-16
cleanup(logger): use universal references to avoid useless copies

Lomadriel:
Use const ref and universal references to avoid to copy a lot of strings especially when the function is "disabled" (like `trace_x`).

### DIFF
--- a/src/modules/fs.cpp
+++ b/src/modules/fs.cpp
@@ -56,7 +56,7 @@ namespace modules {
     // Warn about "unreachable" format tag
     if (m_formatter->has(TAG_LABEL_UNMOUNTED) && m_remove_unmounted) {
       m_log.warn("%s: Defined format tag \"%s\" will never be used (reason: `remove-unmounted = true`)", name(),
-          TAG_LABEL_UNMOUNTED);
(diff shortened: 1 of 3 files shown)


src/modules/fs.cpp
```diff
@@ -56,7 +56,7 @@ namespace modules {
     // Warn about "unreachable" format tag
     if (m_formatter->has(TAG_LABEL_UNMOUNTED) && m_remove_unmounted) {
       m_log.warn("%s: Defined format tag \"%s\" will never be used (reason: `remove-unmounted = true`)", name(),
-          TAG_LABEL_UNMOUNTED);
+          string{TAG_LABEL_UNMOUNTED});
```
Lomadriel: Added because the constexpr member variable isn't inline in C++14 (C++17 is needed for that) and so we can't use it when it is not ODR-used.
See: 
https://en.cppreference.com/w/cpp/language/definition#ODR-use
https://en.cppreference.com/w/cpp/language/constexpr

To use it without the `string{}`, it is needed to add the definition into the cpp.
Since it is quite annoying, I decided to use add a surrounding `string{}` to be in a ODR-use case.

Issue #1211, opened 2018-05-04
Network module: failed to query interface


ArcticXWolf:
Hey,

I have a problem with the network module not working for my USB network interface.

This is a simple version config:
```
[colors]
background = #00000000
foreground = #ccffffff
dark-transparent = #40000000
focused-background = #99ffffff
focused-foreground = #cc000000
urgent-background = #99ff3333

[globalbars]
width = 100%
height = 50
radius = 0
fixed-center = true
background = ${colors.background}
foreground = ${colors.foreground}
border-size = 0
border-color =
padding = 0
padding-right = 2
module-margin = 2
font-0 = "Source Code Pro Semibold:pixelsize=12;5"

[bar/top]
inherit = globalbars
padding-left = 0
background = ${colors.dark-transparent}
modules-left = network1
modules-center =
modules-right =
tray-position =

[module/network1]
type = internal/network
interface = enx949aa984ea97
unknown-as-up = true
interval = 3.0

format-connected-padding = 2
format-connected-prefix = " "
label-connected = %local_ip%

label-disconnected = "test"

[settings]
screenchange-reload = true

[global/wm]
margin-top = 0
margin-bottom = 0
```

The output of polybar top -s:
```
warn: No monitor specified, using "eDP1"
warn: module/network1: Failed to query interface 'enx949aa984ea97'
warn: module/network1: Failed to query interface 'enx949aa984ea97'
error: Disabling tray manager (reason: Invalid position "")
%{l}test
warn: module/network1: Failed to query interface 'enx949aa984ea97'
warn: module/network1: Failed to query interface 'enx949aa984ea97'
warn: Termination signal received, shutting down...
```

I debugged a bit myself and identified [the IOCTL call](https://github.com/jaagr/polybar/blob/095d68fad08c223bc924e63ce9ac1021bcbc2feb/src/adapters/net.cpp#L230) as the issue, the parsed errno is always "Operation not supported".

Now my question: Is this an issue with my driver or do I just lack some package dependency? Or is this a bug within ethtool or polybar?
I'm using Ubuntu 17.10.

Thanks!

patrick96:
We had a similar issue in #351. The problem seems to be that `wireless_tools`, the tool we use for wireless support, doesn't support some USB interfaces. There is a PR in the pipeline to move to a newer tool (#1009), but it seems to have gone stale.

I don't know when this will be fixed, this is quite a big change.

su8:
>the tool we use for wireless support

[...]

PR #1772, opened 2019-05-16
update linkspeed for wired networks to fail to N/A instead of crashing module

hsartoris:
#1211 
See commit comment. Less frustrating behavior on failure to query interface for link speed.


src/adapters/net.cpp
```diff
@@ -306,6 +306,7 @@ namespace net {
    * about the current connection
    */
   string wired_network::linkspeed() const {
+    if (m_linkspeed == -1) return "N/A";
```
patrick96: We could probably just have the function only read:
```cpp
  string wired_network::linkspeed() const {
    return m_linkspeed == -1 ? "N/A" : (to_string(m_linkspeed) + " Mbit/s");
  }
```

Don't really know why 0 should print out `???`

severinkaderli:
Is there any update on this PR? I also suffer from this problem and would like this fix.

src/adapters/net.cpp
hsartoris: Finally remembered to make this change; see https://github.com/polybar/polybar/pull/1772/commits/9809d9e9c75c74f98f863931f949543ff1d3c4aa

hsartoris:
I don't really know much about codecov, but I can't see why there would be decreased coverage on files that weren't changed. @patrick96 merge anyway?

patrick96:
That sometimes happens with older PRs because the number of files tracked is different between this branch and `master` I think. Doesn't really matter.

Thanks a lot :)

hsartoris:
No problem; thanks for all the work you do!

Issue #1730, opened 2019-04-07
Compile error (Ubuntu bionic): request for member ‘composite’ is ambiguous


sjaenick:
## Build Process
*Describe how you build polybar, list the exact commands you are using:*

build configured  via build.sh, 
USE_GCC=ON
ENABLE_I3=OFF
ENABLE_ALSA=OFF
ENABLE_PULSEAUDIO=ON
ENABLE_NETWORK=ON
ENABLE_MPD=OFF
ENABLE_CURL=ON
ENABLE_IPC_MSG=ON

**If you build polybar directly from this repository:**

* Output of `git describe --tags`:

$ git describe --tags
3.3.0-68-g7d0c630

## Build Log

```
+ cmake -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DENABLE_ALSA:BOOL=OFF -DENABLE_PULSEAUDIO:BOOL=ON -DENABLE_I3:BOOL=OFF -DENABLE_MPD:BOOL=OFF -DENABLE_NETWORK:BOOL=ON -DENABLE_CURL:BOOL=ON -DBUILD_IPC_MSG:BOOL=ON ..
-- Trying to enable ccache
-- Couldn't locate ccache, disabling ccache...
-- The C compiler identification is GNU 7.3.0
-- The CXX compiler identification is GNU 7.3.0
-- Check for working C compiler: /usr/bin/gcc
-- Check for working C compiler: /usr/bin/gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/g++
-- Check for working CXX compiler: /usr/bin/g++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- No build type specified; using RelWithDebInfo
-- Using supported compiler GNU-7.3.0
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.1") 
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Check if compiler accepts -pthread
-- Check if compiler accepts -pthread - yes
-- Found Threads: TRUE  
-- Checking for module 'cairo-fc'
--   Found cairo-fc, version 1.15.10
[...]

PR #1788, opened 2019-06-01
build: Don't allow users to enable RENDER, DAMAGE, SYNC

patrick96:
Some build recipes (exherbo, sourcemage) for some reason enable those
three even though they don't do anything any only cause compilation
issues.

Ref: #1730


Lomadriel:
I think we shouldn't use an option for these variables.
According to CMake documentation (https://cmake.org/cmake/help/latest/command/option.html), an option is something `that the user can optionally select.`

Since these variable should not be modified by the end-user it might be a better solution to use
`set` (https://cmake.org/cmake/help/latest/command/set.html) with the `INTERNAL` type.

However, `WITH_XRANDR` seems to be used by the travis:
https://github.com/polybar/polybar/blob/7724c4dccb18ed9e48b050713540a2d05eacd2a1/common/travis/configure.sh#L10-L24

patrick96:
> However, `WITH_XRANDR` seems to be used by the travis:

We're only disabling XR**E**NDER :wink: 

I will just completely remove all three options and their extensions from the code base. If we ever want to use them we can just revert the commit that removed them.

Lomadriel:
> We're only disabling XR**E**NDER wink

Ha, my bad!

Issue #1732, opened 2019-04-08
Concurrency issues in polybar (list of bugs)


Lomadriel:
This issue is a list of concurrency issues that must find solution:

1. Issue in `timer_module`:
1.1.
    https://github.com/jaagr/polybar/blob/7414e9800879cc3692af5ca217fdb18978ba8b4a/include/modules/meta/timer_module.hpp#L33
Here there is a call to a virtual method however when the class is destroyed by the controller
https://github.com/jaagr/polybar/blob/7414e9800879cc3692af5ca217fdb18978ba8b4a/src/components/controller.cpp#L140-L149
the thread still run and this may leads to a non desired behavior since it is quite dangerous to call virtual methods in a constructor or a destructor (during construction or destruction, the dynamic dispatch considers that the object is of the exact type of current destructor). <br>
Moreover since the vtable changes during the destruction, acceding a virtual method from another thread may resolves in an undefined behavior (I am not really sure about the last sentence, please feel free to verify my statement).
This issue can be solved by adding a `final` to the `running` method. (See https://github.com/Lomadriel/polybar/commit/7821065e4051747b720eb39aef683e69ab35210b)
Since the final is in an upper class (`module`), the dynamic dispatch is not used. The call to `running` method is now statically resolved inside the `timer_module` class and so calling `running` is equivalent to a call to any non-virtual function. <br>
This issue is also present in `event_module` and `inotify_module`, 
https://github.com/jaagr/polybar/blob/7414e9800879cc3692af5ca217fdb18978ba8b4a/include/modules/meta/event_module.hpp#L32
https://github.com/jaagr/polybar/blob/7414e9800879cc3692af5ca217fdb18978ba8b4a/include/modules/meta/inotify_module.hpp#L28
1.2.
https://github.com/jaagr/polybar/blob/7414e9800879cc3692af5ca217fdb18978ba8b4a/include/modules/meta/timer_module.hpp#L23-L26
When `update` is called, `build` can't be called since the internal state is not protected.
this must be:
```cpp
const auto check = [&]() -> bool {
  std::lock(this->m_updatelock, this->m_buildlock);
  std::unique_lock<std::mutex> guard_update(this->m_updatelock, std::adopt_lock);
  std::unique_lock<std::mutex> guard_build(this->m_buildlock, std::adopt_lock);

  return CAST_MOD(Impl)->update();
};
```
[...]

PR #1791, opened 2019-06-01
fix(modules): move join in a new method

Lomadriel:
Addresses point 2 of #1732.

Since the joins are in the `stop` method, the vtable is no longer acceded during the destruction. This also solve the CRTP problem.

### DIFF
--- a/include/modules/meta/base.hpp
+++ b/include/modules/meta/base.hpp
@@ -131,6 +131,7 @@ namespace modules {
     virtual bool input(const string& action, const string& data) = 0;
 
     virtual void start() = 0;
+    virtual void join() = 0;
     virtual void stop() = 0;
     virtual void halt(string error_message) = 0;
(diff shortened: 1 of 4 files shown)


Lomadriel:
I found a deadlock :(

Lomadriel:
I moved this in a new method rather than in the `stop`.
If the `join` if in the `stop` method then if the module thread calls `halt`, the thread try to join itself and it makes polybar crash :/

Lomadriel:
I did that in https://github.com/polybar/polybar/commit/440b515507dac605a5438cf2429fcae564175262 but it introduces a deadlock but I just saw why. I'll try to do that.

patrick96 (approved):
Seems I stumbled on the same issue in #2518 but I solved it by moving join into the `stop` method which has the same issue when a module calls `halt` from any other thread than the main thread.

PR #1792, opened 2019-06-01
fix(interrupt_handler): remove undefined behavior and useless code

Lomadriel:
In C++14, from a signal handler we can't modify variable whose type is not `std::atomic` or `volatile std::sig_atomic_t` (https://en.cppreference.com/w/cpp/utility/program/signal).

Moreover there is no need to write something into the eventpipe since `select` is interrupted by signals.

### DIFF
--- a/src/components/controller.cpp
+++ b/src/components/controller.cpp
@@ -21,20 +21,19 @@
 
 POLYBAR_NS
 
-array<int, 2> g_eventpipe{{-1, -1}};
-sig_atomic_t g_reload{0};
-sig_atomic_t g_terminate{0};
+namespace {
+  array<int, 2> g_eventpipe{{-1, -1}};
+  std::atomic_bool g_reload{false};
+  std::atomic_bool g_terminate{false};
+}
 
 void interrupt_handler(int signum) {
   if (g_reload || g_terminate) {
     return;
   }
 
-  g_terminate = 1;
+  g_terminate = true;
   g_reload = (signum == SIGUSR1);
-  if (write(g_eventpipe[PIPE_WRITE], &g_terminate, 1) == -1) {
-    throw system_error("Failed to write to eventpipe");
-  }
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/components/controller.cpp
```diff
@@ -21,20 +21,19 @@
...
 
 void interrupt_handler(int signum) {
   if (g_reload || g_terminate) {
     return;
   }
 
-  g_terminate = 1;
+  g_terminate = true;
   g_reload = (signum == SIGUSR1);
-  if (write(g_eventpipe[PIPE_WRITE], &g_terminate, 1) == -1) {
```
patrick96: I have read the manpage of `pselect` and I agree that signals should make it return. But during testing this didn't work. Triggering a reload with `kill -SIGUSR1` always works, but If I use `polybar-msg cmd restart`, it doesn't. On a restart command polybar will call `raise(SIGUSR1)` in its code.

The interrupt handler is also called, but in that case, it seems `select` doesn't return until there is another event.

I used the following minimal config:

```dosini
[bar/example]
modules-left = text

enable-ipc = true

[module/text]
type = custom/text
content = Filler
```

patrick96: Oh, and of course when I again add the `write` call, it works. But calling `write` is probably undefined behavior.

Lomadriel: I know why...
`select` is not interrupted because the `raise` triggers on the event thread :/
See https://unix.stackexchange.com/questions/225687/what-happens-to-a-multithreaded-linux-process-if-it-gets-a-signal for more information.
I'll try to modify this PR to take that in account.

patrick96: Wouldn't this even mean that `kill -SIGUSR1` only works by accident, since it could deliver the signal to any thread?

patrick96:
We now have a proper signal handler with libuv that doesn't have these issues.

PR #1807, opened 2019-06-11
feat(cmake): removing globs and using imported target

### LINKED ISSUE #1536: Separate hard and soft dependencies in cmake
hard dependencies like xcb-randr or xcb-composite shoud be impossible to disable or cmake itself should fail when you try (as opposed to the compilation failing when `randr.hpp` is included).

Soft dependencies should work as before.

With this we can give more informative error messages instead of the user hitting
```
 #error "X Composite extension is disabled..."
```
10 times during compilation.

I don't think we should remove the `WITH_XRANDR`, `WITH_XCOMPOSITE`, etc. `#define`s entirely but still use them for `polybar -vvv`.

Ref #1430

Lomadriel:
This is the first of many PRs aiming at splitting polybar modules into separate shared objects.

In this PR, I removed all the globs in the cmake.
Moreover I also modernized the cmake by using imported targets rather than using `dirs` and `libs` variables.

EDIT:
Closes #1536.

### DIFF
--- a/cmake/03-libs.cmake
+++ b/cmake/03-libs.cmake
@@ -3,28 +3,55 @@
 #
 
 find_package(Threads REQUIRED)
-list(APPEND libs ${CMAKE_THREAD_LIBS_INIT})
+find_package(CairoFC REQUIRED)
 
-querylib(TRUE "pkg-config" cairo-fc libs dirs)
-
-querylib(ENABLE_ALSA "pkg-config" alsa libs dirs)
-querylib(ENABLE_CURL "pkg-config" libcurl libs dirs)
-querylib(ENABLE_MPD "pkg-config" libmpdclient libs dirs)
-if(WITH_LIBNL)
-  querylib(ENABLE_NETWORK "pkg-config" libnl-genl-3.0 libs dirs)
-else()
-  querylib(ENABLE_NETWORK "cmake" Libiw libs dirs)
+if (ENABLE_ALSA)
+  find_package(ALSA REQUIRED)
 endif()
-querylib(ENABLE_PULSEAUDIO "pkg-config" libpulse libs dirs)
 
-querylib(WITH_XCOMPOSITE "pkg-config" xcb-composite libs dirs)
-querylib(WITH_XKB "pkg-config" xcb-xkb libs dirs)
-querylib(WITH_XRANDR "pkg-config" xcb-randr libs dirs)
-querylib(WITH_XRANDR_MONITORS "pkg-config" "xcb-randr>=1.12" libs dirs)
-querylib(WITH_XRM "pkg-config" xcb-xrm libs dirs)
-querylib(WITH_XCURSOR "pkg-config" xcb-cursor libs dirs)
+if (ENABLE_CURL)
+  find_package(CURL REQUIRED)
+endif()
+
+if (ENABLE_MPD)
+  find_package(LibMPDClient REQUIRED)
+endif()
+
+if (ENABLE_NETWORK)
+  if(WITH_LIBNL)
+    find_package(LibNlGenl3 REQUIRED)
+  else()
+    find_package(Libiw REQUIRED)
+  endif()
+endif()
+
+if (ENABLE_PULSEAUDIO)
+  find_package(LibPulse REQUIRED)
+endif()
+
+# Randr is required
+set(XORG_EXTENSIONS RANDR)
+if (WITH_XCOMPOSITE)
+  set(XORG_EXTENSIONS ${XORG_EXTENSIONS} COMPOSITE)
+endif()
+if (WITH_XKB)
+  set(XORG_EXTENSIONS ${XORG_EXTENSIONS} XKB)
+endif()
+if (WITH_XCURSOR)
+  set(XORG_EXTENSIONS ${XORG_EXTENSIONS} CURSOR)
+endif()
+if (WITH_XRM)
+  set(XORG_EXTENSIONS ${XORG_EXTENSIONS} XRM)
+endif()
+
+set(XCB_VERSION "")
+if (WITH_XRANDR_MONITOR)
+  set(XCB_VERSION "1.12")
+endif()
+
+find_package(Xcb ${XCB_VERSION} REQUIRED COMPONENTS ${XORG_EXTENSIONS})
 
 # FreeBSD Support
 if(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
-  querylib(TRUE "pkg-config" libinotify libs dirs)
+  find_package(LibInotify REQUIRED)
 endif()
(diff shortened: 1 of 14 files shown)


Lomadriel:
Oops, I changed the output directory of the binaries :/

Lomadriel:
I think, we also should (not in this PR) remove all modifications of the `${CMAKE_*}` variables, this is really a bad practice and the dev of cmake recommends not to use them.

patrick96:
Before we only had to call `querylib` which in turn called `pkg_check_modules` to load flags and directories for dependencies. Now we have a separate `FindX.cmake` with lots of code for each dependency. How is this better? I understand that it is probably cleaner because we have a cmake target for each dependency but is there no way around using `FindX.cmake` files?

>I think, we also should (not in this PR) remove all modifications of the ${CMAKE_*} variables, this is really a bad practice and the dev of cmake recommends not to use them.

You mean like in `01-core.cmake` where we modify compiler flags? How else are we supposed to set those flags?

Lomadriel:
> Before we only had to call `querylib` which in turn called `pkg_check_modules` to load flags and directories for dependencies. Now we have a separate `FindX.cmake` with lots of code for each dependency. How is this better? I understand that it is probably cleaner because we have a cmake target for each dependency but is there no way around using `FindX.cmake` files?

The main reason why I did that is to prepare the next phase of the split.
If we have separated targets, we can link the right target to the right module.
For example, the main program should not be linked against `libcurl` only the "github" module should.

But If you want I can simplify the each `FindX.cmake` to be quite the same as before in the `querylib`.

-----

> You mean like in `01-core.cmake` where we modify compiler flags? How else are we supposed to set those flags?

The recommended way is to add them per target. I know this is quite annoying but it avoids the propagation to the dependencies (like for example XPP). This would allow us to potentially activate more warnings without having to handle the warnings of the dependencies.
I think the best way is to create a variable which contains the warnings and to add this variable to each target where we want to activate the warnings.
`target_compile_options(my_library PRIVATE ${POLYBAR_COMMON_WARNINGS})`

Lomadriel:
> But If you want I can simplify the each `FindX.cmake` to be quite the same as before in the `querylib`.

Example of the simplification of a `FindX.cmake`:
```cmake
find_package(PkgConfig REQUIRED)

pkg_check_modules(PC_CairoFC REQUIRED cairo-fc)

set(CairoFC_INCLUDE_DIR ${PC_CairoFC_INCLUDE_DIRS})
set(CairoFC_LIBRARY ${PC_CairoFC_LIBRARIES})

mark_as_advanced(CairoFC_INCLUDE_DIR CairoFC_LIBRARY)

set(CairoFC_LIBRARIES ${CairoFC_LIBRARY})
set(CairoFC_INCLUDE_DIRS ${CairoFC_INCLUDE_DIR})

if(NOT TARGET Cairo::CairoFC)
  add_library(Cairo::CairoFC INTERFACE IMPORTED)
  set_target_properties(Cairo::CairoFC PROPERTIES
    INTERFACE_INCLUDE_DIRECTORIES "${CairoFC_INCLUDE_DIRS}"
    INTERFACE_LINK_LIBRARIES "${CairoFC_LIBRARIES}"
  )
endif()
```

patrick96:
Don't we need `find_package_handle_standard_args`?

I'm also wondering whether or not there is any difference between these two:

```cmake
  set_target_properties(Cairo::CairoFC PROPERTIES
    INTERFACE_INCLUDE_DIRECTORIES "${CairoFC_INCLUDE_DIRS}"
    INTERFACE_LINK_LIBRARIES "${CairoFC_LIBRARIES}"
  )
```

```cmake
    set_target_properties(LibInotify::LibInotify PROPERTIES
      INTERFACE_LINK_LIBRARIES "${PC_LibInotify_LIBRARIES}")

    target_include_directories(LibInotify::LibInotify SYSTEM INTERFACE ${LibInotify_INCLUDES})
```

> The recommended way is to add them per target

Oh well, we'll keep it in mind :wink:

Lomadriel:
> Don't we need `find_package_handle_standard_args`?

Yeah indeed, that works without it but that's definitively better with it.



> I'm also wondering whether or not there is any difference between these two:

The main difference between the two is the `SYSTEM` it indicates to cmake to include the directory with `-isystem` to avoid the warning is these includes.
After reading  the [cmake documentation](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_SYSTEM_INCLUDE_DIRECTORIES.html), they are two ways to suppress warnings:
```cmake
set_target_properties(LibInotify::LibInotify PROPERTIES
      INTERFACE_LINK_LIBRARIES "${PC_LibInotify_LIBRARIES}")
target_include_directories(LibInotify::LibInotify SYSTEM INTERFACE ${LibInotify_INCLUDES})
```
or 
```cmake
set_target_properties(LibInotify::LibInotify PROPERTIES
      INTERFACE_LINK_LIBRARIES "${PC_LibInotify_LIBRARIES}"
      INTERFACE_INCLUDE_DIRECTORIES "${LibInotify_INCLUDES}"
      INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${LibInotify_INCLUDES}"
)
```
But the first way seems to be recommended. I think I'll modify that for all  the dependencies. Because currently I forgot to silence the warnings of most of our dependencies. :/

Lomadriel:
> Don't we need `find_package_handle_standard_args`?

Yeah indeed, that works without it but that's definitively better with it.



> I'm also wondering whether or not there is any difference between these two:

The main difference between the two is the `SYSTEM` it indicates to cmake to include the directory with `-isystem` to avoid the warning is these includes.
After reading  the [cmake documentation](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_SYSTEM_INCLUDE_DIRECTORIES.html), they are two ways to suppress warnings:
```cmake
set_target_properties(LibInotify::LibInotify PROPERTIES
      INTERFACE_LINK_LIBRARIES "${PC_LibInotify_LIBRARIES}")
target_include_directories(LibInotify::LibInotify SYSTEM INTERFACE ${LibInotify_INCLUDES})
```
or 
```cmake
set_target_properties(LibInotify::LibInotify PROPERTIES
      INTERFACE_LINK_LIBRARIES "${PC_LibInotify_LIBRARIES}"
      INTERFACE_INCLUDE_DIRECTORIES "${LibInotify_INCLUDES}"
      INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${LibInotify_INCLUDES}"
)
```
But the first way seems to be recommended. I think I'll modify that for all  the dependencies. Because currently I forgot to silence the warnings of most of our dependencies. :/

Lomadriel:
I just updated the find packages. I added a function and a macro to avoid duplicate code in the cmake.

patrick96:
Closing in favor of #2322

PR #1816, opened 2019-06-18
fix(msg): use stdout when exiting with code 0

bricewge:
This PR change `polybar-msg` to output messages to stdout when the exit code is 0.

Currently wanting to only log errors coming from `polybar-msg` is very akward and non-unixy, to do so one is forced to write something like the following:
``` sh
err=$(polybar-msg hook module 1 2>&1) || echo "$err"`
# or
polybar-msg hook module 1 2>&1 | awk '!/Successfully wrote/ {print $0}'
```
With this change you can apply the conventional redirection pattern:
```sh
polybar-msg hook module 1 1>/dev/null
```

In my case not redirecting successful output result in a flood of the logs of ~10k message a day.

### DIFF
--- a/src/ipc.cpp
+++ b/src/ipc.cpp
@@ -15,8 +15,8 @@ using namespace std;
 #define IPC_CHANNEL_PREFIX "/tmp/polybar_mqueue."
 #endif
 
-void log(const string& msg) {
-  fprintf(stderr, "polybar-msg: %s\n", msg.c_str());
+void display(const string& msg) {
+  fprintf(stdout, "%s\n", msg.c_str());
 }
 
 void log(int exit_code, const string& msg) {
@@ -33,7 +33,7 @@ void remove_pipe(const string& handle) {
   if (unlink(handle.c_str()) == -1) {
     log(1, "Could not remove stale ipc channel: "s + strerror(errno));
   } else {
-    log("Removed stale ipc channel: " + handle);
+    display("Removed stale ipc channel: " + handle);
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/ipc.cpp
```diff
@@ -15,8 +15,8 @@ using namespace std;
 #define IPC_CHANNEL_PREFIX "/tmp/polybar_mqueue."
 #endif
 
-void log(const string& msg) {
-  fprintf(stderr, "polybar-msg: %s\n", msg.c_str());
+void display(const string& msg) {
+  fprintf(stdout, "%s\n", msg.c_str());
```
patrick96: I think we should keep the `polybar-msg: ` prefix.

```suggestion
  fprintf(stdout, "polybar-msg: %s\n", msg.c_str());
```

bricewge: This was intentional, identifying the origin of a message only make sense when writing to `stderr`. This [blog post](https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout) is good summary of why and how `stdout` and `stderr` are meant to be used.

If you insist I'll apply your suggestion though.

patrick96: Interesting. Though it doesn't say anything about `stdout` not needing to identify itself. But I see your point. I originally thought about us possibly adding more functionality to `polybar-msg` where polybar actually answers to certain commands and then we would actually need to identify the origin. But chances are that we're not even going to do anything like that ever, so I'll merge it like this.

PR #1823, opened 2019-06-24
fix: Only remove cloned monitors when necessary

### LINKED ISSUE #1191: Can't find monitor when running two displays in mirrored mode
When I run an external screen in mirrored mode (i.e. the same viewport as my main monitor), polybar no longer recognizes my main monitor and exits with:
```
◆ polybar -r example
error: Monitor "eDP-1" not found or disconnected
```

Here is the relevant part of my config:
```
[default]
```

XRandR still lists the monitor with the exact same name:

```
◆ xrandr --listmonitors
Monitors: 2
 0: +*eDP-1 2560/309x1440/174+0+0  eDP-1
 1: +HDMI-2 2560/677x1440/381+0+0  HDMI-2

◆ xrandr -q
Screen 0: minimum 320 x 200, current 2560 x 1440, maximum 8192 x 8192
eDP-1 connected primary 2560x1440+0+0 (normal left inverted right x axis y axis) 309mm x 174mm
   2560x1440     60.00*+
   // ....
HDMI-2 connected 2560x1440+0+0 (normal left inverted right x axis y axis) 0mm x 0mm
   1920x1080     60.00*+  50.00    59.94    30.00    25.00    24.00    29.97    23.98
   // ...
```

... but polybar does not recognize it:
```
◆ polybar --list-monitors
HDMI-2: 2560x1440+0+0
```

I am running the latest Git master with `WITH_XRANDR_MONITORS` enabled:
```
◆ polybar -v
polybar 3.1.0-110-gff39c9d-git

Features: -alsa +curl -i3 -mpd +network +pulseaudio
```

### LINKED ISSUE #1794: Polybar doesn't recognize VGA monitor width, causes error to launch
## Describe the issue
`polybar example` doesn't work because polybar doesn't recognize exteranal monitor width 

**Expected behavior:**

A working polybar

**Actual behavior:**

`polybar -m` output: 

```
VGA1: 0x900+0+0
LVDS1: 1600x900+0+0 (primary)

```
Width for some reason is zero

**Was it working before?**

No, I just installed it from AUR

## To Reproduce
polybar config
```
[bar/example]
monitor = VGA1
width = 5%
height = 27
;offset-x = 1%
;offset-y = 1%
radius = 6.0
fixed-center = false


```

## Polybar Log

```
$ polybar example
error: Resulting bar width is out of bounds (0)

```

## Environment:
* WM: i3-gaps
* Distro: Arch Linux
* Output of `polybar -vvv`: polybar 3.3.1

```

Build type: RelWithDebInfo
Compiler: /usr/bin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -Wall -Wextra -Werror -Wno-noexcept-type -pedantic -pedantic-errors
```

XRandr output:

```
$ xrandr -q | grep -w connected
LVDS1 connected primary 1600x900+0+0 (normal left inverted right x axis y axis) 310mm x 170mm
VGA1 connected 1440x900+0+0 (normal left inverted right x axis y axis) 410mm x 260mm   

$ xrandr --listmonitors
Monitors: 2
 0: +*LVDS1 1600/310x900/170+0+0  LVDS1
 1: +VGA1 1440/410x900/260+0+0  VGA1

```

patrick96:
When a monitor is connected which is automatically mirrored polybar can crash when the monitor name it is currently running gets removed in `get_monitors` (see #1191).

This also adds a new commandline argument `-M --list-all-monitors` that also lists cloned monitors. I couldn't change the behavior of `-m` since some people rely on `polybar -m` to display polybar on all monitors.

~I have not tested this yet but I will once I get some time at home.~

This also fixes the bugged clone removal code.

Fixes #1191
Fixes #1794

### DIFF
--- a/contrib/bash/polybar
+++ b/contrib/bash/polybar
@@ -31,10 +31,11 @@ _polybar() {
 	               -v --version
 	               -l --log=
 	               -q --quiet
-	               -c --config=
+                 -c --config=
 	               -r --reload
 	               -d --dump=
 	               -m --list-monitors
+	               -M --list-all-monitors
 	               -w --print-wmname
 	               -s --stdout
 	               -p --png='
--- a/contrib/bash/polybar
+++ b/contrib/bash/polybar
@@ -1,94 +1,94 @@
 _polybar_config_file() {
-	local config_path=${XDG_CONFIG_HOME:-$HOME/.config}/polybar/config
+  local config_path=${XDG_CONFIG_HOME:-$HOME/.config}/polybar/config
 
-	for ((i = 0; i < COMP_CWORD; i++)); do
-		case ${COMP_WORDS[i]} in
-		--config)
-			config_path=${COMP_WORDS[i + 2]}
-			break
-			;;
-		-c)
-			config_path=${COMP_WORDS[i + 1]}
-			break
-			;;
-		esac
-	done
+  for ((i = 0; i < COMP_CWORD; i++)); do
+    case ${COMP_WORDS[i]} in
+    --config)
+      config_path=${COMP_WORDS[i + 2]}
+      break
+      ;;
+    -c)
+      config_path=${COMP_WORDS[i + 1]}
+      break
+      ;;
+    esac
+  done
 
-	# Use eval + cd for to get bash's parameter/tilde expansion etc
-	(eval cd $(dirname "$config_path"); echo $PWD/$(basename "$config_path"))
+  # Use eval + cd for to get bash's parameter/tilde expansion etc
+  (eval cd $(dirname "$config_path"); echo $PWD/$(basename "$config_path"))
 }
 
 _polybar_bars() {
-	local config_file=$(_polybar_config_file)
+  local config_file=$(_polybar_config_file)
 
-	if [ -r "$config_file" ]; then
-		grep -Po '\[bar/\K(.*)(?=\])' "$config_file"
-	fi
+  if [ -r "$config_file" ]; then
+    grep -Po '\[bar/\K(.*)(?=\])' "$config_file"
+  fi
 }
 
 _polybar() {
-	local options='-h --help
-	               -v --version
-	               -l --log=
-	               -q --quiet
+  local options='-h --help
+                 -v --version
+                 -l --log=
+                 -q --quiet
                  -c --config=
-	               -r --reload
-	               -d --dump=
-	               -m --list-monitors
-	               -M --list-all-monitors
-	               -w --print-wmname
-	               -s --stdout
-	               -p --png='
+                 -r --reload
+                 -d --dump=
+                 -m --list-monitors
+                 -M --list-all-monitors
+                 -w --print-wmname
+                 -s --stdout
+                 -p --png='
 
-	local log_levels='error
-	                  warning
-	                  info
-	                  trace'
+  local log_levels='error
+                    warning
+                    info
+                    trace'
 
-	COMPREPLY=()
+  COMPREPLY=()
 
-	local cur=${COMP_WORDS[COMP_CWORD]}
-	case "$cur" in
-	-*)
-		COMPREPLY=( $(compgen -W "$options" -- "$cur") )
-		;;
-	*)
-		local prev=${COMP_WORDS[COMP_CWORD - 1]}
-		if [ "$prev" = "=" ]; then
-			prev=${COMP_WORDS[COMP_CWORD - 2]}
-		fi
+  local cur=${COMP_WORDS[COMP_CWORD]}
+  case "$cur" in
+  -*)
+    COMPREPLY=( $(compgen -W "$options" -- "$cur") )
+    ;;
+  *)
+    local prev=${COMP_WORDS[COMP_CWORD - 1]}
+    if [ "$prev" = "=" ]; then
+      prev=${COMP_WORDS[COMP_CWORD - 2]}
+    fi
 
-		case "$prev" in
-		-l|--log)
-			COMPREPLY=( $(compgen -W "$log_levels" -- "$cur") )
-			return 0
-			;;
-		-c|--config)
-			COMPREPLY=( $(compgen -f "$cur") )
-			return 0
-			;;
-		-p|--png)
-			COMPREPLY=( $(compgen -f -X "!*.png" "$cur") )
-			return 0
-			;;
-		# TODO: read properties of the selected bar from config
-		-d|--dump)
-			return 0
-			;;
-		*)
-			COMPREPLY=( $(compgen -W "$options $(_polybar_bars)" -- "$cur") )
-			;;
-		esac
-	esac
+    case "$prev" in
+    -l|--log)
+      COMPREPLY=( $(compgen -W "$log_levels" -- "$cur") )
+      return 0
+      ;;
+    -c|--config)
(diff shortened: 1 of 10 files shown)


contrib/bash/polybar
```diff
@@ -30,11 +30,11 @@ _polybar() {
 	local options='-h --help
 	               -v --version
 	               -l --log=
-	               -q --quiet
-	               -c --config=
+	               -q --quiet -c --config=
```
Lomadriel: Is this still the same thing as before (I'm not a bash expert)?

patrick96: That wasn't intentional, will upload a fixed version

Lomadriel:
As suspected in your comment: https://github.com/polybar/polybar/issues/1794#issuecomment-505194710
The code that removes the clone is bugged.
https://github.com/polybar/polybar/blob/c689330715f7f093c65505c38facbb2ba49433c5/src/x11/extensions/randr.cpp#L139-L141
When the erase is called, some iterators are invalidated.
https://github.com/polybar/polybar/blob/c689330715f7f093c65505c38facbb2ba49433c5/src/x11/extensions/randr.cpp#L144-L148
The iterator `m` is invalidated so it's undefined behavior to use it, even for doing something like `m++`.
See https://en.cppreference.com/w/cpp/container/vector (section "Iterator invalidation") for more information.
If you want I can fix that this weekend but feels free to do it if you can do it before. :wink:

patrick96:
@Lomadriel I already started working on a fix, the whole thing is a mess anyway, maybe I can clean it up a little.

patrick96:
Was able to clean up a little. The `get_monitors` method will be changed again anyway once we fix #1437 and correctly support xrandr's new monitor concept (which we don't really do right now).

Lomadriel:
https://github.com/polybar/polybar/blob/59007c3eb9ec4ce647c81d1f65d8c0caa06087b0/src/x11/extensions/randr.cpp#L179-L187

Maybe we should also fix this sort, It seems to be a "C comparator" copy/paste that haven't been adapted for `std::sort`

contrib/zsh/_polybar
```diff
@@ -21,7 +22,8 @@ _polybar() {
     "($C)"{-c,--config=}'[Path to the configuration file]:configuration file:_files' \
     "($R)"{-r,--reload}'[Reload when the configuration has been modified]' \
     "($D $R $M $W $S)"{-d,--dump=}'[Print parameter value in bar section and exit]:parameter name' \
-    "($M $D $R $W $S)"{-m,--list-monitors}'[Print list of available monitors and exit]' \
+    "($MM $M $D $R $W $S)"{-m,--list-monitors}'[Print list of available monitors and exit (Removes cloned monitors)]' \
```
NBonaparte: Might be better to say 'excluding' rather than 'removes'. Also should end with 'and exit' to keep everything consistent.

src/x11/extensions/randr.cpp
```diff
@@ -32,6 +32,16 @@ bool randr_output::match(const position& p) const {
   return p.x == x && p.y == y;
 }
 
+/**
+ * Checks if inner is contained withing this.
```
NBonaparte:
```suggestion
 * Checks if inner is contained within this.
```

NBonaparte:
@Lomadriel also why does it return -1 in the second case?

Lomadriel:
> @Lomadriel also why does it return -1 in the second case?

For me, it is just because originally it was a C comparator like `strcmp`. Anyway I don't know why the monitors are sorted by their size, I don't see the reason, so I can't figure what should be the comparator.

mjhough:
I was having the issue described in #1191. For me, this prevents polybar from crashing, but when I plug in my monitor this is what happens:
1. Plug in monitor, automatically in non-scaled mirroring mode. Polybar is on both monitors.
2. Run xrandr code (below) so that monitors are displayed side by side and the 1080p one is scaled.
```xrandr --output eDP-1 --primary --mode 2560x1440 --pos 3072x288 --rotate normal --output DP-1 --off --output DP-2 --off --output HDMI-1 --mode 1920x1080 --pos 0x0 --rotate normal --scale 1.6x1.6```

After step 2, I have polybar on my external monitor, even though in my config I have it set to eDP-1 (my laptop monitor). Killing polybar and starting it again brings it back to the correct screen.

patrick96:
> Maybe we should also fix this sort, It seems to be a "C comparator" copy/paste that haven't been adapted for std::sort.

@Lomadriel looking through the git history it seems jaagr copied over the sorting method from lemonbar to be consistent, I doubt that we still need this especially since we are now displaying on the primary monitor by default.

@NBonaparte Thanks for the review. I agree the wording could use some work, I'll look it over again.

patrick96:
@mjhough I think you are experiencing this because polybar uses the monitor position to make sure the bar is displayed on that monitor. In your case your bar is positioned at `0x0` and when you reconfigure your monitors, your bar stays at `0x0` and your main monitor moves to `3072x288`. We recommend setting `screenchange-reload = true` in the `[settings]` section.

mjhough:
> @mjhough I think you are experiencing this because polybar uses the monitor position to make sure the bar is displayed on that monitor. In your case your bar is positioned at `0x0` and when you reconfigure your monitors, your bar stays at `0x0` and your main monitor moves to `3072x288`. We recommend setting `screenchange-reload = true` in the `[settings]` section.
It seems I do have `screenchange-reload = true` in the [settings] section.

patrick96:
@mjhough Indeed, this seems like an oversight in the code. Polybar doesn't reload if monitor positions change, only when the dimensions change. I will fix this

patrick96:
I have now adapted the screenchange handling code to compare the monitor lists more thouroughly and reload if anything changed.

patrick96:
> Checked with upscale (nvidia drivers) #1437 is also fixed.

Were you able to reproduce the issue? Amazing! I thought we needed to rewrite much of the monitor handling code.

@NBonaparte what exactly is the restacking error? I'm unable to reproduce this.

Lomadriel:
> Were you able to reproduce the issue? Amazing! I thought we needed to rewrite much of the monitor handling code.

The issue is fixed because we reload polybar when a monitor change ;)

patrick96:
I think #1437 happened even if polybar is restarted.

Lomadriel:
> What was the issue you thought of?

When I was modifying my resolution, polybar wasn't reloading and so was misplaced/out of the screen, etc...

patrick96:
Ah, I see.

NBonaparte:
@patrick96 just tested this again, I'm only getting one monitor with `--list-monitors` which is expected, but I still get the restacking error: `error: Failed to restack bar window`.

I have two monitors, HDMI-0 and DP-1, with bspwm. For testing I set DP-1 to be a clone of HDMI-0 (`xrandr --output DP-1 --same-as HDMI-0`) and in the config I set `monitor = DP-1`.

patrick96:
@NBonaparte I still can't reproduce this. The bspwm restacking mechanic doesn't rely on the monitor name. It searches for a bspwm root window with the exact same coordinates as the monitor that polybar is displayed on, so this patch changes nothing about this behavior. On my machine bspwm has one root window per monitor, even if they are cloned.

I am going to merge this, if you want could you share the output of `xwininfo -root -tree` and `polybar -M` when this happens?

PR #1825, opened 2019-06-25
feat(github): offline label & fixes

### LINKED ISSUE #1815: Github module displays label when offline
## Describe the issue
When offline the github module displays regardless of `empty-notifications = false`. 

**Expected behavior:**

The module either displays only when the last query resulted in a non-zero number of notifications or does not display even when offline (alternatively, allows for a custom offline label to be set).

**Actual behavior:**

Github module always displays when offline.

**Was it working before?**

Same behavior since I've first used github module (~3 weeks).

* Did you also experience this bug in an earlier version of polybar?

Yes.

* If no, what was the last version where this worked correctly?

## To Reproduce
```dosini
[colors]
background = #222
background-alt = #444
foreground = #dfdfdf
foreground-alt = #888
primary = #ffb52a
secondary = #e60053
alert = #bd2c40

[bar/main]
width = 100%
height = 2.5%
radius = 6.0
fixed-center = false
bottom = true

background = ${colors.background}
foreground = ${colors.foreground}

border-size = 0
line-size = 2
padding = 1
module-margin = 1

font-0 = "Noto Sans Mono:size=10;1"
font-1 = "Noto Sans Mono:size=10;1"
font-2 = "Noto Sans Symbols:size=10;1"
font-3 = "Noto Sans Symbols2:size=10;1"

modules-left = 
modules-center =
modules-right = github

tray-position = right
tray-padding = 2
tray-maxsize = 24

[global/wm]
margin-top = 0

[module/github]
type = internal/github
token = ${file:~/.config/github/notifications.token}
empty-notifications = false
;label = %{u#ff9900 +u} %notifications% %{u-}
label = %{u#0099ff +u} 📫 %{u-}
interval = 10

[settings]
screenchange-reload = true

; vim:ft=dosini

```
*List any other steps needed to reproduce the issue besides starting polybar with the config you posted above.*

Disconnect any internet connection.

## Polybar Log
```
warn: module/github: cannot complete the request to github: Couldn't connect to server
```

## Screenshots
*If applicable, add screenshots to help explain your problem.*
With full config:
* Online (No notifications):
2019-06-17-165021_1919x23_scrot
* Offline:
2019-06-17-165110_1919x22_scrot
## Environment:
* WM: xmonad 0.15-12
* Distro: Arch Linux
* Output of `polybar -vvv`:

polybar 3.3.1

Features: +alsa +curl -i3 -mpd +network(libnl) +pulseaudio +xkeyboard

Build type: RelWithDebInfo
Compiler: /usr/bin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -Wall -Wextra -Werror -Wno-noexcept-type -pedantic -pedantic-errors

## Additional context
*Add any other context that you think is necessary about the problem here.*
Appears to occur as a result of drawing whenever the condition in line 81 of github.cpp is met (`0 != notifications || m_empty_notifications`), in conjunction with the choice to return `-1` from `get_number_of_notifications` when a non 200 http response is received and passing that directly to `update_label`.

kopecs:
Fixes #1815.

Should resolve the segfault issue with #910. The defaults are such that if offline the module displays `Offline` instead of the current `Notifications: -1`. Notifications are still set to -1 and a flag is set to indicate online/offline status which causes `format-offline` to be used when offline.

I haven't tested exhaustively but everything seemed to be working with a couple different configs on my machine.

MAINTAINER EDIT:
Closes #910

### DIFF
--- a/include/modules/github.hpp
+++ b/include/modules/github.hpp
@@ -16,18 +16,24 @@ namespace modules {
 
     bool update();
     bool build(builder* builder, const string& tag) const;
+    string get_format() const;
+
 
    private:
     void update_label(int);
     int get_number_of_notification();
     string request();
     static constexpr auto TAG_LABEL = "<label>";
+    static constexpr auto TAG_LABEL_OFFLINE = "<label-offline>";
+    static constexpr auto FORMAT_OFFLINE = "format-offline";
 
     label_t m_label{};
+    label_t m_label_offline{};
     string m_api_url;
     string m_accesstoken{};
     unique_ptr<http_downloader> m_http{};
     bool m_empty_notifications{false};
+    std::atomic<bool> m_offline{false};
   };
 }
 
--- a/src/modules/github.cpp
+++ b/src/modules/github.cpp
@@ -1,10 +1,10 @@
+#include "modules/github.hpp"
+
 #include <cassert>
 
 #include "drawtypes/label.hpp"
-#include "modules/github.hpp"
-#include "utils/concurrency.hpp"
-
 #include "modules/meta/base.inl"
+#include "utils/concurrency.hpp"
 
 POLYBAR_NS
 
@@ -29,9 +29,15 @@ namespace modules {
 
     if (m_formatter->has(TAG_LABEL)) {
       m_label = load_optional_label(m_conf, name(), TAG_LABEL, "Notifications: %notifications%");
-
-      update_label(0);
     }
+
+    m_formatter->add(FORMAT_OFFLINE, TAG_LABEL_OFFLINE, {TAG_LABEL_OFFLINE});
+
+    if (m_formatter->has(TAG_LABEL_OFFLINE)) {
+      m_label_offline = load_optional_label(m_conf, name(), TAG_LABEL_OFFLINE, "Offline");
+    }
+
+    update_label(0);
   }
 
   /**
@@ -68,9 +74,14 @@ namespace modules {
     try {
       content = request();
     } catch (application_error& e) {
-      m_log.warn("%s: cannot complete the request to github: %s", name(), e.what());
+      if (!m_offline) {
+        m_log.info("%s: cannot complete the request to github: %s", name(), e.what());
+      }
+      m_offline = true;
       return -1;
     }
+
+    m_offline = false;
 
     size_t pos{0};
     size_t notifications{0};
@@ -82,12 +93,14 @@ namespace modules {
     return notifications;
   }
 
+  string github_module::get_format() const {
+    return m_offline ? FORMAT_OFFLINE : DEFAULT_FORMAT;
+  }
+
   void github_module::update_label(const int notifications) {
-    if (0 != notifications || m_empty_notifications) {
+    if ((0 != notifications || m_empty_notifications) && m_label) {
       m_label->reset_tokens();
       m_label->replace_token("%notifications%", to_string(notifications));
-    } else {
-      m_label->clear();
     }
   }
 
@@ -95,12 +108,16 @@ namespace modules {
    * Build module content
    */
   bool github_module::build(builder* builder, const string& tag) const {
-    if (tag != TAG_LABEL)
-      return false;
+    if (tag == TAG_LABEL) {
+      builder->node(m_label);
+      return true;
+    } else if (tag == TAG_LABEL_OFFLINE) {
+      builder->node(m_label_offline);
+      return true;
+    }
 
-    builder->node(m_label);
-    return true;
+    return false;
   }
-}
+}  // namespace modules
 
 POLYBAR_NS_END
--- a/src/modules/github.cpp
+++ b/src/modules/github.cpp
@@ -98,9 +98,16 @@ namespace modules {
   }
 
   void github_module::update_label(const int notifications) {
-    if ((0 != notifications || m_empty_notifications) && m_label) {
+    if (!m_label) {
+      return;
+    }
+
+    if (0 != notifications || m_empty_notifications) {
       m_label->reset_tokens();
       m_label->replace_token("%notifications%", to_string(notifications));
+    }
+    else {
+      m_label->clear();
     }
   }
 
--- a/src/modules/github.cpp
+++ b/src/modules/github.cpp
@@ -105,8 +105,7 @@ namespace modules {
     if (0 != notifications || m_empty_notifications) {
       m_label->reset_tokens();
       m_label->replace_token("%notifications%", to_string(notifications));
-    }
-    else {
+    } else {
       m_label->clear();
     }
   }


include/modules/github.hpp
```diff
@@ -16,17 +16,23 @@ namespace modules {
...
     static constexpr auto TAG_LABEL = "<label>";
+    static constexpr auto TAG_LABEL_OFFLINE = "<label-offline>";
+    static constexpr auto FORMAT_OFFLINE = "format-offline";
 
     label_t m_label{};
+    label_t m_label_offline{};
     string m_accesstoken{};
     unique_ptr<http_downloader> m_http{};
     bool m_empty_notifications{false};
+    bool m_offline{false};
```
Lomadriel:
```suggestion
    std::atomic<bool> m_offline{false};
```
Since this variable is accessed from the thread of the module and the main thread, it must be protected to avoid some problems.

patrick96:
Thanks for reviewing @Lomadriel. I think we should stay with `format` and `label` since those represent the default state. Having `"Offline"` as the default should be fine. Though mpd actually has a default of `""` for `format-offline`, so you only need to set `label-offline` when you set `format-offline`, I think that is counter intuitive so this approach here seems better since it gives sane defaults.

src/modules/github.cpp
```diff
@@ -24,9 +24,15 @@ namespace modules {
...
+    }
+
+    m_formatter->add(FORMAT_OFFLINE, TAG_LABEL_OFFLINE, {TAG_LABEL_OFFLINE});
 
-      update_label(0);
+    if (m_formatter->has(TAG_LABEL_OFFLINE)) {
+      m_label_offline = load_optional_label(m_conf, name(), TAG_LABEL_OFFLINE, "Offline");
     }
+
+    update_label(0);
```
Lomadriel: I'm not sure if we should keep this line.
Indeed it prevents to display the label without applying the data but this isn't done in any module.
Maybe, we should address this globally rather that adding quick fixes in every modules.

kopecs: As that was already the case, perhaps this isn't the best pr to address changing it in? Other than that the issues you noted should be fixed.

Lomadriel: I think we can let this line for now. When I'll post a PR that solves these issues I'll remove it.

src/modules/github.cpp
```diff
@@ -64,9 +70,12 @@ namespace modules {
       content = request();
     } catch (application_error& e) {
       m_log.warn("%s: cannot complete the request to github: %s", name(), e.what());
```
Lomadriel: I didn't think about that before, but since "offline" is now a normal state, maybe we should replace this line by:
```suggestion
      m_log.info("%s: cannot complete the request to github: %s", name(), e.what());
```

What do you think of that @patrick96?

kopecs:
Any more steps we need to take to get this merged?

Lomadriel:
> Any more steps we need to take to get this merged?

This PR will be merged soon but the core team is a little busy.
I think your PR will be integrated into the 3.5.0 since we have already started to package the 3.4.0.

kopecs:
Sounds good! Thanks for the update!

Lomadriel:
Hello, one of my PR (#1841) just get merged and it introduced a conflict with yours.
Can I ask you to resolve it?

kopecs:
Can you confirm that change resolves any conflict? It looks like we just both added a new member in the same place.

Lomadriel:
Looks fine, that was the only conflict.
Thanks !

src/modules/github.cpp
patrick96: Good idea, we don't want to spam users with warnings that aren't really warnings.

src/modules/github.cpp
```diff
@@ -69,9 +75,12 @@ namespace modules {
...
```
patrick96:
```suggestion
      m_log.info("%s: cannot complete the request to github: %s", name(), e.what());
```

kopecs: Resolved in cb83b83

kopecs:
Best I can tell the segfault occurs as a result of attempting to clear the label in `update_label` when the `format` property is of certain (invalid?) format, e.g., text without any labels. Is there documentation on the grammar of the `format` property besides the description [here](https://github.com/polybar/polybar/wiki/Formatting)? I'm not totally sure what the resolution here should be; if `m_label` is a `nullptr` should I just immediately return from `update_label` with a warning, or is another solution preferred?

patrick96:
Oh that segfault isn't even a result of your PR, but was there from the beginning. In other modules we always first check if a label is valid before dereferencing it, for example in the cpu module:

https://github.com/polybar/polybar/blob/4ef2bd55751591026183ad44a4e9cf5bd9484392/src/modules/cpu.cpp#L69-L78

We shouldn't print a warning since not having a label is completely valid (though not really useful)

As far as I can tell you only need to update `update_label`. However, this isn't really a bug in your PR, so we can also just merge this as-is.

We don't really have any developer documentation on the formats. But a format can contain an arbitrary string. Occurences of certain label, ramp, bar, animation names inside angle brackets (`<`, `>`) are treated as referencing the contents of that label, ramp etc. So `format = foo` is just as valid as `format = foo <label> bar` or just `format = <label>`.

> As an additional note, I'm assuming any build issues currently present are a result of google/googletest#2678.

Yes, the test build fails because of that. It's fixed on master.

kopecs:
@patrick96 I've made the requested change about the logging level. I also added checks to confirm there aren't any `nullptr` dereferences (although if you'd prefer those are fixed separately I can roll it back). Is there any more action to be take at the moment?

kopecs:
Yeah, the most recent revision should be sufficient. I just looked at it again today and I can't think of any reason I'd need to do that and it seems fine without.

> Because I would have thought checking if `m_label` is null and directly returning at the start of `update_label` would be enough.

In this case wouldn't I still need to check `m_offline_label`, or if one is non-null are they both set? I'm not totally clear on what holds with the label setinigs.

patrick96:
> In this case wouldn't I still need to check m_offline_label

You only need to check `m_offline_label` before you dereference it. And there is no need to call `reset_tokens` on `m_offline_label` because it doesn't have any tokens. So you never derference it. You pass it to the builder though, but there the builder does its own checks.

kopecs:
The most recent commit shouldn't have any `NULL` dereferences and doesn't clear tokens for things that don't have them.

src/modules/github.cpp
```diff
@@ -68,10 +74,13 @@ namespace modules {
     try {
       content = request();
     } catch (application_error& e) {
-      m_log.warn("%s: cannot complete the request to github: %s", name(), e.what());
+      m_log.info("%s: cannot complete the request to github: %s", name(), e.what());
```
patrick96:
```suggestion
      if (!m_offline) {
        m_log.info("%s: cannot complete the request to github: %s", name(), e.what());
      }
```

I think we should only show the message when we are no longer online, otherwise it would spam the logs for offline users.

kopecs:
No problem! Nit picks appreciated.

kopecs:
Let me know if there's still an issue with merging; I pulled master into this branch and it seemed to resolve it fine but I hadn't and don't see anywhere on github where it's indicating issues with merging to me (modulo that fact that I have changes requested).

patrick96:
This is starting to make me mad. Now there is a merge conflict agian :angry:

patrick96:
Not even cherry-picking the individual commits in this PR worked. I have now taken the diff of this entire PR and just applied it to our master branch and just force pushed. I have also noticed that we missed something: When empty-notifications is false, the label is supposed to be empty if there are no notifications, but this PR removed the call to `clear`, if have also fixed this.

kopecs:
Sorry that this was such a headache to get done. Thanks for the update!

PR #1830, opened 2019-06-30
add ability to filter xworkspace names with regex

maxaudron:
new property filter: takes regex expression and
 matches them against the workspace names filtering
 out all that don't match

### DIFF
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -102,6 +102,7 @@ namespace modules {
     bool m_pinworkspaces{false};
     bool m_click{true};
     bool m_scroll{true};
+    string m_filter{""};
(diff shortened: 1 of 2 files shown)


patrick96:
Is this ready to be reviewed or is there still something that you need to change?

I think this approach breaks the `%index%` token and the click functionality (which relies on the workspace index) since the index is calculated based on the size of the `m_desktop_names` list which is changed here.

maxaudron:
Yep i noticed that shortly after opening the pull request. I'll rework it hopefully by the weekend

maxaudron:
@patrick96 should be good to go now

maxaudron:
im closing this as i am no longer in need for it and never got it to fully work in the first place.

PR #1831, opened 2019-07-01
fix(config): Allow empty string as ref fallback

patrick96:
There is also a lot of refactoring that could be done here but since we are rewriting parts of the config parser already we'll do it there.

### DIFF
--- a/include/components/config.hpp
+++ b/include/components/config.hpp
@@ -268,9 +268,15 @@ class config {
   T dereference_env(string var) const {
     size_t pos;
     string env_default;
+    /*
+     * This is needed because with only the string we cannot distinguish
+     * between an empty string as default and not default
+     */
+    bool has_default = false;
 
     if ((pos = var.find(':')) != string::npos) {
       env_default = var.substr(pos + 1);
+      has_default = true;
       var.erase(pos);
     }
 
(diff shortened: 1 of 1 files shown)


include/components/config.hpp
```diff
@@ -268,17 +268,23 @@ class config {
...
+     */
+    bool has_default = false;
 
     if ((pos = var.find(':')) != string::npos) {
       env_default = var.substr(pos + 1);
+      has_default = true;
       var.erase(pos);
     }
 
     if (env_util::has(var.c_str())) {
```
Lomadriel:
```suggestion
    if (env_util::has(var)) {
```

include/components/config.hpp
```diff
@@ -268,17 +268,23 @@ class config {
...
       string env_value{env_util::get(var.c_str())};
```
Lomadriel:
```suggestion
      string env_value{env_util::get(var)};
```

patrick96:
> I also have a question, does it make sense to authorize empty file name?

Not really, why?

Lomadriel:
Mostly because of your modification in `dereference_file` but I think it makes sense to be consistent and allow empty fallback everywhere.

patrick96:
The fallback in `dereference_file` is an actual value for when the file isn't found not another path.

Lomadriel:
> The fallback in `dereference_file` is an actual value for when the file isn't found not another path.

:man_facepalming: Indeed... (sorry)

PR #1836, opened 2019-07-18
aur: Make feature deps hard deps

patrick96:
In the AUR optdepends means that the package can run without optdepends
installed. In polybar most features, if enabled at compile time, cannot
run without their dependencies and will crash polybar. Now the
optdepends only contains truly optional dependencies.

Polybar can run without the i3-wm package because it only relies on the
`i3` executable and is not dynamically linked against any library in i3.

The main reason I didn't want to do this before is that I didn't want to force people to have i3 installed to run polybar, but since i3 is actually truly optional this should be not too bad since most people already have those dependencies and they're not too big.

Before we merge this I'd like to also downgrade the `pulseaudio` dependency to only `libpulse` if possible. Though polybar checks for the `pulseaudio` binary in `02-opts.cmake`, it doesn't use it anywhere.


Lomadriel:
I agree this is a good idea for now. We'll have to "rollback" this when we'll introduce the separation in shared objects but that's better for now.

patrick96:
> We'll have to "rollback" this when we'll introduce the separation in shared objects but that's better for now

True, we will then move all dependencies back to `optdepends` and also add them to `makedepends`.

I have now removed the check for the `pulseaudio` binary.

PR #1841, opened 2019-07-26
feat(github): add support for github enterprise

### LINKED ISSUE #1812: Support for GitHub Enterprise in github module
It would be great to have that work offloaded to github module instead of using curl scripts.

I guess the only change would be a URL string, pointing to github enterprise installation.

So https://github.com/polybar/polybar/blob/3f9f2dc37fd8a82b7f7c7688d9bc633c67e55738/src/modules/github.cpp#L44

would become

```c++
 string content{m_http->get("https://hub.domain.tld/api/v3/notifications?access_token=" + m_accesstoken)}; 
```

Lomadriel:
Add support for github enterprise with a new key `api-url`.

This feature needs to be checked properly since I don't have access to the github of an entreprise.
I just checked that it works correctly with `github.com`.

Closes #1812.

### DIFF
--- a/include/modules/github.hpp
+++ b/include/modules/github.hpp
@@ -18,7 +18,7 @@ namespace modules {
     bool build(builder* builder, const string& tag) const;
 
    private:
-    void update_label(const int);
+    void update_label(int);
     int get_number_of_notification();
     string request();
     static constexpr auto TAG_LABEL = "<label>";
--- a/include/modules/github.hpp
+++ b/include/modules/github.hpp
@@ -24,6 +24,7 @@ namespace modules {
     static constexpr auto TAG_LABEL = "<label>";
 
     label_t m_label{};
+    string m_api_url;
     string m_accesstoken{};
     unique_ptr<http_downloader> m_http{};
     bool m_empty_notifications{false};
--- a/src/modules/github.cpp
+++ b/src/modules/github.cpp
@@ -16,8 +16,6 @@ namespace modules {
    */
   github_module::github_module(const bar_settings& bar, string name_)
       : timer_module<github_module>(bar, move(name_)), m_http(http_util::make_downloader()) {
-    using namespace std::string_literals;
-
     m_accesstoken = m_conf.get(name(), "token");
     m_api_url = m_conf.get(name(), "api-url", "https://api.github.com/"s);
     if (m_api_url.back() != '/') {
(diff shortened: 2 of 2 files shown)


src/modules/github.cpp
```diff
@@ -16,7 +16,14 @@ namespace modules {
    */
   github_module::github_module(const bar_settings& bar, string name_)
       : timer_module<github_module>(bar, move(name_)), m_http(http_util::make_downloader()) {
+    using namespace std::string_literals;
```
patrick96: I think this is already included via `common.hpp`

Lomadriel: Yeah indeed, I'll remove that.

Lomadriel:
> Looks good otherwise. We probably don't need to validate that it is a proper URL.

The aim isn't validate the URL.
It's to support both: `http://api.github.com/` and `http://api.github.com` and to avoid guessing if you need to put (or not) the final `/`

patrick96:
>  It's to support both: http://api.github.com/ and http://api.github.com and to avoid guessing if you need to put (or not) the final /

Oh, that's not what I meant. It was just a general comment that we probably don't want to add extra validation code.

We should definitely force a `/` at the end there ;)

Lomadriel:
> We should definitely force a / at the end there ;)

Okay. I'll add an exception.

> Oh, that's not what I meant. It was just a general comment that we probably don't want to add extra validation code.

Ha my bad but I think curl does that verification for us no? (See the `get` method) If it isn't a valid URL an exception is thrown, I think.

patrick96:
> Okay. I'll add an exception.

What do you mean? This is already in the code

> I think curl does that verification for us no?

Probably, yeah

Lomadriel:
> What do you mean? This is already in the code

~~Currently, if there is no `/`. I add it, but since we should force having a `/` at the end then I should add an exception no?~~

Damn I think I misunderstood everything... By 'we' you wanted to say 'polybar/ the code'...
Sorry

patrick96:
Yeah, I meant the code should add the `/`, just as it is now ;)

Lomadriel:
Updated to take the review into account.

include/modules/github.hpp
```diff
@@ -18,12 +18,13 @@ namespace modules {
     bool build(builder* builder, const string& tag) const;
 
    private:
-    void update_label(const int);
+    void update_label(int);
```
patrick96: Is this from clang-format?

Lomadriel: clang-tidy rather than clang-format but yes it's from a clang tool ;)

Issue #1000, opened 2018-02-02
debian documentation is missleading 


icf20:
> If you are using Debian, polybar is available from the GetDeb repository.

there is no release for debian only for ubuntu http://www.getdeb.net/app/Polybar

keithieopia:
See the [Don't make a FrankenDebian](https://wiki.debian.org/DontBreakDebian#Don.27t_make_a_FrankenDebian) page on why using Ubuntu repos in Debian is bad. Honestly, the documentation should just say to build it from source for Debian.

patrick96:
You two are right. I just opened a PR to change the Debian mention to Ubuntu.

icf20:
I don't know if u got the time to look at this, but if I remember correctly Debian stable don't have the lib's version required only Debian unstable/testing have them

patrick96:
Unfortunately I don't have that kind of time. But as far as I remember, no one has had problems locating the dependent packages on debian, so I think we're good. If someone does at some point, it might be worth adding it to the wiki

PR #1846, opened 2019-08-03
travis: create an irc one-liner

x70b1:
Combine the information into one line.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,8 +101,7 @@ notifications:
     channels:
       - "irc.freenode.org#polybar"
     template:
-      - "%{author} pushed #%{commit} \"%{commit_subject}\" to %{repository_name}/%{branch}"
-      - "Build #%{build_number} %{result} / %{build_url} (diff: %{compare_url})"
+      - " %{repository_slug}(%{branch})#%{build_number} | \"%{commit_subject}\" by %{author} | Commit #%{commit} %{result}"
     use_notice: true
     on_success: change
     on_failure: change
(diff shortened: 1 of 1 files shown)


Lomadriel:
So if I'm not wrong.
`" %{repository_slug}(%{branch})#%{build_number} | \"%{commit_subject}\" by %{author} | Commit #%{commit} %{result}"`
would result in something like this:
`polybar/polybar(develop)#1000 | "travis: create an irc one-liner" by x70b1 | Commit #sha success`

Wouldn't it be better to use `%{repository_name}` rather than `%{repository_slug}` since in theory, the organization name is always the same?

Otherwise I'm not using the IRC so much (for now), so I'll let the review to @patrick96 but it seems good to me.

x70b1:
The problem is that even forked projects will start posting in the IRC channel if they activate travis.

The argument for the slug is that it is the only way to quickly recognize that it is the repository from the organization.

Lomadriel:
Indeed, we should use `%{repository_slug}` in this case. Is there a way to block this?
I don't think we really want forked projects to push notification on our IRC channel if they activate travis.

x70b1:
At the moment there is no simple solution. But it's not a big problem either, because forked projects very rarely enable Travis.

x70b1:
I tested that already here: https://github.com/polybar/polybar-scripts/commit/313ff76cb66a8600ca6f6f1a8d39cec4f7324a87. 

It didn't work in this first attempt. Patrick had no success too at this. If its gonna be a problem, we could try it again. We decided its ok for the moment.

Lomadriel:
> It didn't work in this first attempt. Patrick had no success too at this. If its gonna be a problem, we could try it again. We decided its ok for the moment.

I agree with you, I don't think this will cause problems for now.

x70b1:
:rocket:

PR #1863, opened 2019-09-02
i3: Add possibility to sort workspaces by name

epontan:
In i3 it is possible to assign workspaces with arbitrary names without
assigning it a specific index. Polybar however, is only able to sort the
workspaces by index, leaving named workspaces in seemingly random order
received from i3.

Make it possible to configure sorting on workspace names by adding an
option ("name-sort") for the i3 module.

### DIFF
--- a/include/modules/i3.hpp
+++ b/include/modules/i3.hpp
@@ -88,6 +88,7 @@ namespace modules {
     bool m_revscroll{true};
     bool m_wrap{true};
     bool m_indexsort{false};
+    bool m_namesort{false};
     bool m_pinworkspaces{false};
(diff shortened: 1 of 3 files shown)


patrick96:
Hey thanks for contributing!

I think in i3 the idea is to prefix named workspaces [with a number and a colon](https://i3wm.org/docs/userguide.html#_named_workspaces) to give it an index as well as a name. With this you can define an order on your workspaces. In polybar you could then set `strip-wsnumbers = true` to only display the name.

Is there some usecase that I am not seeing which wouldn't work with the approach described but only with your PR?

patrick96:
Closing due to inactivity

PR #1864, opened 2019-09-04
Check for OpenBSD and set -Wno-c99-extensions

jasperla:
Please find below a patch to get polybar building on OpenBSD.

The basics work fine, though many modules lack support for OpenBSD (e.g. due to absence of `/proc`) but I'll have a look at addressing this in due time.

### DIFF
--- a/cmake/01-core.cmake
+++ b/cmake/01-core.cmake
@@ -32,6 +32,11 @@ if (CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -L/usr/local/lib")
 endif()
 
+if (CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
+  # libinotify uses c99 extension, so suppress this error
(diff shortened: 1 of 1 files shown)


Lomadriel:
Is `-Werror` activated by default on OpenBSD?
If I'm not wrong we removed it from the CMake not so long ago so it should compile fine.


> The basics work fine, though many modules lack support for OpenBSD (e.g. due to absence of `/proc`) but I'll have a look at addressing this in due time.

This might code later but I don't think polybar is officially support OpenBSD. (I might be wrong)

jasperla:
You're right, I think this was a left-over patch from pre-clang. Please disregard this one.

And while polybar may not officially support OpenBSD, I'd like to see what can be implemented in terms of support for OpenBSD in the various modules.

patrick96:
Our position on supporting non-linux OSes is that it's only incidental. If an OS can be supported by small changes to the build system we'll likely add it but we don't actively test on those OSes and we're very reluctant to change the actual code to support other OSes. There is however https://github.com/polybar/polybar/issues/1529 which is a plan to provide some way to more easily support modules in other OSes, this is however only a draft and nothing concrete yet.

PR #1891, opened 2019-10-03
doc: Write syntax definition in polybar.5

patrick96:
And finally the documentation for the config file syntax defined in #1377


patrick96:
Thanks @kronn 😊

PR #1893, opened 2019-10-05
Add a tokenize function to take empty tokens into account

### LINKED ISSUE #1881: string_util::split_into ignores final delimiter
## Describe the issue
**Expected behavior:**

string_util::split_into should not ignore the final delimiter.

The string `foo;bar;` should be split into `["foo", "bar", ""]`

**Actual behavior:**

The string `foo;bar;` is split into `["foo", "bar"]`

**Was it working before?**
I don't think this is a regression. It may even be the intended behaviour. There are a couple of reasons why I think this should be changed:

1. This is not consistent with some other string splitting functions. I have tested the Python string.split as well as boost::split.

2. Declaring an empty i3 (and possibly others? - I only checked i3) workspace icon does not use the expected syntax. I.e. using `ws-icon-0 = foo;` will cause workspaces called foo to use `ws-icon-default` rather than an empty value. My current workaround is to do `ws-icon-0 = foo;;`

## Environment:
* WM: i3-gaps
* Distro: Arch Linux
* Output of `polybar -vvv`:
```
polybar 3.4.0

Build type: Release
Compiler: /usr/bin/c++
```

If you accept this as a bug, I can submit a PR to fix it.

Lomadriel:
Resolves #1881.

I didn't use the new function everywhere most of the time we need to ignore empty tokens.

### DIFF
--- a/tests/unit_tests/utils/string.cpp
+++ b/tests/unit_tests/utils/string.cpp
@@ -65,19 +65,39 @@ TEST(String, join) {
 }
 
 TEST(String, splitInto) {
-  vector<string> strings;
-  string_util::split_into("A,B,C", ',', strings);
-  EXPECT_EQ(3, strings.size());
-  EXPECT_EQ("A", strings[0]);
-  EXPECT_EQ("C", strings[2]);
+  {
+    vector<string> strings = string_util::split("A,B,C", ',');
+    EXPECT_EQ(3, strings.size());
+    EXPECT_EQ("A", strings[0]);
+    EXPECT_EQ("B", strings[1]);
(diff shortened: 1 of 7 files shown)


tests/unit_tests/utils/string.cpp
```diff
@@ -65,19 +65,39 @@ TEST(String, join) {
 }
 
 TEST(String, splitInto) {
```
patrick96:
```suggestion
TEST(String, split) {
```

Lomadriel:
Take review into account and rebase to resolve the config.

PR #1901, opened 2019-10-15
fix(bar): Configure position before remapping

### LINKED ISSUE #1484: Polybar shows on wrong monitor after hide
## Describe the issue
When the polybar hides and then shows again it appears on the wrong monitor. 

**Expected behavior:**

The hidden polybars should be appearing on the same monitor on which they were hidden.

**Actual behavior:**

Running `polybar-msg cmd toggle` twice makes all the polybars from all my monitors  show up on a single monitor.

## To Reproduce
```dosini
# File: ~/.config/polybar/config
[bar/middle]
monitor = ${env:MONITOR_MIDDLE}

[bar/right]
inherit = bar/middle
monitor = ${env:MONITOR_RIGHT}
```

```bash
# File: ~/.profile
# Monitors
export MONITOR_MIDDLE=DP-2
export MONITOR_RIGHT=DVI-D-0
```

## Screenshots
polybars-screenshot

## Environment:
* WM: i3-gaps
```
polybar 3.2.1

Features: +alsa +curl +i3 +mpd +network +pulseaudio +xkeyboard

X extensions: +randr (+monitors) -render -damage -sync -composite +xkb +xrm +xcursor

Build type: RelWithDebInfo
Compiler: /usr/sbin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -Wall -Wextra -Werror -Wno-noexcept-type -O2 -pedantic -pedantic-errors
```

patrick96:
Some WMs like i3 discard position information when unmapping the bar and
because of that the bar would be at the wrong position after being
remapped.

Fixes #1484
Ref: https://github.com/i3/i3/pull/3834

### DIFF
--- a/include/components/bar.hpp
+++ b/include/components/bar.hpp
@@ -74,6 +74,8 @@ class bar : public xpp::event::sink<evt::button_press, evt::expose, evt::propert
 
  protected:
   void restack_window();
+  void reconfigue_window();
+  void reconfigure_geom();
   void reconfigure_pos();
   void reconfigure_struts();
   void reconfigure_wm_hints();
--- a/src/components/bar.cpp
+++ b/src/components/bar.cpp
@@ -403,6 +403,11 @@ void bar::show() {
   try {
     m_log.info("Showing bar window");
     m_sig.emit(visibility_change{true});
+    /**
+     * First reconfigures the window so that WMs that discard some information
+     * when unmapping have the correct window properties (geometry etc).
+     */
+    reconfigue_window();
     m_connection.map_window_checked(m_opts.window);
     m_connection.flush();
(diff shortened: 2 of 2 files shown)


patrick96:
I have now changed it to reconfigure the entire window (restacking, struts, wm hints, geometry) to be on the safe side. Please review again.

patrick96:
> It is okay for me but maybe we should put the reconfigure_window in map_window_checked (and rename it) to avoid any problem if we add a map_window_checked somewhere.

I first started out doing this. But I couldn't come up with a situation where we would need to unmap and remap the bar, so I didn't put it in a function.

Issue #845, opened 2017-11-15
module/menu question


Choromanski:
I was wondering if there was the ability in the menu module such that when I click on menu-2-0 for instance, it executes a command and doesn’t close the menu. I have tried to get this working for some time with no luck.

patrick96:
I'm not aware of any such feature, but this could be useful. I'll add this to our todo list.

Choromanski:
Thank you, I figured that would be a cool feature to have

PR #1907, opened 2019-10-21
Use module name in action string

### LINKED ISSUE #1172: Multiple MPD modules: second one controls wrong server

patrick96:
**Release Note:**
Once this is part of a release we can do the following:

* Remove ["With multiple modules of the same type, actions on those modules control the leftmost module"](https://github.com/polybar/polybar/wiki/Known-Issues#with-multiple-modules-of-the-same-type-actions-on-those-modules-control-the-leftmost-module) from the known issues
* Remove the note on the menu module page that only a single menu module can exist.
* Add a note about migration to all the wiki pages that use the legacy action names:
  * i3 module
  * bspwm module
  * menu module
  * ipc (not the module)
  * (maybe others)

This PR changes the behavior of the `prev` and `next` actions in the i3 and bspwm module (formerly `i3wm-wsnext`, `i3wm-wsprev`, `bspwm-desknext`, and `bspwm-deskprev`).
The old behavior of those actions was affected by the `reverse-scroll` setting. This behavior is very counterintuitive and it makes more sense that `next` and `prev` select the next and previous workspace respectively. `reverse-scroll` is applied when selecting the correct action for the two scroll directions.

**/Release Note**

This is my first draft to make input handlers more robust, make input action differentiable from actions that execute commands and map input action to a single module.

~~For now I left the old `input_handler` interface and used the dynamic pointer cast. I don't really know what is considered best practice here but I'm not really fond of it. 
I also considered just dropping `input_handler` and making all modules input handlers.
What do you think @Lomadriel?~~

EDIT: I integrated the input_handler into the module_interface class. Now only modules are input handlers.

I now need to also adapt every single module that is an input handler as well as create an exception for the `menu-*`, `i3wm-*`, and `bspwm-*` actions since they are part of the public API and can only be removed in the next major version.

Fixes #1172 

EDIT:

The built documentation for this branch can be viewed [here](https://polybar-patrick.readthedocs.io/en/issue-1172/user/actions.html)

TODO:

* [x] Add `exec-N` action to menu module so we know when to close the menu (With this we could easily implement #845). The `exec-N` action should produce a new action (not just run the shell command). This prevents code duplication for running shell commands and also allows menus to execute internal polybar actions (as they currently can as well). With this a user could trigger an infinite loop of actions by setting `menu-0-0-exec = #menu.exec.0-0`
* [x] Maybe use `#name.action` as the format. Looks a bit nicer
* [x] Document how actions work
* [x] Add list of all supported actions
* [x] Add migration documentation
* [x] More descriptive error message when legacy action names are used.
* [x] Some actions have data attached (e.g. `menu-open-1` or `bspwm-deskfocus0+4`). Maybe we could use a format like `#name.action[.data]` for this.
* [ ] Hooks for the IPC module could now also be triggered this way instead of using a separate category (Separate PR).


patrick96:
I have now made all modules input handlers, easier this way. But the controller still stores the input handlers, in case we want to introduce other kinds of input handlers in the future. I also removed the error for non unique module names and now deliver inputs to all matching input handlers.

patrick96:
Hey @Lomadriel, could I get some feedback on this to see if you think I'm going in the right direction. Or do you see any issues with this?

Lomadriel:
> Hey @Lomadriel, could I get some feedback on this to see if you think I'm going in the right direction. Or do you see any issues with this?

The implementation seems quite good to me. I'll check that ASAP.

For the idea itself, I really like your modification about the unique name.
I prefer to deliverer the output to all matching input handlers (so modules with same name) rather than just do nothing or forbid it.

I think you are going in the right direction :+1:

patrick96:
Quick status update: I have now switched over all the modules to support this new format. I have also renamed many of the actions using more inuitive and consistent names.

I have not yet tested any of this functionality. I also want to look into how much effort it would take to write some unit tests here.

Right now the menu module is a bit broken because before it intercepted all actions that polybar received (this includes shell commands) and compared them to its exec commands, if one of them matched, it closed the menu and tell the controller that it doesn't support that command.

I am now working on forwarding all the legacy action names to the proper place.

There are also actions that take data (for example which workspace to focus), I think we should make this data a part of the action specification.

include/modules/meta/factory.hpp
```diff
@@ -43,73 +42,71 @@
...
-#if not(ENABLE_I3 && ENABLE_MPD && ENABLE_NETWORK && ENABLE_ALSA && ENABLE_PULSEAUDIO && ENABLE_CURL && ENABLE_XKEYBOARD)
 #include "modules/unsupported.hpp"
-#endif
 
 POLYBAR_NS
 
 using namespace modules;
 
 namespace {
   module_interface* make_module(string&& name, const bar_settings& bar, string module_name, const logger& m_log) {
```
Lomadriel, line 52: (Not for this PR)
We should use a sort of global unordered_map and register each module in a map
This function is so ugly...

patrick96: I agree. I tried to make it a bit nicer in this PR by moving the type string into the module header but still.

But I'm not really sure how to achieve this, C++ doesn't have classes at runtime (only instances), so you can't do something like this: `clazz.newInstance()` as you can do in java.

Lomadriel: You can do something like:
```
struct BaseModuleFactory  {
    BaseModuleFactory() { /* do stuff */ }

    virtual ModuleInterface* build(...) = 0;
};

template <typename T>
struct ModuleFactory : BaseModuleFactory {
   ModuleInterface* build(...) { return new T(...); }
};

ModuleFactory<MyModule> var;
```

patrick96: Oh, and then we map module types to `ModuleFactory` instances?

Lomadriel: Yeap
```C++
// Factory header
void registerModuleFactory(std::string name, BaseModuleFactory& f);

// Factory source
static std::unordered_map<std::string, BaseModuleFactory&> getFactories(); // Returns factory map to avoid static order initialization fiasco
void registerModuleFactory(std::string name, BaseModuleFactory& f)  {
  getFactories()[name] = f;
}
```

patrick96: Neat! We should definitely look into that. But maybe after this PR is merged.

patrick96:
@Lomadriel I think this is ready for review now. I am going to work on documentation, but code-wise this should be done.

src/components/controller.cpp
```diff
@@ -386,36 +390,195 @@ void controller::process_eventqueue() {
...
 }
 
+/**
+ * Tries to match the given command to a legacy action string and sends the
+ * appropriate new action (and data) to the right module if possible.
+ *
+ * \returns true iff the given command matches a legacy action string and was
+ *          successfully forwarded to a module
+ */
+bool controller::try_forward_legacy_action(const string cmd) {
```
Lomadriel, line 400: Should take a const ref I think

src/components/controller.cpp
```diff
@@ -386,36 +390,195 @@ void controller::process_eventqueue() {
...
+   *
+   * The action will be delivered to the first module of that type so that it is consistent with existing behavior.
+   * If the module does not support the action or no matching module is found, the command is forwarded to the shell.
+   *
+   * TODO Remove when deprecated action names are removed
+   */
+// clang-format off
+#define A_MAP(old, module_name, event) {old, {string(module_name::TYPE), string(module_name::event)}}
+
+  const std::map<string, std::pair<string, const string>> legacy_actions{
```
Lomadriel, line 415: You could add a `static` to avoid recreation of this map at each call.

src/components/controller.cpp
```diff
@@ -386,36 +390,195 @@ void controller::process_eventqueue() {
...
+          // TODO make this message more descriptive and maybe link to some documentation
+          // TODO use route to string methods to print action name that should be used.
+          if (data.empty()) {
+            m_log.warn("The action '%s' is deprecated, use '#%s.%s' instead!", cmd, handler_name, action);
+          } else {
+            m_log.warn("The action '%s' is deprecated, use '#%s.%s.%s' instead!", cmd, handler_name, action, data);
+          }
+          m_log.info(
+              "Forwarding legacy action '%s' to module '%s' as '%s' with data '%s'", cmd, handler_name, action, data);
+          if (!handler_ptr->input(std::forward<string>(action), std::forward<string>(data))) {
```
Lomadriel, line 490: Maybe we should pass these parameters as const ref since we always need to copy them (line 471 and 473)

src/components/controller.cpp
```diff
@@ -386,36 +390,195 @@ void controller::process_eventqueue() {
...
 void controller::process_inputdata() {
-  if (!m_inputdata.empty()) {
-    string cmd = m_inputdata;
-    m_inputdata.clear();
+  if (m_inputdata.empty()) {
+    return;
+  }
+
+  const string cmd = m_inputdata;
+  m_inputdata.clear();
```
Lomadriel, line 518:
```suggestion
  const string cmd = std::move(m_inputdata);
  m_inputdata = string{}
```

src/components/controller.cpp
```diff
@@ -386,36 +390,195 @@ void controller::process_eventqueue() {
...
-      m_command->exec();
-      m_command.reset();
-      process_update(true);
-    } catch (const application_error& err) {
-      m_log.err("controller: Error while forwarding input to shell -> %s", err.what());
+  try {
+    // Run input as command if it's not an input for a module
+    m_log.info("Forwarding command to shell... (input: %s)", cmd);
+
+    if (m_command) {
```
Lomadriel, line 570: Useless code no ? Since we are always resetting the pointer (L.578)

patrick96:
Thanks for the quick review :)

src/components/controller.cpp
patrick96: Yes that works here, I'm not directly modifying cmd

src/components/controller.cpp
patrick96: Good idea

Lomadriel: Moreover, it can be an `unordered_map` since string are hash-able.

src/components/controller.cpp
patrick96: That code was already there. But you are right. I guess the original intent was to let the command run while polybar does other things, but since we always wait for completion, that isn't the case.

patrick96:
I'm still unsure if there is a good way to test all this. If we want to write tests to check if actions are routed to the proper handlers (same with legacy) actions, we will have to instantiate all the modules, which is virtually impossible in a test environment.

Lomadriel:
At the time I don't think this is easily feasible but in the future I want to separate each module and then we could create mock modules, mock renderer, etc... to test the routing.

patrick96:
Yes, that's my vision as well. I'd also like to get modules more lightweight and not have every module replicate all the logic for managing, formats, tokens, labels and other drawtypes.

patrick96:
I believe I am now finally finished. From what I can tell everything works correctly.

I think how actions are delivered to modules could still be cleaned up a bit. For example, each module could register itself as well as all its actions to an action router together with whether the action takes data and a reference to a member function that should be called when that action is triggered.

This would remove the logic from all modules that determines which action should be executed. But we should leave this for a future PR, the changes here are enough.

PR #1908, opened 2019-10-22
build: drop python2

### LINKED ISSUE #1892: python2 will be unmaintaned soon
Python 2.7 will not be maintained past 2020, see: https://pythonclock.org/
A lot of packages also drop `python2.7` support: https://python3statement.org/

Some critical bug can be found in `python2.7` which will not be fixed. Also, many users will remove `python2.7` from their machines.

User `python3.*` instead and `python3` libraries & dependencies.

Stop using python (it seems, this project doesn't have a lot of python code)

## Additional context
I can't remove `python2.7`from Gentoo right now only because `polybar` requires it:
```
!!! The ebuild selected to satisfy "x11-misc/polybar" has unmet requirements.
- x11-misc/polybar-3.2.1::gentoo USE="alsa curl ipc network -i3wm -mpd -pulseaudio" ABI_X86="(64)" PYTHON_TARGETS="-python2_7"

  The following REQUIRED_USE flag constraints are unsatisfied:
    python_targets_python2_7
```

patrick96:
Ref: https://github.com/polybar/xpp/pull/17
Fixes #1892

**Release notes:** All mentions of python2 can be removed from the wiki.
The `polybar` `PKGBUILD` can be updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ If you are using **Gentoo**, both release and git-master versions are available 
 A compiler with C++14 support ([clang-3.4+](http://llvm.org/releases/download.html), [gcc-5.1+](https://gcc.gnu.org/releases.html)), [cmake 3.1+](https://cmake.org/download/), [git](https://git-scm.com/downloads)
 - `cairo`
 - `libxcb`
-- `python2`
+- `python`
 - `xcb-proto`
 - `xcb-util-image`
 - `xcb-util-wm`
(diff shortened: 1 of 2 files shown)


patrick96:
We generally update the non-git build files after the release because they need the new sha256 value and also right now the non-git version (3.4.0) still requires python2. As for the rpm spec @jubalh said he would maintain that file, so I'm not going to update that (also I wouldn't have a system to test that on)

jubalh:
Thanks for the ping. I created https://github.com/polybar/polybar/pull/1928 to update the spec file.

PR #1919, opened 2019-10-27
feat: Add option to not enforce unique module names

patrick96:
Adds a new config key `allow-multiple` to every module section, by
default it's set to `false`.

If set to true, the controller will not emit an error when that module
appears multiple times.

There are still some cases where using the same module multiple times is
desirable. For example when you want to have different separators on the
left side than on the right side. adi1090x's polybar-themes does this
in several places.

Since in the future the module name is used to target actions, this will
lead to unspecified behavior when such a module receives an action.
I think when sending actions to modules, we will check allow_multiple()
and either not deliver the action at all or just emit an error.

Ref: https://github.com/adi1090x/polybar-themes

### DIFF
--- a/include/modules/meta/base.hpp
+++ b/include/modules/meta/base.hpp
@@ -39,7 +39,7 @@ namespace drawtypes {
   using animation_t = shared_ptr<animation>;
   class iconset;
   using iconset_t = shared_ptr<iconset>;
-}
+}  // namespace drawtypes
 
 class builder;
 class config;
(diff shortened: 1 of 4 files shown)


patrick96:
Maybe we should give this a better name. `allow-multiple` conveyes the purpose of the option but doesn't tell users about what it actually does only the side-effects (except when reading the docs). Something like `disable-actions` would describe what the module does. And we could then adapt the error message for duplicate modules to explicitly say that modules with actions cannot appear multiple times.

I don't want to dynamically determine which modules actually support actions and which do not because this would make adding actions to a module that hasn't had any before a breaking change.

patrick96:
Alright, I decided it may be better to allow to list the same module multiple times. But I think when actions are delivered, they should be delivered to *all* modules that can handle them. I think that is much more natural and it allows people to have multiple text modules with the same content (e.g. for separators) and not get any errors or have to add a new key.

PR #1921, opened 2019-10-28
build: Move non-macro variables into settings.cpp

patrick96:
Since APP_VERSION is different for every commit and almost all file
include settings.hpp, the whole project has to be rebuilt for every
commit. With this, hopefully, this can be greatly reduced and only
changed files need to be rebuilt. This will also help ccache

### DIFF
--- a/include/components/ipc.hpp
+++ b/include/components/ipc.hpp
@@ -13,18 +13,9 @@ class signal_emitter;
 /**
  * Message types
  */
-struct ipc_command {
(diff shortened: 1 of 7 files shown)


patrick96:
It's only the default for functions, and I can remove it there if you want. But for variables it changes the semantics, and because they are `const`, you get an `uninitialized const` error when you remove it.

Looking at this again, I think I will move everything that doesn't has to be used in a macro into a variable.

PR #1926, opened 2019-10-30
Update PKGBUILD for 3.4.1

patrick96:


contrib/polybar.aur/PKGBUILD
```diff
@@ -1,8 +1,8 @@
 # Maintainer: Michael Carlberg <c@rlberg.se>
```
Lomadriel: Should we update that or add a new contributor?

patrick96: I updated it. Since I am the only one that updates that package, I should also be the contact person.

PR #1936, opened 2019-11-22
Renderer alignment fixes

### LINKED ISSUE #591: System tray sometimes overlapping the rest of the bar

### LINKED ISSUE #1268: Bar itself shifting to right
Hello, my bar sometimes shifts all contents to right, that clock isn't visible. It's very irritating. 
This is my fault or internal bug?

image

Polybar config: [in my repo](https://github.com/l7ssha/dotfiles/blob/master/config/polybar/config)
Polybar version: `polybar 3.1.0-git`
OS: Arch Linux, 4.16.9-1-ARCH, Openbox WM

### LINKED ISSUE #1838: Blocks get under systray even when there is free space in the bar

When one of the blocks (namely `modules-right`) grows too  big, it starts to "get under" the system tray, hiding the rightmost module. However, there is plenty of space between the leftmost module and the `modules-left` block. I don't know if this is due to some maximum sizes or padding being forced, but I think it would be great to be able to configure this.

Correctly displayed bar: 
DeepinScreenshot_select-area_20190724122213

Overlapping the tray:
DeepinScreenshot_select-area_20190724122241

As it can be seen, there is still plenty of space between `modules-right` and `modules-left`. Why is it not being used?

Current behavior looks bad and wastes space.

I don't know exactly the root cause of this, but the user should be able to define its own margins, paddings, or minimum widths to avoid this effect. Or maybe this are just hardcoded values set too high and can be fixed without user configuration.

### LINKED ISSUE #1903: wrong alignments with scripts and tail output
A bug in polybar annoys me for a long time. But I could never reproduce it. But I think I'm on the track now I guess.

The elements on the bar go out of the screen from time to time. The reason seems to be scripts that change their width. I've seen this problem with tail scripts, but I can't say if it also happens with simple scripts. But I never noticed that at simple scripts. Also it seems that you need more than one tail scripts. It happens at a group of right aligned modules. Its hard to reproduce.

Apparently the alignment is calculated too rarely or something. I wrote a test script to reproduce it.

**Expected behavior:**

A wondeful `polybar`.

**Actual behavior:**

Everything is ugly and messed up.

**To Reproduce**
```sh
#!/bin/sh

while [ true ]; do
    echo "X X X X X"
    sleep 3

    echo "X"
    sleep 3
done
```
Ugly:
2019-10-17-21:54:00-screenshot
Wonderful
2019-10-17-21:54:03-screenshot

```
polybar 3.4.0-20-g4ab251f3

Build type: Release
Compiler: /usr/bin/c++
```

patrick96:
I finally found the issue in https://github.com/polybar/polybar/issues/591!!

Turns out it wasn't the tray code but the code that calculates the block positions. It didn't work when the center block was empty and the right block grew too large (with `fixed-center = true`).

There were also some bugs with `fixed-center` which basically made `fixed-center = true` not work reliably if the right block was too large. I basically rewrote the entire logic there.

The definition of what `fixed-center` exactly does wasn't really clear as well and since the code was full of bugs, I found it hard to figure out what the intention was. Here is what I came up with:

* `fixed-center = true`: The center block stays at the center of the bar
whenever possible. It can be pushed to the left if the right block takes
too much space and to the right if the left block takes too much space
* `fixed-center = false`: The center block will be in the middle between
the left and right block whenever possible. If there is not enough space
between those two, the center block will be directly to the right of the
left block and pushes out the right block

This should preserve the same behavior as it has now (for the cases where there were no bugs).

Finally there were some bugs with how the falloff gradient was drawn.

@x70b1 Can you check that this also fixes #1903

Fixes #591 
Fixes #1903
Fixes #1838
Fixes #1268

### DIFF
--- a/include/cairo/context.hpp
+++ b/include/cairo/context.hpp
@@ -108,10 +108,9 @@ namespace cairo {
     }
 
     context& operator<<(const linear_gradient& l) {
-      if (l.steps.size() >= 2) {
+      auto stops = l.steps.size();
+      if (stops >= 2) {
         auto pattern = cairo_pattern_create_linear(l.x1, l.y1, l.x2, l.y2);
-        *this << pattern;
-        auto stops = l.steps.size();
         auto step = 1.0 / (stops - 1);
         auto offset = 0.0;
         for (auto&& color : l.steps) {
@@ -124,6 +123,7 @@ namespace cairo {
           // clang-format on
           offset += step;
         }
+        *this << pattern;
         cairo_pattern_destroy(pattern);
       }
       return *this;
--- a/src/components/renderer.cpp
+++ b/src/components/renderer.cpp
@@ -341,7 +341,7 @@ void renderer::flush(alignment a) {
   double w = static_cast<int>(block_w(a) + 0.5);
   double h = static_cast<int>(block_h(a) + 0.5);
   double xw = x + w;
-  bool fits{xw <= m_rect.x + m_rect.width};
+  bool fits{xw <= m_rect.width};
 
   m_log.trace("renderer: flush(%i geom=%gx%g+%g+%g, falloff=%i)", static_cast<int>(a), w, h, x, y, !fits);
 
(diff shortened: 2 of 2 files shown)


Lomadriel:
Nice catch. I'll try to review that PR ASAP.

patrick96:
So my comments and findings on this are all over place in the linked issues, so I'm going to do a little post-mortem here.

The issue was that, if the center module list was empty, its position would still be used when trying to calculate whether the modules on the right have enough space. In most cases this wasn't apparent because even with `fixed-center = true`, the center block would be shifted out of the way of the right block grew too large. But there were some edge cases:
* If there was a tray or a border on the right, that calculation that was supposed to shift the center block to the left if the right block was too large was completely wrong because it only considered the size of the right block and not the tray or border
* Even without tray or border the calculation for shifting the center block didn't consider the `BLOCK_GAP`, so if the right block would end in a 20px wide area directly to the right of the bar center, the right block would still shift to the right.

PR #1938, opened 2019-11-25
Add: Multiple-time and recursive dereferencing of config parser.

SleepyBag:
In-string references will be supported now then.
Specifically, config parser will dereference all references shown in a field. Besides, if there are nested references, it will dereference them recursively until there is no "${}" left.

Though this cause some uncompatibility. That is, if someone wrote "${}" in a 'exec' script, config parser will try to dereference it. One has to add another '$' before "${}" to get avoid of this.

This commit is experiment with the following config file:

[A]
a1 = a1
a2 = a2
a3 = a3
a4 = a4

[B]
b1 = ${A.a1}
b2 = b2${A.a2}
b3 = ${A.a3}b3
b4 = b4${A.a4}b4

[C]
c1 = ${B.b1}
c2 = c2${B.b2}
c3 = ${B.b3}c3
c4 = c4${B.b4}c4

[bar/bar]
a1 = ${A.a1}
a2 = ${A.a2}
a3 = ${A.a3}
a4 = ${A.a4}

b1 = ${B.b1}
b2 = ${B.b2}
b3 = ${B.b3}
b4 = ${B.b4}

c1 = ${C.c1}
c2 = ${C.c2}
c3 = ${C.c3}
c4 = ${C.c4}

d1 = d1${C.c1}${C.c1}
d2 = ${C.c2}${C.c2}d2
d3 = d3${C.c3}d3${C.c3}
d4 = ${C.c4}${C.c4}d4

e1 = ${A.${A.${A.a1}}}
e2 = e2${A.${A.a2}}

f1 = $$f1
f2 = $$${C.c2}
f3 = $${C.c3}

### DIFF
--- a/include/components/config.hpp
+++ b/include/components/config.hpp
@@ -214,23 +214,72 @@ class config {
    */
   template <typename T>
   T dereference(const string& section, const string& key, const string& var, const T& fallback) const {
-    if (var.substr(0, 2) != "${" || var.substr(var.length() - 1) != "}") {
-      return fallback;
+    // Find the first reference
+    string::size_type reference_begin = var.find("$");
+    while (reference_begin != string::npos && reference_begin != var.length() - 1 && var[reference_begin + 1] != '{') {
+      if (var[reference_begin + 1] == '$') {
+        ++reference_begin;
+      }
+      reference_begin = var.find("$", reference_begin + 1);
     }
-
-    auto path = var.substr(2, var.length() - 3);
-    size_t pos;
-
-    if (path.compare(0, 4, "env:") == 0) {
-      return dereference_env<T>(path.substr(4));
-    } else if (path.compare(0, 5, "xrdb:") == 0) {
-      return dereference_xrdb<T>(path.substr(5));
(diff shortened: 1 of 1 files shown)


patrick96:
Oh wow, this is awesome! This would make the config a lot more expressive.

Thanks a lot!
Will take a look at this later.

patrick96:
Sorry for never getting to this.

I had some reservations about this and thought I would address them once I got around to working on the config parser again, but that never happened.

The goal with the config parser is to not resolve variables when looking up config values, but actually when parsing the config file. This would also allow us to fix the infinite recursion issue (#783). The nested references would make such an endeavor quite difficult, and I don't think polybar needs nested references.

We also just recently decided to make the backslash the escape character in config values.

As for next steps, I of course don't expect you to still work on this, it has been quite some time. But if you still want to, let me know. Doing references inside the config parser is not feasible right now, because it does require quite a few architectural changes. But I think we can just keep recursively resolving references for now until we get that functionality into the config parser.

PR #1939, opened 2019-11-25
doc: Add Gitter room

patrick96:
The Gitter room https://gitter.im/polybar/polybar should serve a similar
purpose as the IRC room with the benefit that messages persist and can
be received when offline.

From now on gitter should be the place where the polybar community
lives. Reddit is great for asking longer questions in the style of a
forum but not so great for quick back and forth conversations. IRC is
better in that aspect but has the serious downside that you have to stay
connected to get an answer and messages are not logged (only by your
client).

Both reddit and IRC are here to stay and I will still check them
regularly, but we should encourage people to join gitter.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,3 +98,10 @@ notifications:
     use_notice: true
     on_success: change
     on_failure: change
+  webhooks:
+    urls:
+      # For the https://gitter.im/polybar/polybar gitter room
+      - https://webhooks.gitter.im/e/10bdbe25961312646ace
+    on_success: change
+    on_failure: always
+    on_start: never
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -98,3 +98,10 @@ notifications:
     use_notice: true
     on_success: change
     on_failure: change
+  webhooks:
+    urls:
+      # For the https://gitter.im/polybar/polybar gitter room
+      - https://webhooks.gitter.im/e/10bdbe25961312646ace
+    on_success: change
+    on_failure: always
+    on_start: neve
```
Lomadriel: Is it really `neve` ?

patrick96: It most definitely is not ^^

patrick96: Fixed

PR #1940, opened 2019-11-25
Right-click on pulseaudio launches control panel

jtdubs:
This is a tiny commit that scratches an itch for me but I suspect others would find useful.  It creates an easy way to get into the audio control panel.

From the commit message:
Adds a new EVENT_OPEN_CONTROL_PANEL event to the pulseaudio module.

The event is bound to right-click, and it's handler invokes the
"control-panel" app as defined in the config file (default is
"pavucontrol")

If the specified control panel command does not end in "&", one is
appended so that the system() all will not block.

### DIFF
--- a/include/modules/pulseaudio.hpp
+++ b/include/modules/pulseaudio.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "settings.hpp"
+#include "utils/command.hpp"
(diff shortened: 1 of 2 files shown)


jtdubs:
Opened pre-maturely.  Will re-open off a feature branch.  And switch to using command_util.

PR #1941, opened 2019-11-26
Feature pulseaudio controlplanel

jtdubs:
This PR adds a config option to specify a control panel for pulseaudio.  The default is pavucontrol.

It also listens for right-click events and uses command_util to launch the specified control panel.

### DIFF
--- a/src/modules/pulseaudio.cpp
+++ b/src/modules/pulseaudio.cpp
@@ -106,6 +106,17 @@ namespace modules {
     string output{module::get_output()};
 
     if (m_handle_events) {
+        auto click_middle = m_conf.get(name(), "click-middle", ""s);
+        auto click_right = m_conf.get(name(), "click-right", ""s);
+
+        if (!click_middle.empty()) {
+            m_builder->cmd(mousebtn::MIDDLE, click_middle);
+        }
+
+        if (!click_right.empty()) {
+            m_builder->cmd(mousebtn::RIGHT, click_right);
+        }
+
       m_builder->cmd(mousebtn::LEFT, EVENT_TOGGLE_MUTE);
       m_builder->cmd(mousebtn::SCROLL_UP, EVENT_VOLUME_UP);
       m_builder->cmd(mousebtn::SCROLL_DOWN, EVENT_VOLUME_DOWN);
--- a/src/modules/pulseaudio.cpp
+++ b/src/modules/pulseaudio.cpp
@@ -106,16 +106,16 @@ namespace modules {
     string output{module::get_output()};
 
     if (m_handle_events) {
-        auto click_middle = m_conf.get(name(), "click-middle", ""s);
-        auto click_right = m_conf.get(name(), "click-right", ""s);
+      auto click_middle = m_conf.get(name(), "click-middle", ""s);
+      auto click_right = m_conf.get(name(), "click-right", ""s);
 
-        if (!click_middle.empty()) {
-            m_builder->cmd(mousebtn::MIDDLE, click_middle);
-        }
+      if (!click_middle.empty()) {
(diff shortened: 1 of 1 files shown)


patrick96:
@jtdubs Thanks for contributing! Unfortunately we will not merge this in its current state, because it doesn't fit with the direction we want click handlers to go. Read my write up in https://github.com/polybar/polybar/issues/1524. If we merged it like this, we would just have to deprecate the `control-panel` key later on.

All hope is not lost though :)
The idea itself is good but I as I said, the approach doesn't fit. I would suggest renaming the config option `click-right` with default value `pavucontrol &`. So you shouldn't create a new event but rather use `m_builder->cmd(mousebtn::RIGHT, ...)` to directly generate action blocks, have a look how the text module does it.

Also don't forcibly add `&` to the end of the command the user specified, that is a bug in polybar that executed commands block polybar.

jtdubs:
@patrick96, thank you for the feedback.  I made this much harder for myself than I needed to given the capabilities of m_builder.  :)  I re-wrote it and squashed the commits.  It's now just a click-right handler bound to mousebtn::RIGHT using m_builder.  I also removed the default value to align with how the text module worked.  It's easy enough to specify click-right=pavucontrol in the config file.

Also, I just noticed the branch name contains a type-o.  "planel".  I don't know how many times I've stared at that and didn't notice.  I can certainly rename it, but it'd require a new PR.

Let me know if/how you want to proceed.  And, thanks again for the feedback.  It's an awesome toolbar, and an easy-to-read code base.

src/modules/pulseaudio.cpp
```diff
@@ -18,6 +18,12 @@ namespace modules {
     // Load configuration values
     m_interval = m_conf.get(name(), "interval", m_interval);
 
+    auto click_right = m_conf.get(name(), "click-right", ""s);
+
+    if (!click_right.empty()) {
+      m_builder->cmd(mousebtn::RIGHT, click_right);
```
patrick96: This won't work. You need to call `m_builder->cmd` in the `get_output` method. The text module has a good example for that.

jtdubs:
@patrick96, interesting.  That explains some inconsistent behavior I was seeing on further testing.  It worked once, but after that produced an error indicating it couldn't find an input area.   The connection between event listeners and get_output() is not obvious to me.

The code is moved, and a middle-click handler is added for consistency with the text module.

patrick96:
> That explains some inconsistent behavior I was seeing on further testing

It probably stopped working after the module updated for the first time, right?

That's because the builder gets reset whenever you update the module contents.

>  The connection between event listeners and get_output() is not obvious to me.

It's not really a direct connection. What you do is you add `%{A}` formatting tags around the module in `get_output`. The renderer then remembers which blocks of text are surrounded by such tags. Whenever the bar is clicked, the bar tries to match the position of the click with the blocks the renderer saved. If a block is found that block contains the command you gave in the action tag and that command is then forwarded to the controller for execution.

patrick96:
Congrats on your first contribution to polybar :rocket:

hpfr:
Hi, the wiki says the click-right and click-middle features are unreleased. Maybe this is out-of-date, but it doesn't seem to work on my machine running 3.4.3

The left click works, so `m_builder->cmd(mousebtn::LEFT, EVENT_TOGGLE_MUTE);` must work. Maybe the variables that read from the config somehow come back empty? My `polybar.conf` includes `click-right = pavucontrol &` under `[module/pulseaudio]`. I'm running it with a `systemd` service and nothing is logged on both right and left click. Any ideas?

patrick96:
The wiki is right. click-right and click-middle are unreleased. That's why they don't work for you.

Issue #809, opened 2017-10-18
Padding for bar elements


Chais:
I found no prior issue and don't see a reason why `bar-NAME-padding` shouldn't be a thing.

patrick96:
What would `bar-NAME-padding` do? Is `NAME` the name of the bar?
There is already a `padding[-{left,right}]` option for the bar, maybe that's what you're looking for

Chais:
Yes, `NAME` is the name of the bar. As in `bar-volume-padding`.
Neither `bar-volume-padding-left = 1` nor `padding-left = 1` have the desired effect. The bar is directly adjacent to the prefix and suffix.

patrick96:
I think we are talking about different things. When I asked, if `<NAME>` was the name of the bar, I meant if you have a section in your config starting with the following
```dosini
[bar/<NAME>]
```

What I think you are talking about is a module named `volume` and you want the contents of the module further away from the prefix and suffix.
Could you please post your config on here in code tags.
You can also try to set `format-volume-padding`

Chais:
I'm referring to [progress bars](https://github.com/jaagr/polybar/wiki/Formatting#progressbar).
I can't use `format-volume padding` because it pads in the wrong place.
`format-volume-padding`:  ![format-padding](https://user-images.githubusercontent.com/1783098/31788116-07197e0a-b50e-11e7-8c11-24544913d11f.png)  
What I need: ![bar-volume-padding](https://user-images.githubusercontent.com/1783098/31788247-79f6eb92-b50e-11e7-8a63-d592a7627a48.png)  
This was done with `format-volume-prefix = " "`

patrick96:
Ah I see what you mean. Sorry for the confusion.
It's probably a good idea to have all other drawtypes (ramp, animation etc) support padding as well. Right now this is not really high priority because you can easily achieve the same thing by adding a space to the prefix or adding a space to the format:
```dosini
format-volume = " <bar-volume>"
```

We're all really busy right now, so if you or anyone else that stumbles accross this wants to implement this feature, please feel free to do so.

Chais:
Ah OK. For some reason that possibility didn't occur to me. I'll do that in the meantime. Thanks.

BVollmerhaus:
> It's probably a good idea to have all other drawtypes (ramp, animation etc) support padding as well. Right now this is not really high priority because you can easily achieve the same thing by adding a space to the prefix or adding a space to the format

[...]

PR #1947, opened 2019-12-01
Add padding support for progressbar drawtype

jtdubs:
This is a basic implementation of padding for progressbar, based loosely on the implementation for label.

I also looked at animation and ramp, but it appears they should get padding support automatically due to the use of load_optional_label() and copy_undefined() in their load methods.

This was targeting at resolving issue #809 .

### DIFF
--- a/include/drawtypes/progressbar.hpp
+++ b/include/drawtypes/progressbar.hpp
@@ -18,6 +18,7 @@ namespace drawtypes {
     void set_indicator(label_t&& indicator);
     void set_gradient(bool mode);
     void set_colors(vector<string>&& colors);
+    void set_padding(side_values padding);
 
     string output(float percentage);
 
@@ -31,6 +32,7 @@ namespace drawtypes {
     unsigned int m_width;
     unsigned int m_colorstep = 1;
     bool m_gradient = false;
+    side_values m_padding{0U, 0U};
 
     label_t m_fill;
     label_t m_empty;
--- a/src/drawtypes/progressbar.cpp
+++ b/src/drawtypes/progressbar.cpp
@@ -31,6 +31,10 @@ namespace drawtypes {
     m_gradient = mode;
   }
 
+  void progressbar::set_padding(side_values padding) {
+    m_padding = padding;
+  }
+
   void progressbar::set_colors(vector<string>&& colors) {
     m_colors = forward<decltype(colors)>(colors);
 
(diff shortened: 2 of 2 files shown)


patrick96:
We could probably make `bar-NAME-format` a label itself so we don't have to duplicate all the code to load and apply the padding right?

jtdubs:
Good point.  I'll take a crack at that later tonight.

jtdubs:
Yeah, that was a good clean-up.  progressbar::m_format has been replaced with a label.  This cleaned up the pre-existing output() code by letting it re-use label::replace_token rather than using string_util::replace_all directly.   Margins and padding work for free.  You can reverse the direction of a bar by setting format="%empty%%indicator%%fill%" in the config, etc.

jtdubs:
Looking back at your comment, I made `bar-NAME` a label, rather than `bar-NAME-format`.  This seemed cleaner as it also makes `bar-NAME-foreground`, etc. all work as I would expect.  I can push it down to the `-forrmat` sub-element, but using `bar-NAME-format-foreground` is less intuitive.  What would you prefer?

patrick96:
Yeah, I didn't like the `string_util::replace_all` in there either. If we are literally replacing tokens we can use the functions we wrote for that ;)

This now no longer uses `bar-NAME-format` for the actual format, right? We should deprecate it then and tell the user to switch.

Also `bar_label->get()` will give you the label content after it has been truncated (minlen, maxlen). You should try to get the raw label contents.

jtdubs:
@patrick96, that spiraled a bit.  Handling deprecation was non-trivial as load_label is where all the config access was occurring.  Additionally, label previously had no method for allowing access to the raw format string.

I ended up extending the load_label and conf.deprecated machinery to handle this.  There's now a variation of conf.deprecated to handle required parameters (before, it only handled optional ones).  load_label now handles there being a deprecated_name, uses the new deprecation machinery, and the right deprecation messages are produced.

It only handles the actual format-string being deprecated in the label, and not all the other possible label fields (padding, margin, etc.).  This is clearly something that can be improved, but it is all that is needed to enable this change to progressbar as there was no `name-BAR-format-LABELCONF` before.

Honestly, I'd prefer to rework the deprecation machinery entirely in a future PR.  It would seem cleaner to do a deprecation pass through the config, issuing warnings, and effectively doing a new in-memory-only conf::set to place the deprecated fields in their new homes, then the users of conf can just assume everything is where it's supposed to be.  This would allow multi-hop (chained) deprecation, and avoid the need for different deprecate methods for different types as well as the optional/required variants.

Something like:
```
// move older_key (if found) to old_key and issue warning if found
conf.deprecate(section, older_key, old_key);
// ditto for old_key to new_key
conf.deprecate(section, old_key, new_key);

... code continues and uses conf.get(section, new_key) only ...
```

Let me know what you want to do regarding this specific progressbar, padding feature.  I don't want to spiral this too far.

PR #1953, opened 2019-12-12
Fix controller.cpp compilation on GCC 6.3

Xyene:
This fixes the below error:

     /code/polybar/src/components/controller.cpp:110:117:   required from here
     /code/polybar/src/components/controller.cpp:100:60: error: passing ‘const volatile std::shared_ptr<polybar::modules::module_interface>’ as ‘this’ argument discards qualifiers [-fpermissive]
          auto equal_predicate = [](auto& m1, auto& m2) { return m1->name() == m2->name(); };

### DIFF
--- a/src/components/controller.cpp
+++ b/src/components/controller.cpp
@@ -97,7 +97,7 @@ controller::controller(connection& conn, signal_emitter& emitter, const logger& 
   auto dup_it = m_modules.cbegin();
 
   do {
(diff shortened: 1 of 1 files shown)


Lomadriel:
> Weird that we don't get that error in newer gcc versions and even weirder that for some reason the modules are passed as volatile.

If I'm not wrong that's a bug in GCC 6.3

patrick96:
Interesting, I was able to reproduce the issue with the latest gcc6, 6.5 though as well. So probably better to fix it on our side that gcc6 users can still compile.

PR #1967, opened 2019-12-24
Add missing <stdexcept> header

jeremyong:
This is dependent on a PR to xpp that does the same. Newer compilers
(GCC10 in particular) are stricter about which headers provide the
std exception types.


patrick96:
Thanks a ton. The fixes should be released soon with polybar 3.4.2

Issue #1971, opened 2019-12-27
Information for Polybar Packager


patrick96:
If you are maintaining a polybar package in some distro, please subscribe to this thread.

We will post a comment here whenever there is a new release and we will also include anything that has changed for building and packaging the new version.

This thread is not intended for discussions of new releases, please do not comment here, we don't want to spam people that have subscribed to this thread. If you have any issues or questions about packaging the new release, please open a new issue (or ask on [Gitter](https://gitter.im/polybar/polybar)).

patrick96:
We have released [polybar 3.4.2](https://github.com/polybar/polybar/releases/tag/3.4.2) :rocket: 

There are no changes to the build workflow.

Build issues on the new gcc10 have been fixed.

patrick96:
We have released [polybar 3.4.3](https://github.com/polybar/polybar/releases/tag/3.4.3) :rocket: 

This release only contains a fix for a gcc10 build issue (#2098).

patrick96:
We have just released [polybar 3.5.0](https://github.com/polybar/polybar/releases/tag/3.5.0) :tada: 

Packaging-wise not much has changed. You can now specify the python executable to use during compilation by passing `-D:PYTHON_EXECUTABLE=/path/to/python` to `cmake`, though for packaging this shouldn't really be necessary, it's more for systems that have conda or pyenv.

The bigger change is probably that the i3ipcpp submodule now doesn't compile jsoncpp itself, it just always uses whatever jsoncpp is available (we have not yet had problems with jsoncpp < 1.7.7).

patrick96:
The first patch release for 3.5 is out now: [polybar 3.5.1](https://github.com/polybar/polybar/releases/tag/3.5.1) :rocket:

From now on, the release archive is a `.tar.gz` file instead of just a tar file and it contains a folder named `polybar-<version>` instead of just `polybar`.

We also fixed some build issues in older clang versions (3.4) and some in more modern clang.

patrick96:
Polybar [3.5.2](https://github.com/polybar/polybar/releases/tag/3.5.2) is out now.

Just a quick bugfix, no build changes.

patrick96:
Just released version [3.5.3](https://github.com/polybar/polybar/releases/tag/3.5.3).

This fixes the issue where you needed to have git installed to build the documentation even if you were building from a release archive.

patrick96:
[...]

PR #1975, opened 2019-12-27
doc: Write page for package maintainers.

patrick96:
I have also created issue #1971 to inform package maintainers about changes to how polybar is built.

### DIFF
--- a/doc/dev/packaging.rst
+++ b/doc/dev/packaging.rst
@@ -1,0 +1,88 @@
+Packaging Polybar
+=================
+
+Do you want to package polybar for a distro? Great! Read this page to get
+started.
+
+First Steps
+-----------
+
+Before you get started, have a look at the `Packaging Label
+<https://github.com/polybar/polybar/issues?q=label%3APackaging>`_ on our GitHub
+repo and `Repology <https://repology.org/project/polybar/versions>`_ to see if
+polybar is already packaged for that distro or if there are efforts to do so.
+
+Even if a package already exists, it might still make sense for you to package
+polybar in some cases. Some of these cases are:
+
+- The existing package is out-of-date and the packager is no longer able/willing
+  to continue maintaining the package (or they are simply not reachable
+  anymore).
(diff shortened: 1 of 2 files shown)


doc/dev/packaging.rst
```diff
@@ -0,0 +1,88 @@
...
+Do you want to package polybar for a distro? Great! Read this page to get
+started.
+
+First Steps
+-----------
+
+Before you get started, have a look at the `Packaging Label
+<https://github.com/polybar/polybar/issues?q=label%3APackaging>`_ on our GitHub
+repo and `Repology <https://repology.org/project/polybar/versions>`_ to see if
+polybar is already packaged for that distro or if there are efforts to do so.
```
Lomadriel:
```suggestion
polybar is already packaged for that distro or if they are efforts to do so.
```

doc/index.rst
```diff
@@ -18,6 +18,12 @@ Welcome to the official polybar documentation.
 
    man/polybar.1
 
+.. toctree::
+   :maxdepth: 1
+   :caption: For Developers:
```
Lomadriel: Maybe
```suggestion
   :caption: For packagers:
```

doc/dev/packaging.rst
patrick96: I'm pretty sure my sentence is correct.

doc/index.rst
patrick96: Yeah, I wasn't sure how to categorize this. It isn't really user documentation but also not quite developer documentation. And this will probably be the only page about packaging polybar, so it shouldn't have its own category. Maybe we can name it contributor documentation.

doc/dev/packaging.rst
Lomadriel: Yeap indeed my bad.

doc/index.rst
Lomadriel: > Maybe we can name it contributor documentation.

Make sense.

PR #1976, opened 2019-12-28
Add Fedora installation instructions

tim77:
Packaged and available in official repos. Now [on QA](https://bodhi.fedoraproject.org/updates/FEDORA-2019-f0e9cd0ce7). Please hold a little bit this PR and don't merge until status in Bodhi will be changed from Pending/Testing to **Stable**.
DeepinScreenshot_select-area_20191228231133

TEAM EDIT:
Closes #1639 
Closes #1645

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,8 @@ If you are using **FreeBSD**, [polybar](https://svnweb.freebsd.org/ports/head/x1
 
 If you are using **Gentoo**, both release and git-master versions are available in the [main](https://packages.gentoo.org/packages/x11-misc/polybar) repository.
 
+If you are using **Fedora**, you can install [polybar](https://src.fedoraproject.org/rpms/polybar) using `sudo dnf install polybar`.
+
 ### Dependencies
 
 A compiler with C++14 support ([clang-3.4+](http://llvm.org/releases/download.html), [gcc-5.1+](https://gcc.gnu.org/releases.html)), [cmake 3.1+](https://cmake.org/download/), [git](https://git-scm.com/downloads)


patrick96:
Nice, we're getting an official fedora package :smiley: 

I wasn't able to find the spec file for 3.4.2, so I just looked at the one for 3.4.1 listed in [BZ#1785874](https://bugzilla.redhat.com/show_bug.cgi?id=1785874): [`polybar.spec`](https://copr-be.cloud.fedoraproject.org/results/atim/polybar/fedora-31-x86_64/01134122-polybar/polybar.spec)

I am not really familiar with the syntax and semantics of spec files but I noticed that the spec file only uses `BuildRequires`. Does that mean that this package has no runtime dependencies and only buildtime dependencies? Because only cmake, gcc, python, sphinx (and technically i3) can be build-only dependencies.

Instead of `wireless-tools-devel` we recommend you use `libnl3-devel`. The wireless-tools code is deprecated and only here for backwards compatibility.

tim77:
> I wasn't able to find the spec file for 3.4.2, so I just looked at the one for 3.4.1 listed in BZ#1785874: polybar.spec

https://src.fedoraproject.org/rpms/polybar/blob/master/f/polybar.spec

> Does that mean that this package has no runtime dependencies and only buildtime dependencies? Because only cmake, gcc, python, sphinx (and technically i3) can be build-only dependencies.

Yes, this just build-only deps. You can check Requires by:

`rpm -qp --requires polybar-3.4.2-2.fc31.x86_64.rpm`
 
> Instead of wireless-tools-devel we recommend you use libnl3-devel. The wireless-tools code is deprecated and only here for backwards compatibility.

[Fixed](https://src.fedoraproject.org/rpms/polybar/blob/master/f/polybar.spec#_38). ✅

patrick96:
> Yes, this just build-only deps. You can check Requires by:
rpm -qp --requires polybar-3.4.2-2.fc31.x86_64.rpm

Interesting. So rpm calculates runtime dependencies dynamically from the libraries linked? That's really cool, I generally like how rpm packages take quite a rigorous approach to dependencies, I'm kind of missing that in Arch.

Anyway, I haven't found any other issues with the package. Thanks a lot for packaging polybar :+1: 

Also make sure to subscribe to #1971 to get notified about build changes in new releases.

Could you notify us, once this becomes stable and is ready to merge?

tim77:
> Also make sure to subscribe to #1971 to get notified about build changes in new releases.

Sure!

> Could you notify us, once this becomes stable and is ready to merge?

There were the holidays + Bodhi was broken for some time but now Polybar already 5 days [in Stable repo](https://bodhi.fedoraproject.org/updates/FEDORA-2019-a0d98da21e). 🎉

patrick96:
Great. Thanks a lot! Great to have polybar in the official repos 🎉🎉

PR #1977, opened 2019-12-29
build: allow building packages with CPack

hmenke:
I have just successfully compiled polybar for Debian Buster and packaged it up into an installable deb package that I can throw into my own local repo.  This has several advantages:  I can build the package in a container and deploy to my normal system without having to install all of the build dependencies on it.  When installing the package from my local repo, APT will do the dependency resolution for me and only install the runtime dependencies.

Right now, the packaging option in CMake is turned off my default.  Building the package works via the new `package` target.

Open questions/future work:
- [ ] Possibly add the package building option to `build.sh`
- [ ] Maybe automatically build packages on Travis CI and deploy via GitHub releases.
- [ ] Perhaps users do not want to build both DEB and RPM (for which they'd have to have `dpkg` and `rpm` installed) so maybe split the CMake option into two.

Here is what building the package looks like in the terminal:
```console
root@debian:~/polybar/build# make package
[  5%] Built target xpp
[  9%] Built target jsoncpp_lib_static
[ 11%] Built target i3ipc++
[ 92%] Built target poly
[ 94%] Built target polybar
[100%] Built target polybar-msg
Run CPack packaging tool...
CPack: Create package using TGZ
CPack: Install projects
CPack: - Run preinstall target for: polybar
CPack: - Install project: polybar
CPack: Create package
CPack: - package: /root/polybar/build/polybar-3.4.0-Linux.tar.gz generated.
CPack: Create package using RPM
CPack: Install projects
CPack: - Run preinstall target for: polybar
CPack: - Install project: polybar
CPack: Create package
-- CPackRPM:Debug: Using CPACK_RPM_ROOTDIR=/root/polybar/build/_CPack_Packages/Linux/RPM
CPackRPM: Will use GENERATED spec file: /root/polybar/build/_CPack_Packages/Linux/RPM/SPECS/polybar.spec
CPack: - package: /root/polybar/build/polybar-3.4.0-Linux.rpm generated.
CPack: Create package using DEB
CPack: Install projects
CPack: - Run preinstall target for: polybar
CPack: - Install project: polybar
CPack: Create package
CPackDeb: - Generating dependency list
CPack: - package: /root/polybar/build/polybar-3.4.0-Linux.deb generated.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -165,3 +165,25 @@ install(FILES config
   COMPONENT config)
 
 # }}}
+# Build Linux packages {{{
+if(CPACK_GENERATOR)
+  # General
+  set(CPACK_PACKAGE_VERSION "${version_txt}")
+  set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "A fast and easy-to-use tool for creating status bars.")
+
+  # DEB
+  set(CPACK_DEBIAN_PACKAGE_ARCHITECTURE "amd64")
+  set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS ON)
+  set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Polybar developers")
+  set(CPACK_DEBIAN_PACKAGE_SECTION "x11")
+  set(CPACK_DEBIAN_PACKAGE_PRIORITY "optional")
+
+  # RPM
+  set(CPACK_RPM_PACKAGE_ARCHITECTURE "x86_64")
+  set(CPACK_RPM_PACKAGE_AUTOREQPROV ON)
+  set(CPACK_RPM_PACKAGE_VENDOR "Polybar developers")
+  set(CPACK_RPM_PACKAGE_GROUP "System/GUI/Other")
+  set(CPACK_RPM_PACKAGE_PRIORITY "optional")
+
+  include(CPack)
+endif()


patrick96:
I knew that cmake could do this but I didn't think it takes that little code.

I very much like the idea of people being able to manage their polybar installation through their package manager, even if they compile from source. 

> When installing the package from my local repo, APT will do the dependency resolution for me and only install the runtime dependencies.

The package created on my system doesn't have any dependencies listed. It even tells you that:

> -- CPACK_DEBIAN_PACKAGE_DEPENDS not set, the package will have no dependencies.

Is this different on systems with apt?


The main issue I have with this is that we don't want to have code in our repo that is not actively maintained. I don't use a distro with apt or rpm, so to test this I will each time need to load up two different VMs and I don't have the capacity to dedicate that much time for something like this. This means unless @Lomadriel or @NBonaparte uses a distro with apt or rpm, there is no one on the team that will really maintain this code.

I was already quite conflicted about adding a [polybar.spec](https://github.com/polybar/polybar/blob/master/contrib/rpm/polybar.spec) file for openSUSE in the `contrib` folder, but there the author said they would update it regularly and secondly the `contrib` folder is meant more for that kind of thing while this PR would actually change the main build flow (even if it is off by default).

I don't really know what to do here, because the benefits are quite nice. @Lomadriel, @NBonaparte what are your thoughts on this?

---

There is also a repo that allows you to build polybar in a container (and create `.deb` packages): https://github.com/ayosec/polybar-debian

And polybar is coming to the official fedora repos: https://github.com/polybar/polybar/pull/1976

hmenke:
Oh, I don't think the DEB package can be reliably built on a non-APT system (similarly for the RPM package).  On an APT-based system, make sure you have `dpkg-dev` installed, then the line
```cmake
set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS ON)
```
will use `dpkg-shlibdeps` to automagically determine which packages are needed to fulfill the dynamically-linked dependencies.

The same holds for RPM-based systems where the magic line is:
```cmake
set(CPACK_RPM_PACKAGE_AUTOREQPROV ON)
```
This works a little bit different an requires the existence of the `file` utility.

This also means that the new CMake code needs exactly zero maintenance because as long as the `make install` target works correctly, `make package` will work as well (on a supported distribution of course).

Further I have thought about the interface again and come to the conclusion that using an extra `ENABLE_PACKAGE` option does not make too much sense and it is much more convenient to hijack the existing `CPACK_GENERATOR` variable for that purpose.  That way I can say for example
```cmake
$ cmake -DCPACK_GENERATOR='DEB;RPM;TGZ' ..
```
to build packages for all generators or just
```cmake
$ cmake -DCPACK_GENERATOR=DEB ..
```
to build only the Debian package.

On another branch (https://github.com/hmenke/polybar/commit/7cd281fbfec535d677b776a42a6d89749a200ddc) I have also integrated building the Debian package into the Travis CI workflow, but that became sort of ugly.  Right now, I'm using `aptly` to set up a local repo that contains the polybar package that falls out of CMake.  Then I fire up a `ubuntu:bionic` Docker container in which I install the local repo and install the polybar package.  After that I check whether `ldd` can resolve all the dynamically linked dependencies. I can understand if you don't want to have this sort of stuff in the build system.

patrick96:
The RPM generator actually did create a proper dependency list (though in rpm in general dependencies are to specific library files and not package names).

> This also means that the new CMake code needs exactly zero maintenance because as long as the make install target works correctly, make package will work as well (on a supported distribution of course).

This assumes there are no errors in the cmake code responsible for configuring cpack. That was my actual concern that the package files generated had some kind of issue (missing files, wrong metadata) that didn't allow it to run (or be installed).

This is something where travis could be very useful.

> I can understand if you don't want to have this sort of stuff in the build system.

Actually, this is something I would like to have in the travis config because it allows us to test the cpack functionality. I does look rather unreadable if it's all written directly into the travis config file, but adding scripts in `common/travis` to automate packaging and testing inside docker seems reasonable to me.

patrick96:
Considering that polybar is now available in debian and fedora as a native package and also as an rpm download from OBS for openSUSE, this PR is largely obsolete.

If someone wants to build a package for a non-release commit, they can also use [`checkinstall`](https://wiki.debian.org/CheckInstall).

hmenke:
Fair enough. Thanks for considering my contribution.

patrick96:
Thank you for contributing, even if it didn't make it into the project :)

PR #1979, opened 2020-01-01
Added Snap package

Mandy91:
I created a Snapcraft config to package Polybar as a Snap

It is published in the Snap store as [polybar-git](https://snapcraft.io/polybar-git) (polybar was already taken).

Please let me know if I need to add some documentation somewhere

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@
 *.swp
 *.tmp
 .tags
+*.snap
+parts
+.snapcraft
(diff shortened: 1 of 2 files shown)


patrick96:
I think having a polybar snap would be nice. Especially if we could automate it with travis to create a new snap on every release. I think I can also try to claim the polybar name in the snap store.

@Mandy91 Do you still plan to work on this?

Mandy91:
@patrick96 Yes I am still working on it, I am struggling with getting it connected to the i3 socket however. Everything else is working. I'll try to get it upload to the snap store so you can check it out

patrick96:
Great!

> I am struggling with getting it connected to the i3 socket however.

Can you share some error messages. The path to the i3 socket is generally provided directly by the running i3 instance with `i3 --get-socketpath`, so the path should be correct.

PR #1929, opened 2019-11-01
fix(xworkspaces): Assign desktops to correct viewport

patrick96:
Before the module would just try to evenly distribute desktops
(workspaces) among the viewports.

But since `_NET_DESKTOP_VIEWPORT` actually maps desktops to viewports,
we can use that information to assign workspaces to the right viewport.

I wasn't able to test this yet since I don't have access to my multimonitor setup right now.

Fixes #1849
Fixes #1764

PR #1984, opened 2020-01-07
fix(xworkspaces): Show _NET_NUMBER_OF_DESKTOPS desktops

### LINKED ISSUE #1983: xworkspaces showing additional non existent workspace
## Describe the issue

The xworkspaces module will show one desktop/workspace too many when the window manager does not provide workspace names (such as 2bwm).

**Expected behavior:**

The number of workspaces displayed should be equal to the actual number of workspaces.

**Actual behavior:**

When no names are provided the number of names is smaller than the number of workspaces.
This results in the names vector being filled with n+1 values even though there are only n workspaces thus adding one name too many.
https://github.com/polybar/polybar/blob/d93a50836dd45287a463235666b5573f2e5c20f6/src/modules/xworkspaces.cpp#L234-L247

The +1 would need to be removed.

**Was it working before?**
No, this bug existed since the xworkspaces module was merged.

## To Reproduce
```dosini
[bar/example]
center = ewmh
[module/ewmh]
type = internal/xworkspaces
pin-workspaces = false
```
A window manager that doesn't set workspace names such as 2bwm is necessary.
Default 2bwm will have 10 workspaces so the bar will display 11 instead.

## Polybar Log
No relevant logs since no crashes etc occur.

## Environment:
* WM: 2bwm
* Distro: Arch 
* `polybar -vvv`:
```
polybar 3.4.0-57-g2cbbc86

Build type: Release
Compiler: /usr/bin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -Wall -Wextra -Wpedantic -O3 -DNDEBUG
```

patrick96:
In the case where _NET_NUMBER_OF_DESKTOPS > |_NET_DESKTOP_NAMES|
the last branch of the get_desktop_names method would return a vector
with _NET_NUMBER_OF_DESKTOPS + 1 elements because we iterate until
_NET_NUMBER_OF_DESKTOPS + 1.

Fixes #1983

### DIFF
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -5,12 +5,11 @@
 
 #include "drawtypes/iconset.hpp"
 #include "drawtypes/label.hpp"
+#include "modules/meta/base.inl"
 #include "utils/factory.hpp"
 #include "utils/math.hpp"
 #include "x11/atoms.hpp"
 #include "x11/connection.hpp"
-
-#include "modules/meta/base.inl"
 
 POLYBAR_NS
 
@@ -240,8 +239,8 @@ namespace modules {
       names.erase(names.begin() + desktops_number, names.end());
       return names;
     }
-    for (unsigned int i = names.size(); i < desktops_number + 1; i++) {
-      names.insert(names.end(), to_string(i));
+    for (unsigned int i = names.size(); i < desktops_number; i++) {
+      names.insert(names.end(), to_string(i + 1));
     }
     return names;
   }


Lomadriel:
I just tested your PR and indeed it fixes this bug but there are still many bugs in the workspace module :cry: 
For example if I've the following x configuration:
```
xrandr --listactivemonitors                                                                                                                                                       !10106
Monitors: 2
 0: +*DP1 3440/800x1440/330+1920+0  DP1
 1: +eDP1 1920/380x1080/210+0+0  eDP1
```
The xworkspace module gives me the following output:
```2 4```
but the i3 module gives me:
```7 | 1 2 4```
| = screen separation.

Xworkspace config:
```
type = internal/xworkspaces
pin-workspaces = true
format = <label-state>
```
However if I have the following config:
```
type = internal/xworkspaces
pin-workspaces = false
format = <label-state>
```
It works as intended for this case but if I modify my workspace like this:
```7 | 1 2 3 4```
the xworkspace module gives:
``` 7 | 1 2 3```
:/

I checked every time with `xprop -root` the result is correct. This isn't an i3 bug.

patrick96:
Those all seem like issues related to assigning workspaces to monitors. Can you test if those errors also occur in my PR that's supposed to fix this #1929?

PR #2008, opened 2020-02-12
Created network latency module

zampierilucas:


src/adapters/net.cpp
```diff
@@ -136,6 +136,21 @@ namespace net {
     }
   }
 
+  /**
+   * Get current latency to the internet
+   */
+  string network::latency() const {
+    try {
+  	auto exec = "ping -c1 google.com 2>&1 | grep -Po 'time=.*' | cut -c 6-i";
```
zampierilucas: This probably can be done in a more elegant way...

PR #2009, opened 2020-02-12
Created network latency check module

zampierilucas:


src/adapters/net.cpp
```diff
@@ -136,6 +136,19 @@ namespace net {
     }
   }
 
+  /**
+   * Get current latency to the internet
+   */
+  string network::latency() const {
+    
+     string latency = "";
```
patrick96:
```suggestion
```

src/modules/network.cpp
```diff
@@ -111,6 +111,7 @@ namespace modules {
 
     auto upspeed = network->upspeed(m_udspeed_minwidth);
     auto downspeed = network->downspeed(m_udspeed_minwidth);
+    auto latency = network->latency();
```
patrick96: This would stall the update of the function by whatever the latency is. So a high latency connection would have much slower update rates for the network module.

zampierilucas:
> There already exists a polybar-script for this task: [info-pingrtt](info-pingrtt)
> 
> A native solution would be nice, but it would also increase the complexity of polybars code.
> Since the native code does not much more then a ping command I see no reason not to use a script for this.

I see, I will close this PR then, thanks

PR #2015, opened 2020-02-22
build: Update i3ipcpp submodule

patrick96:
With this we no longer distribute the jsoncpp 1.7.7 source with polybar.
It is possible that polybar will no longer build on older systems that
have an ancient version of jsoncpp (>= 5 years).

Ref: polybar/i3ipcpp#9

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ addons:
         - libxcb1-dev
         - xutils-dev
         - i3-wm
+        - libjsoncpp-dev
         - libasound2-dev
         - libpulse-dev
         - libcairo2-dev


Lomadriel:
Shouldn't we update the dependencies ?

patrick96:
Which dependencies?

samueloph:
Just giving a suggestion, if you're aware of a minimum version required for the lib that polybar needs to work, you can declare that in the README, so packagers will be aware of that.

But I don't think you're gonna have problems with any distro, the ones which are old enough for this be an issue are not packaging the latest releases, and if they are, they can also package the newer lib.

patrick96:
The problem is, we're not really aware of a minimum version :man_facepalming: 

The i3ipcpp project did compile version 1.7.7 if a more recent version couldn't be found, but in your package, you just decrease that to 1.7.4, so versions lower than 1.7.7 seem to work, we just don't know how far down.

samueloph:
@patrick96 that's ok, it's very common to not be aware of required version of dependencies, and in this case you can just ignore it until some issue arrives. But I'm pretty sure that no issue will show up, as we know that an "old enough" version is supported.

Lomadriel:
> Which dependencies?

My bad, "jsoncpp" was already in the dependency list :man_facepalming:

PR #2026, opened 2020-03-01
Checks if the configuration file exists

nekiter:
**Description**

Today when polybar start it does the following configuration search:

1.  ` -c/--config ` parameter
2. `$XDG_CONFIG_HOME/polybar/config` , if the variable is set
3. `$HOME/polybar/config` , if the variable is set
4. Throw error

But it doesn't check if the file exists, this pull request solves that, so the new configuration search is:

1.  `-c/--config`  parameter (if specified)
2. `$XDG_CONFIG_HOME/polybar/config` , if the file exists (variable still being checked)
3. `$HOME/.config/polybar/config` , if the file exists (variable still being checked)
4. Throw error

**Solves**

This PR closes the issue [2016](https://github.com/polybar/polybar/issues/2016), this issue also ask for a change that makes a lot of sense but it'd be another PR:

move the example of config file from `.../doc/polybar/config` to `.../doc/polybar/examples/config` , that change need an update on wiki.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -92,7 +92,7 @@ int main(int argc, char** argv) {
     //==================================================
     // Load user configuration
     //==================================================
-    string confpath = "";
+    string confpath;
 
     // Make sure a bar name is passed in
     if (!cli->has(0)) {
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -106,11 +106,20 @@ int main(int argc, char** argv) {
...
+        confpath = "";
+      }
+    }
+    if (env_util::has("HOME") && confpath.empty()) {
       confpath = env_util::get("HOME") + "/.config/polybar/config";
-    } else {
+      if (!file_util::exists(confpath)) {
+        confpath = "";
+      }
+    }
```
patrick96: I would suggest putting this part into a separate function here in `main.cpp`. This way you don't have to use the workaround of setting `confpath` to the empty string everytime and can just return early.

nekiter: I've made this change, but I'm not sure if the place that I putted the function is appropriated, I was thinking about put it on config class, but it would force polybar instantiate the config class earlier making this small change affect more the code. What do u think?

patrick96: I think it would also have been fine in `main.cpp`, but this is also good.

PR #2027, opened 2020-03-01
Log notice

patrick96:
**Release Note:**
This requires an update on the font wiki page where it suggests using info logging. It also requires updating the home wiki page with the new usage.
**/Release Note:**

This adds a new default log level 'notice' for things the user should know about but that aren't warnings. Please read my commit messages for more details and motivations.

### DIFF
--- a/contrib/bash/polybar
+++ b/contrib/bash/polybar
@@ -42,6 +42,7 @@ _polybar() {
 
   local log_levels='error
                     warning
+                    notice
                     info
                     trace'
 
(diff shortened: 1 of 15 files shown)


patrick96:
> Maybe we should add an explanation about that log level into the man

What did you have in mind? Like the purpose of each logging level?

Lomadriel:
> What did you have in mind? Like the purpose of each logging level?

Not every logging level, just the `notice` one.

patrick96:
I don't quite understand why we would only add an explanation for the notice level and no the others.

PR #2029, opened 2020-03-04
Update GitHub API deprecating authentication

nekiter:
**Description** 

The github module only authenticate by query string, and this method is deprecated:
https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters

There is no reason to remove it before the method stop working, so I've made possible to the user choose which authentication method he will use:

* The parameter token remain unchanged.
* If the parameter user is passed then the module will use the not deprecated method, passing user and token on the body of the requisition. Otherwise the module will use the deprecated method.

PS: Some of the files that I've worked on this PR was not on Polybar code style, and that makes some style updates.

**Solves**

This PR closes the issue [2002](https://github.com/polybar/polybar/issues/2002).

**Extra changes**

It's nice to update [this page](https://github.com/polybar/polybar/wiki/Module:-github) of Polybar wiki, here a [pastebin](https://pastebin.com/cqwFpBWn) with the suggested modification.

### DIFF
--- a/include/modules/github.hpp
+++ b/include/modules/github.hpp
@@ -18,7 +18,6 @@ namespace modules {
     bool build(builder* builder, const string& tag) const;
     string get_format() const;
 
-
    private:
     void update_label(int);
     int get_number_of_notification();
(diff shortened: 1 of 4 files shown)


patrick96:
@Lomadriel, could you review this? I will have barely any capacity the next three weeks.

patrick96:
Thanks! :D

PR #2037, opened 2020-03-09
Allow viewing free memory via memory module

### LINKED ISSUE #2028: Add "free memory" to the memory module's label options
There are two ways to measure memory usage in Linux:

* **Available** memory is being used for something, but is *readily* available. Linux does this to run programs more efficiently if RAM is not being used.
* **Free** memory is not being used for anything. You could call it wasted RAM. It is almost always very low and lower than available memory.

[linuxatemyram.com](https://www.linuxatemyram.com/) explains this concept a little bit.

Currently, polybar's `memory` module only allows displaying available memory. For most users, this is all they care about- it is a better representation of how much memory is available. 

Free memory is also a metric of interest for some users, including myself.

## Current Situation

The `memory` module currently labels memory as follows:

* `%percentage_free%`: percentage of memory that is **free** memory
* `%percentage_used%`: `100 - %percentage_free%`

## Possible Solution

An option should be made to display this "available memory" via the `memory` module, accessible in the `label` via `%percentage_available%` and `%percentage_available_used%` like so:

* `%percentage_available_used%`: memory owned by processes and unusable
* `%percentage_available%`: memory available for use
* `%percentage_free_used%`: memory being used (will almost always be very high)
* `%percentage_free%`: memory completely unused (will almost always be very low)

> This naming scheme could definitely be improved.

This way, I can see how much RAM is being *used* but is *available* (100 - available memory), and also see how much RAM is truly being used (100 - free memory).

jallbrit:
Add labels to memory module that allow users to display both free and
available memory separately. Closes #2028.

Free memory is memory that is truly not being used, while available
memory includes memory that is technically being used but is readily
available for other processes.

### DIFF
--- a/include/modules/memory.hpp
+++ b/include/modules/memory.hpp
@@ -8,6 +8,7 @@ POLYBAR_NS
 namespace modules {
   enum class memtype { NONE = 0, TOTAL, USED, FREE, SHARED, BUFFERS, CACHE, AVAILABLE };
 
+
   class memory_module : public timer_module<memory_module> {
    public:
     explicit memory_module(const bar_settings&, string);
@@ -31,6 +32,10 @@ namespace modules {
     progressbar_t m_bar_memfree;
     int m_perc_memused{0};
     int m_perc_memfree{0};
+    int m_perc_memavail{0};
+    int m_perc_memavailused{0};
+    int m_perc_memactualfree{0};
+    int m_perc_memactualfreeused{0};
     ramp_t m_ramp_memused;
     ramp_t m_ramp_memfree;
     progressbar_t m_bar_swapused;
--- a/src/modules/memory.cpp
+++ b/src/modules/memory.cpp
@@ -52,6 +52,7 @@ namespace modules {
   }
 
   bool memory_module::update() {
+    unsigned long long kb_free{0ULL};
     unsigned long long kb_total{0ULL};
     unsigned long long kb_avail{0ULL};
     unsigned long long kb_swap_total{0ULL};
@@ -73,6 +74,7 @@ namespace modules {
         parsed[id] = value;
       }
 
+      kb_free = parsed["MemFree"];
       kb_total = parsed["MemTotal"];
       kb_swap_total = parsed["SwapTotal"];
       kb_swap_free = parsed["SwapFree"];
@@ -92,6 +94,10 @@ namespace modules {
 
     m_perc_memfree = math_util::percentage(kb_avail, kb_total);
     m_perc_memused = 100 - m_perc_memfree;
+    m_perc_memavail = math_util::percentage(kb_avail, kb_total);
+    m_perc_memavailused = 100 - m_perc_memavail;
+    m_perc_memactualfree = math_util::percentage(kb_free, kb_total);
+    m_perc_memactualfreeused = 100 - m_perc_memactualfree;
     m_perc_swap_free = math_util::percentage(kb_swap_free, kb_swap_total);
     m_perc_swap_used = 100 - m_perc_swap_free;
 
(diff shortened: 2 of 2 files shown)


include/modules/memory.hpp
```diff
@@ -31,6 +32,10 @@ namespace modules {
     progressbar_t m_bar_memfree;
     int m_perc_memused{0};
     int m_perc_memfree{0};
+    int m_perc_memavail{0};
```
Lomadriel: I think we shouldn't have two variables storing the same thing. We should keep the new ones and removed the old ones.

src/modules/memory.cpp
```diff
@@ -106,6 +112,10 @@ namespace modules {
       m_label->replace_token("%mb_total%", string_util::filesize_mb(kb_total, 0, m_bar.locale));
       m_label->replace_token("%percentage_used%", to_string(m_perc_memused));
```
Lomadriel: Should we deprecate these labels @patrick96 ?

src/modules/memory.cpp
```diff
@@ -106,6 +112,10 @@ namespace modules {
...
       m_label->replace_token("%percentage_free%", to_string(m_perc_memfree));
+      m_label->replace_token("%percentage_available%", to_string(m_perc_memavail));
+      m_label->replace_token("%percentage_available_used%", to_string(m_perc_memavailused));
+      m_label->replace_token("%percentage_actualfree%", to_string(m_perc_memactualfree));
```
Lomadriel: I don't really like the name of these labels, but I don't have a better idea for now :/

include/modules/memory.hpp
jallbrit: Would this not be considered a breaking change?

Lomadriel: I just want to remove the variables `m_perc_memused` and `m_perc_memfree`, not the labels ^^

ExperiBass:
Is this PR still active? if not, can i try and re-implement it with a updated master?

patrick96:
@ExperiBass I'm not sure where to go from here. The main issue is naming. Ideally the `*_free` tokens would show `MemFree` and the `*_available` tokens would show `MemAvailable`. However, that would change what everyone's modules display. As I already described in #2028, we should introduce `*_available` tokens that display the same as the `*_free` tokens and deprecate the `*_free` tokens. This way, in the 4.0.0 release we can introduce a breaking change and make the `*_free` tokens actually use `MemFree`.

This leaves two naming issues.
First, how to name the tokens that represent `MemFree` in the meantime? This PR uses `*_actualfree`, but I'm not the biggest fan. Additionally, this is temporary name is just one that we will need to deprecate in the future.

Second, how to call the tokens for the "used" counterparts? Currently `*_used` tokens represent the value `MemTotal - MemAvailable`, but we will also need a name for tokens that represent `MemTotal - MemFree`. Here I think using `*_used_available` and `*_used_free` should be fine.

Before we figure out the naming for these tokens and the associated progressbars and ramps, it makes little sense to actually write any code and update the PR.

If you have any ideas or inspiration, I'd appreciate your input :)

ExperiBass:
> @ExperiBass I'm not sure where to go from here. The main issue is naming. Ideally the `*_free` tokens would show `MemFree` and the `*_available` tokens would show `MemAvailable`. However, that would change what everyone's modules display. As I already described in #2028, we should introduce `*_available` tokens that display the same as the `*_free` tokens and deprecate the `*_free` tokens. This way, in the 4.0.0 release we can introduce a breaking change and make the `*_free` tokens actually use `MemFree`.

i agree with that, it should be deprecated to give everyone warning and then make the breaking change in the 4.0 release.

> This leaves two naming issues. First, how to name the tokens that represent `MemFree` in the meantime? This PR uses `*_actualfree`, but I'm not the biggest fan. Additionally, this is temporary name is just one that we will need to deprecate in the future.

i cant think of anything better than `*_actualfree` for the time being, its clunky but it works and thats what matters.

> Second, how to call the tokens for the "used" counterparts? Currently `*_used` tokens represent the value `MemTotal - MemAvailable`, but we will also need a name for tokens that represent `MemTotal - MemFree`. Here I think using `*_used_available` and `*_used_free` should be fine.

yeah, `*_used_free/_available` matches the naming scheme of the other tokens.

> Before we figure out the naming for these tokens and the associated progressbars and ramps, it makes little sense to actually write any code and update the PR.
> 
> If you have any ideas or inspiration, I'd appreciate your input :)

i think the current naming idea fits, it doesnt really affect ricing other than adding a little bit more characters to type. id say go for it.

src/modules/memory.cpp
```diff
@@ -18,19 +18,31 @@ namespace modules {
   memory_module::memory_module(const bar_settings& bar, string name_) : timer_module<memory_module>(bar, move(name_)) {
     m_interval = m_conf.get<decltype(m_interval)>(name(), "interval", 1s);
 
-    m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL, TAG_BAR_USED, TAG_BAR_FREE, TAG_RAMP_USED, TAG_RAMP_FREE,
+    m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL, TAG_BAR_AVAILABLE_USED, TAG_BAR_AVAILABLE_FREE, TAG_BAR_FREE_USED, TAG_BAR_FREE, TAG_RAMP_AVAILABLE_USED, TAG_RAMP_AVAILABLE_FREE, TAG_RAMP_FREE_USED, TAG_RAMP_FREE,
```
koterevma, line 21: This single line is more than 200 characters long. It's harder to read and understand. Maybe split it into 3-4 lines?

koterevma (approved):
Changes are great, can't wait to see them in beta!

PR #2054, opened 2020-04-08
net: increase speedrate precision

jpalus:
more granular units need lower precision, while less granular need
higher precision. assume sane default of:

unit | precision
KB   | 0
MB   | 1
GB   | 2


patrick96 (approved):
LGTM

patrick96:
Thanks :rocket:

minhdanh:
@jpalus @patrick96 This looks like not working correctly for me. I'm using polybar version 3.5.6, and the network module always shows the download/upload speeds without precision.

I'm reporting this because I use another tmux plugin to monitor the network speed. When the download speed of this plugin reports 2.65MB/s, the network module of Polybar shows 3 MB/s. Never a decimal for the precision.
Maybe we need unit tests to make sure this work properly?

patrick96:
Correct. This feature will be part of 3.6.0

minhdanh:
Oh I see. My bad. I supposed a PR merged 10 months ago should have made it to the latest version now :sweat_smile:

PR #2068, opened 2020-04-21
module::network: Ability to change speed unit text (B/s)

CromFr:
I feel like having `B/s` at the end of the speed value is generally not necessary and takes too much space.

This pull request adds a `udspeed-unit` variable to be able to display the unit text in other formats. By default the value is `B/s` to keep the old behaviour. The K/M/G suffixes are still hardcoded and are displayed just before the unit.
Note: this is a "text-only" feature, setting `speed-unit = "b/s` won't change the output value from bytes per second to bits per second.

Examples:
`speed-unit = ""` will display speeds like: `128 ` `42K` `64G`
`speed-unit = "Bytes/sec"` will display speeds like: `128 Bytes/sec` `42KBytes/sec` `64GBytes/sec`

### DIFF
--- a/include/adapters/net.hpp
+++ b/include/adapters/net.hpp
@@ -82,14 +82,14 @@ namespace net {
 
     string ip() const;
     string ip6() const;
-    string downspeed(int minwidth = 3) const;
-    string upspeed(int minwidth = 3) const;
+    string downspeed(int minwidth = 3, const string& unit = "B/s") const;
+    string upspeed(int minwidth = 3, const string& unit = "B/s") const;
     void set_unknown_up(bool unknown = true);
 
    protected:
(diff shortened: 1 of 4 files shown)


kronn:
Would it make sense to rename the setting to `speed-unit`? The `ud` makes sense when you read the source, but feels like a typo otherwise :-)

CromFr:
The `udspeed-unit` name was chosen to match the other `udspeed-minwidth` setting, but since it is deprecated, it makes sense to change the name to `speed-unit` as you suggested :)

I also fixed the merge conflicts

patrick96 (approved):
LGTM!

Thanks @CromFr and thank you @kronn for the review.

PR #2074, opened 2020-04-25
Add low and critical battery states

wz1000:
Use `low-at` and `crit-at` configuration variables to control thresholds. Set to 30% and 10% respectively by default.


wz1000:
Superseded by #2199

PR #2084, opened 2020-05-03
Creaded dynamic interface string

zampierilucas:
Added the possibility to Polybar to detect current in-use network adapter by setting "interface = <default-interface>" on config file.


patrick96:
Hey, thanks for contributing! But, there is already a PR at #2025 that does the same thing. It also does not rely on external shell commands.

Closing in favor of #2025.

PR #2089, opened 2020-05-06
build: Update xpp module

patrick96:
Fixes developer warnings in cmake 3.17+

Ref: polybar/xpp#24


Lomadriel:
Hum it doesn't seem to compile on travis

patrick96:
That's weird. This is actually the invocation of `find_package_handle_standard_args` where I barely changed anything and according to the cmake docs my change shouldn't have changed any functionality.

I will try to debug this in a VM with an older cmake version.

patrick96:
:man_facepalming: Older cmake version interpret the `NAME_MISMATCHED` flag as a variable it needs to check instead of as a flag.

I guess we're not getting around using `FPHSA_NAME_MISMATCHED`. I will quickly push that change to xpp and then update this PR

patrick96:
> I hope it works for all versions x)

It should the function signature I used exists since cmake 3.0

PR #2117, opened 2020-05-30
refactor: Deprecate throttle-input-for setting

patrick96:
If an input is enqueued as a response to an input, the new input will be
swallowed because it will likely be enqueued less than 30ms after the
original event.
This is not something that is an issue right now but it is required to
finish #1907 where, in order to close the menu after a click, the menu
module gets an exec action that closes the menu and adds a command to
the event queue.

The setting also isn't too useful since it will just break polybar input
handling if inputs arrive too fast instead of (possibly) slowing down
the bar.

### DIFF
--- a/include/components/controller.hpp
+++ b/include/components/controller.hpp
@@ -132,16 +132,6 @@ class controller
   std::chrono::milliseconds m_swallow_update{10};
 
   /**
-   * \brief Time to throttle input events
-   */
-  std::chrono::milliseconds m_swallow_input{30};
-
-  /**
-   * \brief Time of last handled input event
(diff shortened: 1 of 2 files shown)


patrick96:
I think the original intention was to not have polybar completely break down if for some reason it gets flodded with inputs.

But I don't think polybar would break down because there can only ever be one unhandled input (though I'd like to change that as well) and if an input can't be handled fast enough, it will just be dropped.

I have also just tried to create an action that adds itself as a new action to the controller and polybar didn't have any issue catching up and I didn't see any increased CPU or memory load.

PR #2122, opened 2020-06-05
Change "amdgpu_bl0" so the "0" isn't hardcoded in

nicey0:
Some devices can have "amdgpu_bl1" or "amdgpu_bl2", but the code hardcoded in the value "amdgpu_bl0". This change tests based on the first characters: "amdgpu_bl".


patrick96:
Thanks :D

PR #2123, opened 2020-06-06
fix(timer_module): sleep until next full interval

### LINKED ISSUE #2064: "Lock" timer_module updates to whole seconds instead of simply sleeping for some amount
The easiest way to explain this is with an example. Imagine a module like this:

```ini
[module/date]
type = "internal/date";
interval = 60;
time = "%H:%M";
```

This would update the clock once every minute, which seems ideal, but it's not entirely accurate, since it doesn't update *at the start of a new minute*.

It would be more obvious if you also show the seconds; you would see 08:10:37, 08:11:37, 08:12:37, etc.

In addition, the inaccuracy will actually drift over time because there's actually `(interval + time_to_update)` seconds between each update; something like 1.002s.

My suggestion would be to sleep until the start of the next {minute, second, whatever interval is specified}. So instead of sleeping for `interval_seconds`, it would be sleeping for something like `interval_seconds - (current_real_time_in_seconds % interval_seconds)`. With an interval of 5 seconds, this might end up sleeping for something like 3.245s the first time depending on when polybar was launched to sync up with the clock, and then 4.999s every time.

This would make clocks more accurate, with virtually no downside to pretty much all users.

I guess you could hack something like this together with custom script modules using tail, but I believe this should be part of polybar itself. [i3status also does this.](https://github.com/i3/i3status/blob/b61e959a58750fe3a4cf9ed2472d4cb7112bd2c3/i3status.c#L855) Currently, to make clocks like `%H:%M` accurate, you'd need an interval of a (few) second(s), but that would mean the module would update too often, wasting hundreds of precious CPU cycles!!!

sl33k:
Any timer_module based module would sleep for the set interval and then
continue running. Depending on the start time of polybar this
sleep pattern might not be aligned, which causes such modules to always
update in a shifted manner.
Consider the date module as an example. If the update interval is set to
60 seconds and polybar was started at 13:37:37, polybar would update the
clock at 13:38:37, 13:39:37 and so on.
To make matters worse, if a module would perform lengthy checks this
interval might drift over time, causing even more inconsistent updating.

This patch extends the base module with a sleep_until method that calls
the corresponding function on the sleephandler. Additionally the
timer_module is extended to compute the remaining time until the next
interval passes and sleep accordingly.

Closes #2064

Co-developed-by: Dominik Töllner <dominik.toellner@stud.uni-hannover.de>

### DIFF
--- a/include/modules/meta/timer_module.hpp
+++ b/include/modules/meta/timer_module.hpp
@@ -34,7 +34,21 @@ namespace modules {
           if (check()) {
             CAST_MOD(Impl)->broadcast();
           }
-          CAST_MOD(Impl)->sleep(m_interval);
+          // wait until next full interval to avoid drifting clocks
+          using clock = chrono::system_clock;
+          using sys_duration_t = clock::time_point::duration;
+
+          auto sys_interval = chrono::duration_cast<sys_duration_t>(m_interval);
+          clock::time_point now = clock::now();
+          sys_duration_t adjusted = sys_interval - (now.time_since_epoch() % sys_interval);
+
+          // The seemingly arbitrary addition of 500ms is due
+          // to the fact that if we wait the exact time our
+          // thread will be woken just a tiny bit prematurely
+          // and therefore the wrong time will be displayed.
+          // It is currently unknown why exactly the thread gets
+          // woken prematurely.
+          CAST_MOD(Impl)->sleep_until(now + adjusted + 500ms);
         }
       } catch (const exception& err) {
         CAST_MOD(Impl)->halt(err.what());
@@ -44,6 +58,6 @@ namespace modules {
    protected:
     interval_t m_interval{1.0};
   };
-}
+}  // namespace modules
 
 POLYBAR_NS_END
--- a/include/modules/meta/base.hpp
+++ b/include/modules/meta/base.hpp
@@ -37,7 +37,7 @@ namespace drawtypes {
   using animation_t = shared_ptr<animation>;
   class iconset;
   using iconset_t = shared_ptr<iconset>;
-}
+}  // namespace drawtypes
 
 class builder;
 class config;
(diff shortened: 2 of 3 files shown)


include/modules/meta/timer_module.hpp
```diff
@@ -34,7 +34,24 @@ namespace modules {
...
+          // The seemingly arbitrary addition of 1ms is due
+          // to the fact that if we wait the exact time our
+          // thread will be woken just a tiny bit prematurely
+          // and therefore the wrong time will be displayed.
+          // Optimally we would round to the next smaller
+          // increment of m_interval (e.g. ms), but the required
+          // functions are c++ 17 only.
+          // For this simple use case, where m_interval is always
+          // in seconds a simple 'magic' constant suffices.
+          CAST_MOD(Impl)->sleep_until(now + adjusted + 1ms);
```
patrick96, line 54: Even with the 1ms added, I experience my date module displaying XX:59 with an interval of 60s. I think we can get away with adding 500ms here.

Do we know why the condition_variable wakes up too early? I can understand why it would wake up later, but not earlier.

sl33k: Unfortunately not no.

We initially though it was due to rounding issues but have since ruled that out.
We then suspected possible call reordering (e.g. hoisting both the `std::time_t(nullptr)` call and the `chrono::system_clock::now()` to the beginning of the `runner()` function to maximize cache efficiency, but this also proved to be a fruitless effort.
Spurious wake-ups have also been ruled out as the `condition_variable::wait_until`, which is used internally, does return that the timeout was reached.
It seems like this might be a bug in the `stl` or possibly `libc` (as the `stl` uses [`pthread_cond_wait`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_cond_timedwait.html) internally or due to some other reason which we haven't considered.

In any case, I've raised the magic addition to `500ms`.
Please let me know if you have further possible ideas for explaining this strange behaviour.

Issue #2141, opened 2020-07-08
[Feature Request] Add different formats for memory module depend on used percentage


juliancoffee:
## Is your feature request related to a problem? Please describe.
Right now polybar has different options to show memory level, like bar-used, ramp-used, etc. But it cannot show different labels depend on used percentage of memory, for example, one label for normal usage, another format for 60%, and third format when memory usage hit 95% and one need think about memory consumption and probably reboot a browser.

## Why does polybar need this feature?
With this feature polybar users will be able to hide memory block when memory usage is normal, show when it's worth noting and make screaming red label when the memory consumption hits the critical level.
Like this is possible with the temperature module.

## Describe the solution you'd like
Add this option to `internal/memory`
- `warn-at` - to specify a percentage of used memory on which choose warn
- `critical-at` - to specify the critical percentage of used memory
- `<label-warn>` - like it is in the temperature module
- ` format-warn` - like it is in the temperature module
- `<label-critical>` - same as on warn
- `format-critical` - same as on warn

## Describe alternatives you've considered
I know there is a custom/script module and exec-if, but as far as I know it's not so customizable - e.g. one cannot set one colour for prefix, another colour for text and draw one continuous line under this block.

## Additional context
I made own proof of concept of this feature and here is a gif
![poc](https://user-images.githubusercontent.com/42647349/86905140-52c25780-c11a-11ea-84b9-84b6e55d7bec.gif)

juliancoffee:
Looks like it is requested in #570, but I don't know if it is a duplicate, because #570 is a more generic issue.

DRKblade:
I have implemented the feature requested by #570 in the pull request #2199, you might want to try it out.

patrick96:
A single warn format has now been implemented in #2199.

PR #2142, opened 2020-07-08
feat(memory): Add different formats for different memory usage

juliancoffee:
Added and implement these options for #2141 
- `warn-at` - to specify percentage of used memory on which choose warn
- `critical-at` - to specify critical percentage of used memory
- `<label-warn>`
- ` format-warn`
- `<label-critical>`
- `format-critical`

I tried to mimic temperature module behaviour for different temperature level, tell me if I can somehow improve code.
(For example, I don't understand why we need to replace tokens for all labels even when there is only one used. But it works :D )

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 /tags
 /compile_commands.json
 /config
+/.clangd
+/.vscode
 *.bak
 *.pyc
 *.swp
--- a/include/modules/memory.hpp
+++ b/include/modules/memory.hpp
@@ -7,16 +7,20 @@ POLYBAR_NS
 
 namespace modules {
   enum class memtype { NONE = 0, TOTAL, USED, FREE, SHARED, BUFFERS, CACHE, AVAILABLE };
+  enum class mem_state { NORMAL = 0, WARN, CRITICAL };
 
   class memory_module : public timer_module<memory_module> {
    public:
     explicit memory_module(const bar_settings&, string);
 
     bool update();
+    string get_format() const;
     bool build(builder* builder, const string& tag) const;
 
    private:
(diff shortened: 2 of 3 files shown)


godalming123:
Does this mean that I could make my CPU marker red when its above 90% and flash red when its above 95% and the same with my memory marker. If so this would be awsome!

juliancoffee:
> Does this mean that I could make my CPU marker red when its above 90% and flash red when its above 95% and the same with my memory marker. If so this would be awsome!

Only for the memory marker.
Honestly this MR is quite old so I'm not sure if this is still not implemented, wasn't following latest features that much.
But if not, you can use this)

juliancoffee:
```
[module/memory]
type = internal/memory
warn-at = 60
critical-at = 90
interval = 2

format = <label>
label = 

format-warn = <label-warn>
label-warn = %percentage_used%%

format-warn-underline = #ff00ff
format-warn-prefix = "M "
format-warn-prefix-foreground = ${colors.foreground-alt}

format-critical = <label-critical>
label-critical = %percentage_used%% HIGH
format-critical-foreground = #ff0000 
format-critical-underline = #ff00ff

format-critical-prefix = "M "
format-critical-prefix-foreground = ${colors.foreground-alt}
```
This is how my memory module looks like with this patch. Hope this helps.
At normal conditions, there is no text at all, at warn percentage is starts to show, and at critical it is red uppercase text.

patrick96:
This is indeed already implemented now (#2199). The feature will land in 3.6.0, it has the same warning behavior as this PR, there is no critical state though.

Sorry for somehow completely missing this PR.

PR #2151, opened 2020-07-24
Module: dwm

mihirlad55:
This is a dwm module for polybar based off i3 and bspwm. I was careful to follow similar code style when creating the module and updating the cmake files for compiling the module and dwmipcpp.

The module requires the dwmipcpp library as a submodule to simplify communication with dwm.

The module also requires that users' dwm has the IPC patch. The `dwm-anybar` patch I've created optionally improves dwm's managing of external status bars including polybar. I am currently working on getting these patches published on the dwm website.

If there is anything you'd like me to change/improve, feel free to let me know. I know there's a lot of people who've been waiting for some sort of polybar support for dwm.

I'd also love to continue to contribute to polybar and help out where needed.

Some relevant links:
https://github.com/mihirlad55/polybar-dwm-module
https://github.com/mihirlad55/dwm-ipc
https://github.com/mihirlad55/dwm-anybar
https://github.com/mihirlad55/dwmipcpp


mihirlad55:
Update:
The patches are now published on the suckless website:
https://dwm.suckless.org/patches/anybar/
https://dwm.suckless.org/patches/ipc/

src/modules/dwm.cpp
```diff
@@ -0,0 +1,493 @@
...
+        auto addr_next = next_layout(*m_current_layout, m_layout_wrap)->address;
+        auto addr_prev = prev_layout(*m_current_layout, m_layout_wrap)->address;
+        // Set address based on scroll direction
+        auto scroll_down_addr = to_string(m_layout_reverse ? addr_prev : addr_next);
+        auto scroll_up_addr = to_string(m_layout_reverse ? addr_next : addr_prev);
+        builder->cmd(mousebtn::SCROLL_DOWN, build_cmd(CMD_LAYOUT_SET, scroll_down_addr));
+        builder->cmd(mousebtn::SCROLL_UP, build_cmd(CMD_LAYOUT_SET, scroll_up_addr));
+      }
+      builder->node(m_layout_label);
+      if (m_layout_click) {
```
patrick96, line 220: This doesn't make a difference here, but in general, command tags should be closed in the reverse order they are opened in. So first closing the scroll tags and then the click tags (they're the same right now, but that might change in the future).

src/modules/dwm.cpp
```diff
@@ -0,0 +1,493 @@
...
+    } catch (const dwmipc::IPCError& err) {
+      throw module_error(err.what());
+    }
+  }
+
+  void dwm_module::stop() {
+    try {
+      m_log.info("%s: Disconnecting from socket", name());
+      m_ipc.reset(nullptr);
+    } catch (...) {
```
patrick96, line 162: We shouldn't just swallow all exceptions. We should at least print an error or warning message.

include/modules/dwm.hpp
```diff
@@ -0,0 +1,371 @@
...
+     * If the layout symbol is clicked on, it will set the layout represented by
+     * this symbol. The default is monocle mode [M].
+     */
+    string m_secondary_layout_symbol{"[M]"};
+
+    /**
+     * Holds the address to the secondary layout specified by the secondary
+     * layout symbol
+     */
+    const dwmipc::Layout* m_secondary_layout = nullptr;
```
patrick96, line 297: Why do we need a pointer for all these variables?

src/modules/dwm.cpp
```diff
@@ -0,0 +1,493 @@
...
+
+    event_module::stop();
+  }
+
+  auto dwm_module::has_event() -> bool {
+    try {
+      return m_ipc->handle_event();
+    } catch (const dwmipc::SocketClosedError& err) {
+      m_log.err("%s: Disconnected from socket: %s", name(), err.what());
+      sleep(chrono::duration<double>(1));
```
patrick96, line 173: I don't think we should sleep here because we have a lock on `m_updatelock`. At some point we will combine `m_updatelock` and `m_buildlock` into a single mutex to fix some concurrency issues and then sleeping here may block the entire bar.

We could sleep in the `idle` method and schedule a reconnect in the next call to `has_event`. I'm not sure we can reconnect in `idle` because we may run into some concurrency issues.

src/modules/dwm.cpp
```diff
@@ -0,0 +1,493 @@
...
+      // Erase '<ipc_cmd>-'
+      cmd.erase(0, ipc_cmd.size() + 1);
+      m_log.info("%s: Sending workspace %s command to ipc handler", name(), ipc_cmd);
+
+      try {
+        m_ipc->run_command(ipc_cmd, (Json::UInt64)stoul(cmd));
+        return true;
+      } catch (const dwmipc::SocketClosedError& err) {
+        m_log.err("%s: Disconnected from socket: %s", name(), err.what());
+        sleep(chrono::duration<double>(1));
```
patrick96, line 270: Same here. This always blocks the entire bar because `input` is called by the controller

src/modules/dwm.cpp
```diff
@@ -0,0 +1,493 @@
...
+    try {
+      if (!m_ipc->is_main_socket_connected()) {
+        m_log.notice("%s: Attempting to reconnect to main socket", name());
+        m_ipc->connect_main_socket();
+        m_log.notice("%s: Successfully reconnected to main socket", name());
+      }
+      if (!m_ipc->is_event_socket_connected()) {
+        m_log.notice("%s: Attempting to reconnect event socket", name());
+        m_ipc->connect_event_socket();
+        m_log.notice("%s: Successfully reconnected event to socket", name());
```
patrick96, line 439:
```suggestion
        m_log.notice("%s: Successfully reconnected to event socket", name());
```

src/modules/dwm.cpp
```diff
@@ -0,0 +1,493 @@
...
+      }
+      return true;
+    } catch (const dwmipc::IPCError& err) {
+      m_log.err("%s: Failed to reconnect to socket: %s", name(), err.what());
```
patrick96, line 443: What happens if this happens?

src/modules/dwm.cpp
```diff
@@ -0,0 +1,493 @@
...
+    if (m_formatter->has(TAG_LABEL_LAYOUT)) {
+      m_layout_label = load_optional_label(m_conf, name(), "label-layout", "%symbol%");
+    }
+
+    if (m_formatter->has(TAG_LABEL_FLOATING)) {
+      m_floating_label = load_optional_label(m_conf, name(), "label-floating", "");
+    }
+
+    if (m_formatter->has(TAG_LABEL_TITLE)) {
+      m_title_label = load_optional_label(m_conf, name(), "label-title", "%title%");
```
patrick96, line 65: How is this different from the `xwindow` module?

patrick96:
Oh, and of course, thanks for contributing :smiley:

We'd love to have you continue contributing. There are a bunch of issues that need help: https://github.com/polybar/polybar/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22

We also always need help triaging incoming issues and trying to figure out what the problem is.

src/modules/dwm.cpp
```diff
@@ -0,0 +1,493 @@
...
+#include "utils/file.hpp"
+#include "x11/extensions/randr.hpp"
+
+POLYBAR_NS
+
+namespace modules {
+  template class module<dwm_module>;
+
+  dwm_module::dwm_module(const bar_settings& bar, string name_) : event_module<dwm_module>(bar, move(name_)) {
+    string socket_path = env_util::get("DWM_SOCKET");
```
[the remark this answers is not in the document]
mihirlad55: No, it's only used by polybar. I noticed the bspwm module had a similar environment variable which is why I added it. I will change it to a config option.

cmake/02-opts.cmake
```diff
@@ -6,6 +6,7 @@ checklib(BUILD_DOC "binary" "${SPHINX_BUILD}")
 
 checklib(ENABLE_ALSA "pkg-config" alsa)
 checklib(ENABLE_CURL "pkg-config" libcurl)
+checklib(ENABLE_DWM "binary" dwm)
```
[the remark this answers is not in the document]
mihirlad55: No, it's not needed if the module is not running. I'll remove it

include/modules/dwm.hpp
```diff
@@ -0,0 +1,371 @@
...
+
+    using tag_mask_t = unsigned int;
+    using window_t = unsigned int;
+
+    /**
+     * Represents the relevant states a tag can have
+     */
+    enum class state_t : uint8_t {
+      FOCUSED,    ///< Monitor is selected and tag is selected, overrides all below
+      URGENT,     ///< Tag is urgent, overrides all below
```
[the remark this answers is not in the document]
mihirlad55: When I was originally making the module, I thought that if a tag was focused, it wouldn't be urgent. Now that I think about it, it would be possible to have a tag in focus, but one of its unfocused windows have the urgent hint set. I will fix this.

include/modules/dwm.hpp
```diff
@@ -0,0 +1,371 @@
...
+      tag_t(string& name, tag_mask_t bit_mask, state_t state, label_t&& label)
+          : name(name), bit_mask(bit_mask), state(state), label(forward<label_t>(label)) {}
+
+      string name;
+      tag_mask_t bit_mask;
+      state_t state;
+      label_t label;
+    };
+
+    auto stop() -> void override;
```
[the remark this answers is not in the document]
mihirlad55: I noticed that the other modules used this older form. I used the trailing return since clang was telling me to.

src/modules/dwm.cpp
```diff
@@ -0,0 +1,493 @@
...
+    }
+    m_log.info("%s: Connected to dwm socket", name());
+
+    // Load configuration
+    m_formatter->add(
+        DEFAULT_FORMAT, DEFAULT_FORMAT_TAGS, {TAG_LABEL_TAGS, TAG_LABEL_LAYOUT, TAG_LABEL_FLOATING, TAG_LABEL_TITLE});
+
+    // Populate m_state_labels map with labels and their states
+    if (m_formatter->has(TAG_LABEL_TAGS)) {
+      m_state_labels.insert(
```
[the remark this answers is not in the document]
mihirlad55: I agree, I guess it was a lazy copy of the i3 module

src/modules/dwm.cpp
mihirlad55: Ah, didn't catch that. I'll change it for the sake of consistency.

src/modules/dwm.cpp
mihirlad55: I agree, another lazy copy of the i3 module.

include/modules/dwm.hpp
mihirlad55: They don't necessarily need to be pointers. Most of these variables are updated as events are received and used in other logic for determining what click/scroll handlers should do, etc. I believe they can all be replaced by array indices which would be safer to use. The main difference would be `m_secondary_layout->...` vs `m_layouts->at(m_secondary_layout)`. It's more to simplify that initial array access. Would you prefer to replace these pointers with indices?

src/modules/dwm.cpp
mihirlad55: Ok, I'll look into refactoring the disconnects and look more into how the different threads operate.

src/modules/dwm.cpp
mihirlad55: I'll look into it

src/modules/dwm.cpp
mihirlad55: This title label holds the title of the currently focused window on the **bar's monitor**, whereas the `xwindow` module only shows the currently focused window title on the currently focused monitor. This title label allows unfocused monitors to display their focused window's title.

src/modules/dwm.cpp
mihirlad55: There should probably be a throw module_error there.

include/modules/dwm.hpp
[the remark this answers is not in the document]
patrick96: I have never gotten this warning. Where and when exactly do you get this warning?

I guess we should disable it.

src/modules/dwm.cpp
patrick96: The i3 module doesn't always follow best practices ;)

mihirlad55:
Thanks a lot for letting this one slide. I understand your reluctance; it makes perfect sense. Of course, I am completely willing to take responsibility over this module and any issues related to it. If not as part of polybar, I would have continued to maintain the module as a fork anyway. I accept all your conditions, they are all reasonable.

I will address the issues you brought up right away. Regarding the concurrency issues, I don't believe the library is thread-safe. It's not something I was thinking about when making it. To be honest, I'm somewhat new to C++ and don't know too much about thread-safety and concurrency. I'll do some research on my own time and address the issues accordingly.

I do have a question regarding the concurrency though. Is thread-safety something the library should be responsible for or something that polybar as a user of the library should be responsible for?

I look forward to polishing up the module and helping out with polybar's issues. You guys do seem a bit short-staffed for a project this large.

include/modules/dwm.hpp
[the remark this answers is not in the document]
mihirlad55: It's the [modernize-use-trailing-return-type](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-trailing-return-type.html) warning from clang when I declare or define any non-void functions.

include/modules/dwm.hpp
patrick96: Hmm, no I think it looks better this way. I was more wondering why we don't just store the instances directly, but if the underlying data is changed by dwmipcpp, this makes sense.

I guess this introduces more concurrency concerns. Does dwmipcpp start any of its own threads? Does it synchronize accesses to these data structures? Acquiring both the update and build locks in the `input` method should probably suffice.

src/modules/dwm.cpp
patrick96: Let me know if you are stuck anywhere. The concurrency model of polybar isn't exactly transparent, I get confused quite often as well.
I think in the future, we should just put a big lock on anything that calls into the module and on the main module loop as well.

src/modules/dwm.cpp
patrick96: I agree

src/modules/dwm.cpp
patrick96: Neat :)

mihirlad55:
Also, is the code coverage check something I should be concerned about. I believe none of the other modules had any unit tests created.

patrick96:
> I accept all your conditions, they are all reasonable.

That's great to hear :smiley: 

> I do have a question regarding the concurrency though. Is thread-safety something the library should be responsible for or something that polybar as a user of the library should be responsible for?

I think either is fine. It is totally reasonable for a library to declare that it is meant to be used in a single-threaded context, especially if the library is also single-threaded.

I also believe that it makes sense for simple libraries to not deal with concurrency and let the client handle that, if they need to.
And since dwmipcpp shares memory with its clients, the clients need to synchronize access to that memory anyway, so they can just as well also synchronize calls into the library.

Feel free to reach out with questions about what kind of threads a module runs in.

> I look forward to polishing up the module and helping out with polybar's issues. You guys do seem a bit short-staffed for a project this large.

We are indeed. I look forward to having you help out as well :)

> Also, is the code coverage check something I should be concerned about. I believe none of the other modules had any unit tests created.

Not really. The way they are now, modules are almost impossible to unit test (especially if they use ipc). This will hopefully change in the future, but for now, you can ignore the coverage when working on modules.

include/modules/dwm.hpp
mihirlad55: Just to clarify, the underlying data isn't changed by dwmipcpp. Most of the pointers are to specific elements of `shared_ptr<vector<...>>` class members which are all initialized in the `dwm_module` constructor. The library creates the shared_pointers to these vectors and returns the shared pointers in the method calls. This data doesn't change; what changes is the module's knowledge of the currently active monitor, currently selected layout, etc. which are kept track of using those pointers for simplicity.

`dwmipcpp` does not start any of its own threads and does not synchronize any access.

Hope this clarifies everything.

src/modules/dwm.cpp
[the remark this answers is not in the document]
mihirlad55: See commit 416bb9888ce9387bdef4a0fe46b62a569db062b6

cmake/02-opts.cmake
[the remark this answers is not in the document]
mihirlad55: See commit fb4b38d411cdbb95a39bf47a26a35f57f08c0b5a

include/modules/dwm.hpp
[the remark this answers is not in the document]
mihirlad55: Do you want me to update `.clang-tidy` to override this warning in my PR, or would you rather do it separately?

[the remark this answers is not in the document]
mihirlad55: See 4399ac008c6571f074892aebb4d03fae081dbbfd

src/modules/dwm.cpp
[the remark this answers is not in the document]
mihirlad55: See e0f5249da1d8c2799bbec0ac9de860c815c003fc

src/modules/dwm.cpp
mihirlad55: See a9fbf7cd5ce0fa67582c823cc8207c261b92718c

src/modules/dwm.cpp
mihirlad55: See b24ce89d5d055273ee0637df166db2d775cc2b44

src/modules/dwm.cpp
mihirlad55: See 30052f4549495abd320f1f5bb037320ab3fbfcc4

mihirlad55:
Just resolved a bunch of the issues. Sorry for the spam, lol. I'll start doing some more research and look into the concurrency issues probably tomorrow.

patrick96:
You're quick :smile: 

I will try to review it this week.

> Let me know if you want me to disable modernize-use-trailing-return-type warnings in .clang-tidy, so I can resolve that issue.

Yes, please go ahead and disable it, but in a separate PR if that's possible

mihirlad55:
Lol thanks. I still actually have to look into the concurrency issues. I got held up with some other projects since my last batch of commits.

include/modules/dwm.hpp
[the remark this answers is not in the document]
mihirlad55: See PR #2167 for disabling modernize-use-trailing-return-type

src/modules/dwm.cpp
mihirlad55: See commits 4a5070ce6c8a152be0945afb7d6bde2ed269315f and b1caed225b660db2edc420e03e504af5650d660c

src/modules/dwm.cpp
mihirlad55: See commits 4a5070ce6c8a152be0945afb7d6bde2ed269315f and b1caed225b660db2edc420e03e504af5650d660c

mihirlad55:
Alright. Sorry for the long delay. School just started back up and I've gotten a bit busy. I got some time today to sit down and figure out polybar's threading model and added a mutex to hopefully avoid any concurrency issues. I also added a reconnect scheduling feature that attempts to reconnect to dwm in `has_event` to avoid blocking other threads.

I think I've addressed all the issues you brought up. Let me know if there's anything else you want me to fix. 

For helping triage polybar's issues and future contributions, I just want to give you a head's up that I might get busy with school, so I can't guarantee I'll be very active, but I'll try my best to help out when I find spare time.

src/modules/dwm.cpp
```diff
@@ -0,0 +1,544 @@
...
+    /**
+     * ---------------------------------------------------------------
+     * |             Tag              |                  |           |
+     * |------------------------------| Monitor Selected |   State   |
+     * | Urgent | Selected | Occupied |                  |           |
+     * |--------|----------|----------|------------------|-----------|
+     * |   Yes  |     *    |     *    |         *        |   Urgent  |
+     * |   No   |    Yes   |     *    |        Yes       |  Focused  |
+     * |   No   |    Yes   |     *    |        No        | Unfocused |
+     * |   No   |    No    |    Yes   |         *        |  Visible  |
```
patrick96, line 291: Is this dwm terminology? From what I can tell tags in this state don't necessarily have to be visible (or am I understanding dwm's tag system wrong?).

src/modules/dwm.cpp
```diff
@@ -0,0 +1,544 @@
...
+  bool dwm_module::input(const string& action, const string& data) {
+    m_log.info("%s: Sending workspace %s command to ipc handler", name(), action);
+
+    try {
+      m_ipc->run_command(action, (Json::UInt64)stoul(data));
+      return true;
+    } catch (const dwmipc::SocketClosedError& err) {
+      m_log.err("%s: Disconnected from socket: %s", name(), err.what());
+      sleep(chrono::duration<double>(1));
+      reconnect_dwm();
```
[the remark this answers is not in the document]
patrick96: I think the best way is to reconnect during the next call to `has_event`.

src/modules/dwm.cpp
```diff
@@ -0,0 +1,544 @@
...
+        this->on_monitor_focus_change(ev);
+      };
+      m_ipc->subscribe(dwmipc::Event::MONITOR_FOCUS_CHANGE);
+      m_log.info("%s: Subscribed to monitor focus change", name());
+    } catch (const dwmipc::IPCError& err) {
+      throw module_error(err.what());
+    }
+  }
+
+  void dwm_module::stop() {
```
patrick96, line 150: I'm not sure we need this function. The pointer is automatically reset and the underlying connection closed when the module is deconstructed. Also, it shouldn't be possible to catch an exception here because the destructor for the connection object (or any destructor for that matter) should never throw an exception.

patrick96 (changes requested):
Sorry for the wait.

There have been quite a few changes on the polybar side that introduced merge conflicts. I have added comments in the relevant files.

The module still has some concurrency issues. In one of your comments you said that you added a mutex, but it seems you removed it again.

From what I can see, only two threads operate on the module during normal operations (ignoring starting and stopping the module):

* The controller thread runs all the action handler code and the `build` function.
* Everything else (including the callback functions) is run in the module thread.

Unfortunately, it is quite difficult to get proper synchronization here. If we have a mutex around `m_ipc`, we block any incoming actions while waiting for dwm events. This is a fundamental issue with our current module system and I did write down some thoughts about this in https://github.com/polybar/polybar/issues/2337.

What other modules do to solve this is to not use the same connection object in the action handler. So one thing to consider is to open a connection when an action comes in and let it terminate once the action is processed.

This would remove the concurrency concerns around `m_ipc` and once we get a more robust system in place, we can simplify this again.

Issue #2156, opened 2020-08-02
Add posibility to escape glyphs/emojis


AlexKapustin:
## Is your feature request related to a problem? Please describe.
Currently it's really hard to manage polybar config files which contain UTF-8 glyphs/emojis.
Editors use different font, therefore shows complitely different content. Some configs use fonts which are quite unique and sometimes used content is not visible in editor at all. Or, for example, github may not output that caracter, when you just browse config. So here are lots of confusion.

## Why does polybar need this feature?
I believe it will be much easier to use and build own configs using escaping glyphs.

## Describe the solution you'd like
For example:
intead of:  [  ]
use: [ \uee89af; ]

PR #2157, opened 2020-08-02
#2156 - added support for glyphs

AlexKapustin:


patrick96:
Hey, thanks for implementing this and sorry I didn't come to this sooner.

I wrote up some of my thoughts about escape characters in #2354. I think we should wait until at least the first step is implemented, before we add proper uses for escape characters.

Also, escaping should happen in the config parser and not when labels are loaded.

PR #2163, opened 2020-08-08
feat(xkeyboard): Add variant support to layout label

### LINKED ISSUE #316: Can you add keyboard layout variant?
When I write in French I use Bépo layout (like dvorak for french language)

$ setxkbmap -query                
rules:      evdev
model:      pc105
layout:     fr
variant:    bepo

How can I display "fr (bepo)" in my polybar?

I can see there's 3 variables in the module/xkeyboard.cpp:
%name%
%layout%
%number%

Is it possible to add the variant variable?

Regards

parmort:
Closes #316

Adds `%variant%` variable to `label-layout` option in the xkeyboard module. As
suggested, I took the group name and grabbed the string inside the parens.

### DIFF
--- a/src/modules/xkeyboard.cpp
+++ b/src/modules/xkeyboard.cpp
@@ -121,6 +121,7 @@ namespace modules {
     if (m_layout) {
       m_layout->reset_tokens();
       m_layout->replace_token("%name%", m_keyboard->group_name(m_keyboard->current()));
+      m_layout->replace_token("%variant%", m_keyboard->variant_name(m_keyboard->current()));
 
       auto const current_layout = m_keyboard->layout_name(m_keyboard->current());
       auto icon = m_layout_icons->get(current_layout, DEFAULT_LAYOUT_ICON);
--- a/src/modules/xkeyboard.cpp
+++ b/src/modules/xkeyboard.cpp
@@ -1,11 +1,11 @@
 #include "modules/xkeyboard.hpp"
+
 #include "drawtypes/iconset.hpp"
 #include "drawtypes/label.hpp"
+#include "modules/meta/base.inl"
 #include "utils/factory.hpp"
 #include "x11/atoms.hpp"
 #include "x11/connection.hpp"
-
-#include "modules/meta/base.inl"
 
 POLYBAR_NS
 
@@ -25,8 +25,6 @@ namespace modules {
    */
   xkeyboard_module::xkeyboard_module(const bar_settings& bar, string name_)
       : static_module<xkeyboard_module>(bar, move(name_)), m_connection(connection::make()) {
-
-
     // Setup extension
     // clang-format off
     m_connection.xkb().select_events_checked(XCB_XKB_ID_USE_CORE_KBD,
@@ -81,7 +79,7 @@ namespace modules {
       m_indicator_icons_on = factory_util::shared<iconset>();
 
       auto icon_pair = string_util::tokenize(m_conf.get(name(), DEFAULT_INDICATOR_ICON, ""s), ';');
-      if(icon_pair.size() == 2) {
+      if (icon_pair.size() == 2) {
         m_indicator_icons_off->add(DEFAULT_INDICATOR_ICON, factory_util::shared<label>(icon_pair[0]));
         m_indicator_icons_on->add(DEFAULT_INDICATOR_ICON, factory_util::shared<label>(icon_pair[1]));
       } else {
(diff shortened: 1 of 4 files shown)


src/modules/xkeyboard.cpp
```diff
@@ -141,9 +140,9 @@ namespace modules {
         }
 
         auto indicator_on = m_keyboard->on(it);
-        auto &indicator_labels = indicator_on ? m_indicator_on_labels : m_indicator_off_labels;
-        auto &indicator_icons = indicator_on ? m_indicator_icons_on : m_indicator_icons_off;
-        auto &indicator_state = indicator_on ? m_indicator_state_on : m_indicator_state_off;
+        auto& indicator_labels = indicator_on ? m_indicator_on_labels : m_indicator_off_labels;
+        auto& indicator_icons = indicator_on ? m_indicator_icons_on : m_indicator_icons_off;
+        auto& indicator_state = indicator_on ? m_indicator_state_on : m_indicator_state_off;
```
kronn, line 145: This seems strange. I do not have a lot of experience in C++, it this syntactically correct? (yes, I am looking to learn something here, I honestly do not know)

parmort: For the changes to these lines? It was the formatter you run when submitting a PR. The lines themselves use the ternary operator, which is a shortened if-statement. It reads `if ? then : else`.

So these lines ask whether the indicator is on; if it is, they display the "on" icons and labels, otherwise they display the "off" icons and labels.

For simple conditional checks, they seem to be preferred because its very concise. It's pretty common in c-style languages. I'm a beginner with cpp as well, so I'm not sure what the `auto &` is. My best guess is that it becomes the type of the object first assigned to it, and it gets the address of that object? Don't quote me on that.

kronn: The ternary was the part I understood. Maybe I should have mentioned that :-) 

thanks for researching the `auto&`, I did a short search for this as well earlier (before reading your comment). The type-inference I have seen before, just the "inference and store as reference"-syntax was new. 

Thank you for teaching me (even if by accident and automatic code-formatting) :-)

src/x11/extensions/xkb.cpp
```diff
@@ -64,6 +64,16 @@ const string keyboard::layout_name(size_t index) const {
...
 }
 
+/**
+ * Get current variant name
+ */
+const string keyboard::variant_name(size_t index) const {
+  string group_name = this->group_name(index);
+  int start = group_name.find('(') + 1;
+  int num_chars = group_name.find(')') - start;
+  return group_name.substr(start, num_chars);
```
kronn, line 74: could you add the string that you expect to find as `group_name` to the comment above? I guess that a real unit-test is hard, but a comment about the expectations would be helpful to understand this method.

src/x11/extensions/xkb.cpp
```diff
@@ -64,6 +64,18 @@ const string keyboard::layout_name(size_t index) const {
...
+/**
+ * Get current variant name
+ * "GROUP (VARIANT)"
+ *         ^^^^^^^
+ */
+const string keyboard::variant_name(size_t index) const {
+  string group_name = this->group_name(index);
+  int start = group_name.find('(') + 1;
+  int num_chars = group_name.find(')') - start;
+  return group_name.substr(start, num_chars);
```
patrick96, line 76: If there are no parentheses, then this just returns the entire group name. I think we should either return nothing or something like `N/A` in that case.

patrick96 (changes requested):
Thanks for contributing and thanks @kronn for reviewing :)

I did some more digging into this and there really doesn't seem to be a direct way to get the variant with xcb. There is a way using Xlib and one using `values.symbolsName` but neither is that straightforward and there also isn't really much documentation about the data format for these two approaches.

I think this way is works well enough.

Can you also add this to the `CHANGELOG.md` (https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog)

parmort:
Once (or if) this is merged, I will update the wiki with the new tag.

patrick96:
Code looks good now. Please also add your changes to the changelog: https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog

You may need to rebase first because when you started this branch the changelog didn't exist yet.

patrick96 (approved):
Perfect! Thank you very much :tada:

parmort:
It's not released. I assume it will be released in 3.6; 3.5.5 is a patch release. If you want it, you'll have to build from master.

Huy-Ngo:
Ah thanks, I assumed wrongly.

PR #2165, opened 2020-08-09
feat(xworkspaces): Add support for showing number of windows in a workspace

### LINKED ISSUE #604: number of open windows in a specific workspace
Is possible to add the number of open windows per workspace  in the xworkspace module?

I want to do something like this
istantanea_2017-06-02_19-09-44

lepisma:
This adds support for an `%nwin%` token in `xworkspaces` that shows the number of windows. This is related to #604 

Here is the relevant config:
```
[module/ewmh]
type = internal/xworkspaces

enable-click = true
enable-scroll = false

icon-0 = TXT;
icon-1 = WEB;
icon-2 = EXT;
icon-3 = FLT;
icon-default = 

label-active = "  %icon% %name% [%nwin%] "
label-active-foreground = ${colors.foreground}
label-active-background = #882E3440
label-active-underline = ${colors.accent-light}

label-occupied = "  %icon% %name% [%nwin%] "
label-occupied-foreground = ${colors.foreground-alt}

label-empty = "  %icon% %name% [%nwin%] "
label-empty-foreground = ${colors.foreground-alt}
```

A screenshot showing the above config in action.

image

Closes #604


src/x11/ewmh.cpp
```diff
@@ -84,6 +84,14 @@ namespace ewmh_util {
     return desktops;
   }
 
+  unsigned int get_number_of_windows(unsigned int desktop) {
+    unsigned int count(0);
+    for (auto& win : get_client_list()) {
+      if (get_desktop_from_window(win) == desktop) count++;
```
patrick96, line 90: Please always use curly bracktes.

src/x11/ewmh.cpp
```diff
@@ -84,6 +84,14 @@ namespace ewmh_util {
     return desktops;
   }
 
+  unsigned int get_number_of_windows(unsigned int desktop) {
```
patrick96, line 87: I think getting the number of windows per workspace is possible in a single iteration over the client list instead of doing it once per workspace.

We already iterate over the client list in `rebuild_clientlist`. I would suggest using that loop to also build a mapping from workspace to number of windows, similarly to `m_clients`. When replacing the tokens you would just need to read from the map.

patrick96:
Closing due to inactivity

PR #2166, opened 2020-08-11
fix(config_parser): Gracefully handle BOM

### LINKED ISSUE #2075: Key name '﻿<feff>;' is empty or contains forbidden characters.
## Describe the issue
I built polybar right after forking and cloning the repo. Now it doesn't start and in the log says [read below].

**Expected behavior:**
Polybar should start.

**Actual behavior:**
Polybar doesn't start.

**Was it working before?**
I downloaded, some days ago, the latest release (with sha256sum: `4d22c977969a561f561fdc7a609073854d8fea8a9eec6941e12a80457edcb63a`) and building with that it was working.
Also with the AUR stable version (the one I use daily) it was working.

## To Reproduce
Clone the repo with `git clone --recursive https://github.com/polybar/polybar`, and build with:
```bash
mkdir build
cd build
cmake ..
make -j$(nproc)
# Optional. This will install the polybar executable in /usr/local/bin
sudo make install
```
(from the wiki)

## Polybar Log
```
---
polybar|notice:  Parsing config file: /home/koraynilay/.config/polybar/config
polybar|error: /home/koraynilay/.config/polybar/config:1: Key name '<feff>﻿;' is empty or contains forbidden characters.
---
```

## Environment:
* WM: i3
* Distro: Arch Linux
* Output of `polybar -vvv`: 

```
polybar 3.4.0-96-gfc42da8-dev

Build type: Release
Compiler: /usr/bin/c++
Compiler flags:  -Wall -Wextra -Wpedantic -O3 -DNDEBUG
Linker flags: 
```

## Additional context
I did a bit of research, and 0xFEFF is the [Byte Order Mark](https://en.wikipedia.org/wiki/Byte_order_mark). So there could be a problem while reading the encoding.

parmort:
Closes #2075

Add a check to handle the byte order mark.

### DIFF
--- a/src/components/config_parser.cpp
+++ b/src/components/config_parser.cpp
@@ -1,7 +1,7 @@
+#include "components/config_parser.hpp"
+
 #include <algorithm>
 #include <fstream>
-
-#include "components/config_parser.hpp"
 
 POLYBAR_NS
 
@@ -193,12 +193,16 @@ line_type config_parser::get_line_type(const string& line) {
     case '#':
       return line_type::COMMENT;
 
-    default:
+    default: {
+      if (string_util::contains(line, "﻿"))
+        throw syntax_error("The file encoding is not supported. Please use UTF-8");
+
       if (string_util::contains(line, "=")) {
         return line_type::KEY;
       } else {
         return line_type::UNKNOWN;
       }
+    }
   }
 }
 
--- a/src/components/config_parser.cpp
+++ b/src/components/config_parser.cpp
@@ -151,6 +151,11 @@ void config_parser::parse_file(const string& file, file_list path) {
 }
 
 line_t config_parser::parse_line(const string& line) {
+  if (string_util::contains(line, "\ufeff")) {
+    throw syntax_error(
+        "This config file uses UTF-8 with BOM, which is not supported. Please use plain UTF-8 without BOM.");
+  }
+
   string line_trimmed = string_util::trim(line, isspace);
   line_type type = get_line_type(line_trimmed);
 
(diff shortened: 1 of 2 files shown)


src/components/config_parser.cpp
```diff
@@ -193,12 +193,16 @@ line_type config_parser::get_line_type(const string& line) {
     case '#':
       return line_type::COMMENT;
 
-    default:
+    default: {
+      if (string_util::contains(line, "﻿"))
```
patrick96, line 197: This will not detect the BOM in header or comment lines. A better place to put this check would be at the beginning of the `parse_line` method.

Please also don't put a literal BOM here because this looks like an empty string on github and probably many editors (BOM inside the text signifies a zero-width nonbreaking space). It's probably best to put a string constant that is constructed from a `char` array in the header file.

src/components/config_parser.cpp
```diff
@@ -193,12 +193,16 @@ line_type config_parser::get_line_type(const string& line) {
...
+        throw syntax_error("The file encoding is not supported. Please use UTF-8");
```
patrick96, line 198: I think we should make this more specific. Since we only test for the UTF-8 encoding of the BOM, something like `This config file uses UTF-8 with BOM, which is not supported. Please use plain UTF-8 without BOM.` would make it clearer.

src/components/config_parser.cpp
parmort: I added the unit test and moved the check in dd65b76. I kept getting this redefined error when I tried to extract the BOM reference to a string constant in the `config_parser.hpp` file, I assume that somewhere the file is being loaded twice.

src/components/config_parser.cpp
parmort: Fixed in dd65b76

PR #2195, opened 2020-10-06
refactor(config): Throw missing section error

patrick96:
Makes error messages less confusing if for some reason a key in an
non-existent section is requested.


patrick96:
I had to remove the `section_error` exception class again because in many places we catch `key_error` for things like fallback values and the `section_error` isn't caught by that. It also doesn't really make sense to have a standalone section error since that is also just a key error (the key cannot be found)

PR #2196, opened 2020-10-07
Add support for keyword include-directory

### LINKED ISSUE #1946: include multiple files with wildcard
We only can declare static path, this is not modular.

Make configuration more modular.

It would be great to create file per module in a ``modules.d`` and add in config
```
include-files = modules.d/*
```

GuilhermeJSilva:
Adds `include-directory` keyword which receives as an argument a path to a directory. This keyword allows the importing of all regular files included in the given directory.
```
include-directory = /path/to/directory
```
Closes #1946

### DIFF
--- a/src/components/config_parser.cpp
+++ b/src/components/config_parser.cpp
@@ -144,6 +144,13 @@ void config_parser::parse_file(const string& file, file_list path) {
 
     if (!line.is_header && line.key == "include-file") {
       parse_file(file_util::expand(line.value), path);
+    } else if (!line.is_header && line.key == "include-directory") {
+      const string expanded_path = file_util::expand(line.value);
+      vector<string> file_list = file_util::list_files(expanded_path);
+      sort(file_list.begin(), file_list.end());
+      for (const auto& filename : file_list) {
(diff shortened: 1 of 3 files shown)


src/components/config_parser.cpp
```diff
@@ -144,6 +144,11 @@ void config_parser::parse_file(const string& file, file_list path) {
 
     if (!line.is_header && line.key == "include-file") {
       parse_file(file_util::expand(line.value), path);
+    } else if (!line.is_header && line.key == "include-directory") {
+        const string expanded_path = file_util::expand(line.value);
+        for (const auto& filename : file_util::list_files(expanded_path)) {
```
patrick96, line 149: We should sort the list here. This gives us a consistent include order and will also prevent issues when one of the files does not with a section header.

patrick96 (changes requested):
The general approach looks good. Just had some minor suggestions.

Thanks :smiley:

patrick96 (approved):
Looks great now.

Congrats on your first contribution! :beers:

PR #2197, opened 2020-10-07
Temperature module, ramp drawstyle uses wrong icon for temperatures near base and warn temperatures

DRKblade:
## Bug Description

**Expected behavior**
In the Temperature module [documentation](https://github.com/polybar/polybar/wiki/Module:-temperature), the ramp settings says:

> The icon selection will range from `base-temperature` to `warn-temperature`,
> temperatures above `warn-temperature` will use the last icon
> and temperatures below `base-temperature` will use `ramp-0`

This leads the user to think that the last icon is _only_ used for temperatures above warn-temperature, and likewise ramp-0 for temperatures below base-temperature.

**Actual behavior**

The the last icon is actually used for temperatures slightly below warn-temperature, and ramp-0 for temperatures slightly above base-temperature. This disrupts the style of the module ([example](https://imgur.com/a/IX9kKIQ)).

## Reproduction

Using this module and font-awesome:

```
[module/temp0]
type = internal/temperature
thermal-zone = 0
base-temperature = 40
warn-temperature = 80

format = <ramp><label>
format-warn = <ramp><label-warn>

label = "%temperature-c%"
label-foreground = ${color.temp-shade}
label-background = ${color.bg}
label-padding= 1

label-warn = "%temperature-c%"
label-warn-foreground = ${color.fg-alt}
label-warn-background = ${color.temp-shade}
label-warn-padding = 2
ramp-0 = 
ramp-1 = 
ramp-2 = 
ramp-3 = 
ramp-4 = 
ramp-5 = 
ramp-6 = 
ramp-padding-left = 1
ramp-foreground = ${color.temp-shade}
ramp-background = ${color.bg}
ramp-6-foreground = ${color.fg-alt}
ramp-6-background = ${color.temp-shade}
```

Which should produce the [above](https://imgur.com/a/IX9kKIQ) output, we can see that the icon in ramp-6 is used for 71C, which is below the warn-temperature of 74C.
This causes the ramp and label nodes to have different styles, making the whole thing ugly.

## What I did to fix it

I made the temperature module calculate the variable `m_perc` as an unbounded-percentage instead of a normal percentage, this will allow temperatures below and above the normal range be reflected in the percentage value.

I wrote a ramp function `get_by_percentage_with_border` which will use the first and last ramp icons exclusively for percentage values outside the range (0, 100) and used it for the ramp in temperature instead

### DIFF
--- a/src/drawtypes/ramp.cpp
+++ b/src/drawtypes/ramp.cpp
@@ -24,7 +24,7 @@ namespace drawtypes {
   }
 
   ramp::operator bool() {
-    return m_icons.size() >= 2;
+    return !m_icons.empty();
   }
 
   /**
--- a/src/drawtypes/ramp.cpp
+++ b/src/drawtypes/ramp.cpp
@@ -19,8 +19,16 @@ namespace drawtypes {
   }
 
   label_t ramp::get_by_percentage_with_borders(float percentage) {
-    size_t index = percentage * (m_icons.size() - 2) / 100.0f + 1;
-    return m_icons[math_util::cap<size_t>(index, 0, m_icons.size() - 1)];
+    size_t index;
+    if (percentage <= 0.0f) {
+      index = 0;
+    } else if (percentage >= 100.0f) {
+      index = m_icons.size() - 1;
+    } else {
+      size_t index = percentage * (m_icons.size() - 2) / 100.0f + 1;
+      index = math_util::cap<size_t>(index, 0, m_icons.size() - 1);
+    }
+    return m_icons[index];
   }
 
   ramp::operator bool() {
--- a/src/drawtypes/ramp.cpp
+++ b/src/drawtypes/ramp.cpp
@@ -25,7 +25,7 @@ namespace drawtypes {
     } else if (percentage >= 100.0f) {
       index = m_icons.size() - 1;
     } else {
-      size_t index = percentage * (m_icons.size() - 2) / 100.0f + 1;
+      index = percentage * (m_icons.size() - 2) / 100.0f + 1;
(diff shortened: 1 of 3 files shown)


DRKblade:
The previous commit was commited by mistake, I'm trying to revert it.

I'm sorry for the clumsiness, this is my first time making a pull request.

src/drawtypes/ramp.cpp
```diff
@@ -18,6 +18,11 @@ namespace drawtypes {
     return m_icons[math_util::cap<size_t>(index, 0, m_icons.size() - 1)];
   }
 
+  label_t ramp::get_by_percentage_with_borders(float percentage) {
+    size_t index = percentage * (m_icons.size() - 2) / 100.0f + 1;
```
patrick96, line 22: For small negative percentages (e.g. -1%), `index` will be 1 because the float is truncated towards 0. I'd suggest explicitly checking if the percentage `>= 100` or `<= 0` and returning `m_icons.size() -1` and `0` respectively.

patrick96 (changes requested):
The current behavior was always the intended behavior, namely that `ramp-0` to `ramp-N` are distributed evenly between the base and warn temperatures and anything outside of that range is also represented by either `ramp-0` or `ramp-N`, the documentation was just a bit ambiguous.

But I agree with you that it makes more sense that `ramp-0` and `ramp-N` represent only the edge values (and anything outside the range).

DRKblade:
Thanks for the insight!
I made the requested change.

patrick96 (approved):
Looks good now. Thanks :)

Issue #956, opened 2018-01-15
Conditional Display of Battery


boydthomson:
My goals are to keep indicators hidden when they're not important.  I don't need to know how much time is remaining on my battery when it's full.  I'd like to only display the battery module when I've got < X time/percentage remaining.
  
I'm able so do something similar with the warn feature in the temperature module, but I can't find anything similar for battery.  This is how I do my cpu temp.  "label =" is blank but label-warn.  Similar 'warn' feature doesn't seem to be present in module/battery.

[module/temperature]
type = internal/temperature
thermal-zone = 0
warn-temperature = 68

format = <ramp> <label>
format-underline = ${colors.background}
format-warn = <ramp> <label-warn>
format-warn-underline = ${self.format-underline}

label =
label-warn = %temperature%
label-warn-foreground = ${colors.secondary}

patrick96:
You can probably use `ramp-capacity` to accomplish this:
```dosini
format-discharging = <ramp-capacity> 
helper-label = My label
ramp-capacity-0 = ${self.helper-label}
ramp-capacity-1 = ${self.helper-label}
ramp-capacity-2 = ${self.helper-label}
ramp-capacity-3 = ${self.helper-label}
ramp-capacity-4 = 
ramp-capacity-5 = 
ramp-capacity-6 = 
ramp-capacity-7 = 
ramp-capacity-8 = 
ramp-capacity-9 = 
```
With this, when discharging, the module should be empty until your battery goes below 40%. Of course this is not perfect, because you will have to set the colors yourself with formatting tags.

boydthomson:
Thanks for the suggestion.  In theory, this is exactly what I want.  The
issue is that <ramp-capacity> doesn't handle variables, so using %time% or
%percentage% in <ramp-capacity-X> or in <helper-label> displays the
variable name "%time%" instead of "4:56".  Another workaround might be to
make the module width zero for the higher ramp-capacity elements, but I
haven't been able to get that working either.

Boyd Thomson  founder / yogi / CEO / janitor
t – 604.351.6810 | stretchvancouver.com

180 E Pender St. Vancouver B.C. V6A 1T3

patrick96:
Yes, I see. The best solution is probably to just implement `label-warn`, I'll add it to the list.

patrick96:
Implemented in #2199

Issue #570, opened 2017-05-09
label-warn for other modules


noctuid:
I think it would make sense to be able to have warning thresholds for the memory, cpu, and battery modules as well.

jaagr:
You could already achieve the same behavior by using `<ramp>` if the module have support for it. For example:

~~~ dosini
[module/cpu]
type = internal/cpu
format = <ramp-load>

ramp-load-0 = low
ramp-load-0-foreground = #0f0
ramp-load-1 = medium
ramp-load-1-foreground = #ff0
ramp-load-2 = high
ramp-load-2-foreground = #f00

# or

ramp-load-0 = normal
ramp-load-1 = normal
ramp-load-2 = normal
ramp-load-3 = warning

noctuid:
I tried that initially, but it doesn't seem possible to use `%percentage%` outside of the label. Also the memory module doesn't have `<ramp>`, and it would be more precise and less verbose to be able to use a specific value for a threshold.

boydthomson:
noctuid, did you find a way to achieve this?  I'm looking for the same feature.  Would be great to have greater visual distraction when my RAM usage is getting up there.

juliancoffee:
It is a great idea. I implemented it for memory in #2142, and want to implement it for CPU as well.

derrgill:
Love this idea, I also would like to be able to set arbitrary warning thresholds (e.g. RAM and swap percentages, even network download/upload speeds, and perhaps eventually disk read/write speeds if a disk I/O module is implemented) and thus conditional label warn foreground colors. Pretty much everything in my polybar is just colored an unobtrusive white and only changes color IF something actually changes, e.g., I have a custom script module that only changes to red if my ISP has changed my WAN ipv4 and/or ipv6 address(es).

DRKblade:
I'm working on this feature. It has been implemented succesfully on the memory, cpu and filesystem modules, only the battery module left and I will upload the code.

But I was wondering, do you need a warning on the alsa/pulseaudio module? Maybe for protection from loud sounds? If so, I will implement this feature on those modules too.

patrick96:
Implemented in #2199

Issue #1871, opened 2019-09-14
filesystem: warn mode below threshold


petrmanek:
## Is your feature request related to a problem? Please describe.
Nope. Just switched to polybar and miss this particular feature.

## Why does polybar need this feature?
Currently, filesystem modules have only 2 modes: _mounted_ or _unmounted_. This is useful in some cases (e.g. network or USB storage mounts), but doesn't capture a more frequent issue when a mount runs out of free space. Other bars support some variants of conditioned appearance in such cases. It would be cool if Polybar caught up.

## Describe the solution you'd like
I propose to replicate the scheme used in the _temperature_ module:

 1. add a new _warn_ mode to allow appearance customization when the free space is low,
 2. add a new _warn-space_ attribute to control when the _warn_ mode is activated; its value could either be a 0-1 float (percentage of the total capacity) or a value in bytes, if the value is negative (which could be the default for compatibility reasons), the _warn_ mode would be permanently disabled

## Describe alternatives you've considered
The same can be done with a custom shell script, but that in my opinion introduces too much overhead. It would be nice to have the same thing along with the low memory footprint of Polybar.

## Additional context
N/A

patrick96:
We will consider this when we rework the module state machines.

patrick96:
This has now been implemented in #2199

PR #2199, opened 2020-10-09
Warn state for CPU, Memory, Filesystem and Battery modules

DRKblade:
This is an [item](https://github.com/polybar/polybar/projects/7#card-25459423) in the todo list from the [Module State Machines project](https://github.com/polybar/polybar/projects/7), and was requested in #570 and #956. The idea is that the modules that monitor system resources will display different labels and animations when that resource requires attention (battery low, high temperature, high cpu/memory usage).

The warn state would work like that of the temperature module. The condition to activate the warn states are as follow:
* Memory module: percentage of memory used >= `warn-percentage`
* CPU module: total percentage of all cores >= `warn-percentage`
* Filesystem module: the filesystem is mounted and percentage of used space >= `warn-percentage`
* Battery module: the charge percentage >= `low-at`

**Release Notes:**

This adds `format-warn` and `label-warn` to the memory, cpu, fs, and battery module.

It also slightly changes how the ramps in those modules behave. The ramps are bounded by these warn thresholds and, the same as the temperature module, the first and last ramp elements are responsible for everything `<=` the lower bound or `>=` the upper bound. Before, the ramp generally went from 0 to 100 and the first and last ramp elements covered the same share inside the range as all other ramp elements as well as everything outside the range.

These two changes in behavior partially cancel each other out, so the change in behavior in existing configs will be barely noticeable.

It also adds the following:

* battery: `animation-low`, `low-at = 10`
* cpu: `warn-percentage = 80`
* fs: `warn-percentage = 90`
* memory: `warn-percentage = 90`

**/Release Notes**

Example configuration:
```
[module/cpu]
type = internal/cpu

interval = 1
warn-percentage = 80

format = <label>
format-prefix = CPU
format-background = #000
format-foreground = #fff
format-padding = 1
label = " %percentage%%"

format-warn = <label-warn>
format-warn-prefix = CPU
format-warn-background = #fff
format-warn-foreground = #000
format-warn-padding = 2
label-warn = " %percentage%%"
```


DRKblade:
This would also provide the feature requested by #1871.

src/modules/cpu.cpp
```diff
@@ -17,15 +17,22 @@ namespace modules {
 
   cpu_module::cpu_module(const bar_settings& bar, string name_) : timer_module<cpu_module>(bar, move(name_)) {
     m_interval = m_conf.get<decltype(m_interval)>(name(), "interval", 1s);
-
+    m_totalwarn = m_conf.get(name(), "warn-percentage", 80);
```
patrick96, line 20:
```suggestion
    m_totalwarn = m_conf.get(name(), "warn-percentage", m_totalwarn);
```

src/modules/fs.cpp
```diff
@@ -104,6 +110,31 @@ namespace modules {
...
+	if (m_labelmounted) {
+	  replace_tokens(m_labelmounted);
+	}
+	if (m_labelwarn) {
+	  replace_tokens(m_labelwarn);
+	}
+	if (m_labelunmounted) {
+          m_labelunmounted->reset_tokens();
+          m_labelunmounted->replace_token("%mountpoint%", mount->mountpoint);
+	}
```
patrick96, line 137: You have some weird indentations here.

tests/unit_tests/drawtypes/ramp.cpp
```diff
@@ -0,0 +1,23 @@
...
+  r.add(factory_util::shared<label>("test1", 0));
+  r.add(factory_util::shared<label>("test2", 0));
+  r.add(factory_util::shared<label>("test3", 0));
+  EXPECT_EQ("test1", r.get_by_percentage(33)->get());
+  EXPECT_EQ("test2", r.get_by_percentage(34)->get());
+  EXPECT_EQ("test3", r.get_by_percentage(67)->get());
+  EXPECT_EQ("test1", r.get_by_percentage_with_borders(19, 20, 40)->get());
+  EXPECT_EQ("test2", r.get_by_percentage_with_borders(21, 20, 40)->get());
+  EXPECT_EQ("test2", r.get_by_percentage_with_borders(39, 20, 40)->get());
+  EXPECT_EQ("test3", r.get_by_percentage_with_borders(41, 20, 40)->get());
```
patrick96, line 19: Please also add the following tests:

```cpp
  EXPECT_EQ("test1", r.get_by_percentage_with_borders(20, 20, 40)->get());
  EXPECT_EQ("test3", r.get_by_percentage_with_borders(40, 20, 40)->get());
```

We should test these edge cases as well.

patrick96 (changes requested):
Code looks good. Thanks!

I do have some concerns about how this changes existing behavior. Because the thresholds for the new warn states would change how existing configs would display once we are in the warn state.

For example right now, if your CPU is over 80%, it would display whatever is in `format` and the labels and ramps used there. If we used the same config in this PR, it would now display just the total load percentage.

We generally want people to be able to update without it changing how their bar looks like (at least not in a substantial way). So we can't really have these warning states enabled by default, or we must make sure that they behave exactly the same as the non-warn states by default.

I can think of two solutions to this:

1. Set the threshold to something that can never be reached. In the battery module, we would set it to anything smaller than `0`. For memory, cpu, and fs, we would have to set it to something larger than 100. It has to be strictly smaller than 0 or larger than 100 because both 0% battery and 100% cpu, ram, or storage can be reached and the warn states are activated *at* those values.
2. Have an extra boolean flag in each module that is set to true only if the new warn formats were defined in the module config. Only use the warn states if the flag is true.

None of these solutions are pretty unfortunately. Can you think of another way of ensuring that the module doesn't suddenly display something completely different in the warn states?

---

And just to note, because it also changes module behavior: This PR adds the same changes to the behavior of the first and last ramp elements for the battery, fs, cpu, and memory module as #2197 added for the temperature module.

While this can be considered a breaking change, it just changes the ranges that ramp elements are responsible for a bit. So, the same as for the temperature module, this is probably fine, we just need to make sure to mention it in the changelogs.

src/modules/fs.cpp
DRKblade: I tried to fix this, but my editor may insert tabs instead of spaces. Tell me if it's still a problem after my commit

DRKblade:
We can create a new fallback tag. The default value for FORMAT_WARN will be this tag, and when `module::get_format()` is called, we check if FORMAT_WARN have this tag to return DEFAULT_FORMAT instead.

DRKblade:
Another idea on my mind is that I can write `module_formatter::try_get()` which will add a format only if it is defined. If we load the warn format using `try_get`, I can then write `module_formatter::has_format()` to check if the warn format exist and use the default format when it isn't.

patrick96:
> Another idea on my mind is that I can write module_formatter::try_get() which will add a format only if it is defined. If we load the warn format using try_get, I can then write module_formatter::has_format() to check if the warn format exist and use the default format when it isn't.

I think that is a good idea. :+1:

DRKblade:
I applied it on all relevant modules.

src/modules/fs.cpp
```diff
@@ -141,7 +172,11 @@ namespace modules {
    * Select format based on fs state
    */
   string fs_module::get_format() const {
-    return m_mounts[m_index]->mounted ? FORMAT_MOUNTED : FORMAT_UNMOUNTED;
+    if (!m_mounts[m_index]->mounted)
+      return FORMAT_UNMOUNTED;
```
patrick96, line 176: Please add curly braces around all if statements.

src/modules/meta/base.cpp
```diff
@@ -138,6 +137,15 @@ namespace modules {
 
     m_formats.insert(make_pair(move(name), move(format)));
   }
+  	
+  void module_formatter::add(string name, string fallback, vector<string>&& tags, vector<string>&& whitelist) {
+    add_value(move(name), m_conf.get(m_modname, move(name), move(fallback)), forward<vector<string>>(tags), forward<vector<string>>(whitelist));
+  }
+  	
+  void module_formatter::add(string name, vector<string>&& tags, vector<string>&& whitelist) {
+   	if (m_conf.has(m_modname, name))
```
patrick96, line 146: Please also add curly braces here. And you seem to have tab indentation here.

src/modules/meta/base.cpp
```diff
@@ -138,6 +137,15 @@ namespace modules {
 
     m_formats.insert(make_pair(move(name), move(format)));
   }
+  	
+  void module_formatter::add(string name, string fallback, vector<string>&& tags, vector<string>&& whitelist) {
+    add_value(move(name), m_conf.get(m_modname, move(name), move(fallback)), forward<vector<string>>(tags), forward<vector<string>>(whitelist));
+  }
+  	
+  void module_formatter::add(string name, vector<string>&& tags, vector<string>&& whitelist) {
```
patrick96, line 145: We should probably name this `add_optional` to make it explicit what it is for.

patrick96 (changes requested):
Looks good. Just some minor style changes.

Please also rebase and resolve the merge conflicts.

DRKblade:
Ok, is it good?

src/modules/fs.cpp
```diff
@@ -104,6 +110,31 @@ namespace modules {
...
+	if (m_labelmounted) {
+	  replace_tokens(m_labelmounted);
+	}
+	if (m_labelwarn) {
+	  replace_tokens(m_labelwarn);
+	}
+	if (m_labelunmounted) {
+    m_labelunmounted->reset_tokens();
+    m_labelunmounted->replace_token("%mountpoint%", mount->mountpoint);
+	}
```
patrick96, line 137: Please indent with spaces

src/modules/meta/base.cpp
```diff
@@ -140,6 +138,16 @@ namespace modules {
...
   }
+  	
+  void module_formatter::add(string name, string fallback, vector<string>&& tags, vector<string>&& whitelist) {
+    add_value(move(name), m_conf.get(m_modname, move(name), move(fallback)), forward<vector<string>>(tags), forward<vector<string>>(whitelist));
+  }
+  	
+  void module_formatter::add_optional(string name, vector<string>&& tags, vector<string>&& whitelist) {
+   	if (m_conf.has(m_modname, name)) {
+      add_value(move(name), m_conf.get(m_modname, move(name)), move(tags), move(whitelist));
+   	}
```
patrick96, line 149: Please reindent with spaces

src/modules/meta/base.cpp
```diff
@@ -86,13 +86,12 @@ namespace modules {
   // }}}
   // module_formatter {{{
 
-  void module_formatter::add(string name, string fallback, vector<string>&& tags, vector<string>&& whitelist) {
-    const auto formatdef = [&](const string& param, const auto& fallback) {
-      return m_conf.get("settings", "format-" + param, fallback);
-    };
+	void module_formatter::add_value(string&& name, string&& value, vector<string>&& tags, vector<string>&& whitelist) {
```
patrick96, line 89: Please reindent with spaces

src/modules/fs.cpp
```diff
@@ -141,7 +172,12 @@ namespace modules {
    * Select format based on fs state
    */
   string fs_module::get_format() const {
-    return m_mounts[m_index]->mounted ? FORMAT_MOUNTED : FORMAT_UNMOUNTED;
+    if (!m_mounts[m_index]->mounted)
+      return FORMAT_UNMOUNTED;
+    if (m_mounts[m_index]->percentage_used >= m_perc_used_warn && m_formatter->has(FORMAT_WARN)) {
```
patrick96, line 177:
```suggestion
    if (m_mounts[m_index]->percentage_used >= m_perc_used_warn && m_formatter->has_format(FORMAT_WARN)) {
```

patrick96 (changes requested):
Functionality wise, everything works now. There was a small bug in the fs module, but I added a comment for that.

The build currently breaks, I have added comments for all the places that need to be fixed.

There are also some places where you use tabs instead of spaces, I have also added comments for that.

After these changes, I think we can merge this.

src/modules/fs.cpp
```diff
@@ -28,18 +28,24 @@ namespace modules {
...
     m_spacing = m_conf.get(name(), "spacing", m_spacing);
     m_interval = m_conf.get<decltype(m_interval)>(name(), "interval", 30s);
 
     // Add formats and elements
     m_formatter->add(
         FORMAT_MOUNTED, TAG_LABEL_MOUNTED, {TAG_LABEL_MOUNTED, TAG_BAR_FREE, TAG_BAR_USED, TAG_RAMP_CAPACITY});
-    m_formatter->add(FORMAT_UNMOUNTED, TAG_LABEL_UNMOUNTED, {TAG_LABEL_UNMOUNTED});
+    m_formatter->add(
+        FORMAT_WARN, {TAG_LABEL_WARN, TAG_BAR_FREE, TAG_BAR_USED, TAG_RAMP_CAPACITY});
+    m_formatter->add_optional(FORMAT_UNMOUNTED, TAG_LABEL_UNMOUNTED, {TAG_LABEL_UNMOUNTED});
```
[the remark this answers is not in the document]
DRKblade: Oops, I thought I have committed that change before pushing.

DRKblade:
I think I finally got the indents right.

patrick96 (approved):
Everything looks good now :tada:

Thanks a lot for all your efforts! :smiley:

PR #2211, opened 2020-10-14
Display binary unit prefixes in memory module

### LINKED ISSUE #2023: Memory values are displayed in binary units, while being indicated as decimal
_This issue is more of a minor nitpick, but I think it should either be fixed or documented._

## Describe the issue
The `memory` module appears to be reading all its data from `/proc/meminfo`, which, despite the `kB` in its output, [lists all values as kibibyte](https://unix.stackexchange.com/questions/199482/does-proc-pid-status-always-use-kb) (1024 bytes). Polybar correctly converts this to mebi- and gibibytes by dividing by 1024 and displays the values as such. However, it indicates the units as "MB" and "GB", which is ambiguous. They should be "MiB" and "GiB" to be unambiguous ([IEC 80000-13](https://en.wikipedia.org/wiki/Binary_prefix)).

**Expected behavior:**
Polybar uses the correct unit indicators (binary units) for the memory values it displays.

**Actual behavior:**
Polybar uses ambiguous unit indicators (decimal units) for the memory values it displays.

## Additional context
https://github.com/polybar/polybar/blob/f02fb6702013033ae0d09c3cd37895c5155ca421/src/utils/string.cpp#L280-L292

_The calculation is correct, but the unit indicators ("MB", "GB") are not. The comments and variable names also all reference decimal units (`kbytes`, `mb`, `gb`, ...) despite referring to binary units._

fardragon:
Closes #2023

### DIFF
--- a/include/utils/string.hpp
+++ b/include/utils/string.hpp
@@ -94,8 +94,8 @@ namespace string_util {
   size_t find_nth(const string& haystack, size_t pos, const string& needle, size_t nth);
 
   string floating_point(double value, size_t precision, bool fixed = false, const string& locale = "");
-  string filesize_mb(unsigned long long kbytes, size_t precision = 0, const string& locale = "");
-  string filesize_gb(unsigned long long kbytes, size_t precision = 0, const string& locale = "");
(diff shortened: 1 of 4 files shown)


patrick96 (approved):
LGTM. Thanks!

PR #2214, opened 2020-10-16
Use an option to customize how space is inserted in the bars

DRKblade:
# The problem
Polybar currently use the space character as the unit of margin/padding. But the size of space characters is dependent of the font used in `font-0` option, which means an update of all the space sizes is needed after changing the font.

## Some unsatisfying solution
* A solution is to use a font for margin/padding space and another font for normal text, but the text font have to be set for every module used as there is no global font setting yet.
* Another solution is the `separator` option, but it only affect the margin between modules and not the padding.

# My solution
I decide to create `space-unit` as an option as it allows for a more consistent character to be used as the unit of space (such as the em/en space). Also with this option, the user can scale all the margin/padding in the bar by changing this option.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: bionic
+dist: focal
 language: cpp
 
 env:
@@ -21,7 +21,7 @@ addons:
         - libxcb-image0-dev
         - libxcb-randr0-dev
         - libxcb-util0-dev
-        - python-xcbgen
+        - python3-xcbgen
         - xcb-proto
       - &optional_deps
         - libxcb-xkb-dev
--- a/include/components/types.hpp
+++ b/include/components/types.hpp
@@ -166,6 +166,7 @@ struct bar_settings {
   struct radius radius {};
   int spacing{0};
   label_t separator{};
+  string space_unit{};
 
   string wmname{};
   string locale{};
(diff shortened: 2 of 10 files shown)


patrick96:
I don't see how the space unit solves your original problem. The regular space, em space, and en space all dependent on the font and/or font size, so if you change your font(size), you will still change all your space sizes.

DRKblade:
In the case even the em space and en space aren't consistent. The user can still change the space size by changing the number of space characters in `space-unit` instead of having to change all the margin and padding fields every time they change the font.

But with `space-unit`, you can use the em space, a character that doesn't appear in normal texts, as the unit, then use an icon font that provides em space but not normal text characters as `font-0` and finally the font for normal text as `font-1`. That way, the font and font size of normal texts can be changed without changing the spacing.

patrick96:
I see your point.

Has the size of the spacing been a problem for you when you change your text font? I would have thought that having the space size tied to your text font would be a good thing because it was designed to look natural with the rest of the text.

There was already some work done to allow users to specify padding and margin as other things than just spaces (pixels and points). I think that would address your problem as well: #1671

DRKblade:
I do encounter problems with spacing when changing fonts, as shown [here](https://imgur.com/a/c2rlhZc).
But using other units of space would be the better solution to my own, so I'm looking forward to that feature. In the meantime though, I will keep using this as a temporary solution.

patrick96:
Closing in favor of #1671

PR #2215, opened 2020-10-16
Bug: Battery time left show wrong value when formatted as %M instead of %H:%M

DRKblade:
# The problem
This battery module shows the wrong number of minutes left of the battery when the actual number is greater than 60
```
[module/battery]
type = internal/battery
battery = BAT1
adapter = ADP1
full-at = 100
time-format = %H:%M
format-discharging = <label-discharging>
label-discharging = "   %time% left   "
format-discharging-foreground = #fff
format-discharging-background = #000
```
[Here](https://imgur.com/a/S9GFgc3) is the module rendered with different `time-format` options, at roughly the same battery level.

# Solution
Looking into the code, it turns out that the number of hours is calculated and deducted from the time even though not used in the final string. This affects the number of minutes that was later calculated.

I fixed the bug by checking for the hour token in the format string before calculate and deduct the number of hour.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: bionic
+dist: focal
 language: cpp
 
 env:
@@ -21,7 +21,7 @@ addons:
         - libxcb-image0-dev
         - libxcb-randr0-dev
         - libxcb-util0-dev
-        - python-xcbgen
+        - python3-xcbgen
         - xcb-proto
       - &optional_deps
         - libxcb-xkb-dev
--- a/include/drawtypes/ramp.hpp
+++ b/include/drawtypes/ramp.hpp
@@ -16,6 +16,7 @@ namespace drawtypes {
     void add(label_t&& icon);
     label_t get(size_t index);
     label_t get_by_percentage(float percentage);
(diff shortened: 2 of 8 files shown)


patrick96:
Please open an issue first before you start fixing something that you perceive as a bug.

This isn't a bug. I understand that it might be intuitive that `%M` gives you the total minutes if you don't specify `%H` but with regards to `strftime` (and most other date formatting functions), `%M` has to be between 0 and 59 and really is just the number of minutes since the last full hour.
In this case, setting `tm_min` to anything outside the range [0,59] violates the C standard and will give you an unspecified output.

DRKblade:
Ah, I see the problem with `strftime`, but this can be misleading for the user. Also `strftime` was intended to be used to format date-time, not the amount of time like in this case where only hour, minute and second values are used, and we can simply replace the %H, %M and %S tokens with the respective values.

patrick96:
I don't think it is that simple. We would also need to support things like escaping the `%` character, flags for the conversion specifiers (`0`, `_`, `-`, ...), as well as minimal widths and padding. If we do that, we can just as well provide tokens in the labels since those should (at least in the future) support similar features.

Issue #1296, opened 2018-06-23
Support additional color definition formats


BVollmerhaus:
Currently Polybar only supports color definitions in RGB hex triplets. It would be nice to have additional, more human-readable formats such has HSL or RGB. Especially HSL would allow for easier fine-tuning of colors.

patrick96:
Good idea, I'm currently rewriting the config parser. I may do a reworking of different value types after that. 
If you want to use hsl right now, you could use a converter like [this one](http://serennu.com/colour/hsltorgb.php).

Sircular:
Additionally, it would be nice if it were possible to define colors based on other colors. I use pywal, which changes the xresources colors, and it would be nice to be able to say "20% darker than color4" for example.

patrick96:
@Sircular once we have a clean implementation of different color definitions, I think we could more or less easily implement this.

NBonaparte:
Hopefully no one will request the CIELAB color space...

Issue #682, opened 2017-08-01
Label Gradient


mlamberti:
Hey, thanks for the awesome work!

Using the CPU/Memory modules with gradient bars I noticed that the bar's foreground color overflows to the next module, unless it's foreground color is explicitly defined.

This gave me the idea that it would be nice to have a label-gradient, just like the bar gradient, but coloring the label.

While my CPU is idle there will be written "CPU 0%" in white, and when the load rises it get's red following a defined gradient. By this one could have pretty small indicators, just using a color instead of bars and numbers.

patrick96:
The thing with the overflowing colors is a known issue and is discussed in #666 and #639 (tl;dr use the latest release)

I think the label gradient is a great idea, I didn't even know that the progressbar could do that.
However I don't have time at the moment to implement new features, and I think jaagr is quite busy as well. If you want to take a stab at it, I'd say go for it, unfortunately I cannot guarantee that it will be merged.

patrick96:
I think the ability to use tokens inside the ramp (#2223) would make this achievable.

patrick96:
Closing in favor of #2223

PR #2218, opened 2020-10-18
Add support for HSL color space

quandangv:
This feature is requested in #1296.

HSL color can now be specified by using the format `hsl(hue, sat, lum)` or `hsla(hue, sat, lum, alpha)` in place of the normal color code.

The conversion from RGB to HSL, while not required, was also implemented in case of future need. I also added some unit tests to cover the conversion code.

### DIFF
--- a/include/utils/color.hpp
+++ b/include/utils/color.hpp
@@ -1,9 +1,11 @@
 #pragma once
 
 #include <cstdlib>
+#include <regex>
 
 #include "common.hpp"
 #include "utils/cache.hpp"
+#include "utils/math.hpp"
 
 POLYBAR_NS
 
@@ -114,6 +116,8 @@ namespace color_util {
   }
 }
 
+          
+    
 struct rgb {
   double r;
   double g;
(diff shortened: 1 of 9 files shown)


include/utils/color.hpp
```diff
@@ -162,4 +166,89 @@ struct rgba {
...
+      h *= 60;
+    }
+
+    return hsl(h, s, l);
+  }
+
+
+  static inline string convert_hsl(string color) {
+    if (color.rfind("hsla") == 0) {
+      std::regex rgx(R"(hsla\( *([0-9\.]+), *([0-9\.]+), *([0-9\.]+), *([0-9\.]+) *\))");
```
patrick96, line 235: We should probably try to match the entire string, so that `hsl(0, 0, 0).....` is not valid.

include/utils/color.hpp
```diff
@@ -162,4 +166,89 @@ struct rgba {
...
+    return rgb(r, g, b);
+  }
+
+  rgba to_rgba(double a) {
+    double r, g, b;
+    get_rgb(r, g, b);
+    return rgba(r, g, b, a);
+  }
+
+  static hsl from_rgb(rgb color) {
```
patrick96, line 211: I think we are never going to need this direction of the conversion. Because once hsl(a) values are parsed from the config, they should be in the argb format and should never be converted back to hsl.

patrick96 (changes requested):
Code looks good. I am gonna trust you on the color conversion code though :wink: 

We just recently merged a very large refactor that cleaned up the color system a lot: #1920 
This unfortunately means that you have a lot of conflicts in this PR now.

The upside is that it should be quite straightforward to adapt your changes to the new system. The only thing you need to do is rewrite your HSL(A) parser to produce an `rgba` instance (32-bit integer) instead of a string and put that into the `convert` function that returns an `rgba` instance in `config.cpp`.

We are not even going to need hsl class because internally, polybar now just moves 32-bit integers around.

DRKblade:
The thing is, while #1920 was in the work, I went on and wrote a separate color conversion tool [here](https://github.com/DRKblade/cspace). Not only it can convert from HSL to RGB, it also supports other color spaces such as HSV, CIELab and RGB(A), and color operations like modifying brightness, saturation and hue.
I could just modify the existing code to make it work for HSL, but I think it might be better to use this opportunity to add all those features to polybar

patrick96:
I'm not sure we need support for all these different color spaces, it's just added complexity for something almost nobody will use.

The color modification functions sound interesting though. But I am a bit wary of the complexity here since this will basically add an entire color modification language to polybar.

DRKblade:
I actually intended to use the tool as an external dependency since I have already uploaded it to the [AUR](https://aur.archlinux.org/packages/cspace/) and I intend to upload it to other repositories too. So this may actually reduce the size of the added code since we won't have to include the conversion functions in polybar.

Color modification can be implemented in config::dereference, i.e. `${color: section.key; brightness*1.5}` would mean _"use the same color in section.key, but with 1.5 times the brightness"_. The `brightness*1.5` part would be handled by the tool.

Also, if we want to implement any kind of gradient in the future (such as requested in #682), we may want to do it in CIELab since gradients in RGB don't look good (as shown [here](https://imgur.com/a/6yFwves), which is taken from [this question](https://stackoverflow.com/questions/22607043/color-gradient-algorithm)).

patrick96:
I don't think a new external dependency is a good idea. Especially one that doesn't exist in the official repos of most distros.
Because polybar will likely no longer be packaged natively in some distros if it means they have to add other third-party packages to the distro as well.

Right now polybar can be compiled in most distros without having to manually compile third-party software and I very much would like to keep it that way.

The original usecase behind #682 is probably better supported by #2223.

I just think we shouldn't go overboard with adding features here without knowing whether they will be even used.

I very much like your initiative and conceptually, having access to a vast library of color manipulation functionality, sounds quite cool.
But there is always a cost involved, in this case a maintenance cost on our side and an implementation cost on your side.

Before we continue going down this path of adding all this color modification code, we should first try to figure out whether this is something that people actually find useful. I will share a post on our subreddit and in our messaging channels to ask people to tell us how they use colors and what abilities they are missing.

For now, I think we should focus on the hsl conversion in `config::convert` and color modification code can be implemented independently from this if/when we go in this direction in `dereference` where the `${color:...}` reference could just load the given color as an `rgba` color and our conversion function would kick in.

Thank you very much for your efforts here :smiley:

patrick96:
I just posted on [reddit](https://www.reddit.com/r/Polybar/comments/khgfum/how_do_you_use_colors_in_polybar_and_what_kind_of/). Let's see what the users think.

DRKblade:
Some users did ask for color "brightness" and "manipulation", that's something Cspace would provide. I suggest we use it as an internal library instead of as a dependency.

patrick96:
I am very conflicted about this.

Adding a new dependency, whether as a submodule or a proper dependency, puts polybar on the hook for making sure nothing breaks with that dependency. While I do believe that you will be very responsive when it comes to issues with cspace, I don't know how the situation will look like in, say, two years. If you ever decide that you no longer want to maintain cspace (and it is totally understandable that you at some point lack the interest or time to do so), polybar will have an unmaintained dependency and we will have to fork and maintain it, like we already do with xpp and i3ipcpp.

I very much appreciate all the work you are doing here, but I don't think we should add cspace to polybar.
I also did not see enough enthusiasm about these color modifications that it would warrant adding an extra dependency for it.

PR #2219, opened 2020-10-18
fix(net): Float interval for network speed

patrick96:
Before the time difference between two measurements was always an
integer number, so for intervals < 1, you would always get 0 and for any
other non-integer interval you would get skewed results.

### DIFF
--- a/src/adapters/net.cpp
+++ b/src/adapters/net.cpp
@@ -1,7 +1,5 @@
 #include "adapters/net.hpp"
 
-#include <iomanip>
-
 #include <arpa/inet.h>
 #include <linux/ethtool.h>
 #include <linux/if_link.h>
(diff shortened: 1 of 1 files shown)


MicroDroid:
I am the original reporter of this bug, and I can confirm this PR solves the problem.

MicroDroid:
Update: `interval` config no longer works as expected though, setting it to 0.25 does not make it update 4 times a second.

patrick96:
Thanks for reviewing :)

> Update: interval config no longer works as expected though, setting it to 0.25 does not make it update 4 times a second.

This PR doesn't change anything about how the module updates (this is managed by `timer_module.hpp`). Intervals <= 0.5s are not really accurate. This is because we try to synchronize module updates to the interval (e.g a 60 second interval will update the module on the full minute) and we had to do some adjusting in #2123.

PR #2220, opened 2020-10-19
Use KiB/MiB/GiB instead of KB/MB/GB

MicroDroid:


patrick96:
Why should we prefer KiB over KB?

MicroDroid:
@patrick96 Because it's technically more correct and more relevant to computers than SI units. It is also the correct number one would get by dividing the Mbit/s rating by 8 to get MiB/s

I assume polybar users are technical enough for this. 

[this](https://ux.stackexchange.com/questions/13815/files-size-units-kib-vs-kb-vs-kb) post is relevant and interesting.

The difference is minor, but the more correct display is better and should be standardized anyways.

vaygr:
It could be a setting to turn on/off such behavior if the user prefers so, but generally kibibytes etc. are [never used](https://en.wikipedia.org/wiki/List_of_interface_bit_rates) in storage / networking, so I don't think it should ever be a default.

MicroDroid:
@vaygr Windows does use KB even though it refers to KiB: https://superuser.com/a/938433

However, Mac *did* use KiB but it recently switched to KB: https://superuser.com/a/493887

Thus yes, a configurable option would be a better option.

PR #2240, opened 2020-11-10
Respect interval default for tail = true

vaygr:
[Module:-script](https://github.com/polybar/polybar/wiki/Module:-script#basic-settings) states that with `tail = true` interval value defaults to 0.

I was wondering why my module wasn't working as expected to realize that this condition is not taken into account.

This proposed patch fixes it.


patrick96 (approved):
LGTM

Thanks!

PR #2256, opened 2020-12-01
Add some precision in network module for MB and GB

cyanpencil:
Add 1 decimal floating precision when net speed is MB, and 2 decimals when speed is in GB.

If you want I can add a config var to enable this, but my bar looked ugly switching from 999 KB to 1 MB. Right now it switchs to 999 KB to 1.0MB, much better imho.

### DIFF
--- a/src/adapters/net.cpp
+++ b/src/adapters/net.cpp
@@ -231,13 +231,15 @@ namespace net {
 
     vector<string> suffixes{"G", "M"};
     string suffix{"K"};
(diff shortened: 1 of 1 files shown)


patrick96:
This does exactly the same thing as #2054 (which I kind of forgot about). I went ahead and merged the other PR.

Thanks for contributing though :)

cyanpencil:
Thanks!

PR #2282, opened 2020-12-10
Use null-byte string as default workspace icon instead of empty string

dylanhiemstra:

* [ ] Refactor
* [ ] Feature
* [x] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
This PR fixes showing empty space when not defining an icon for a workspace. Best shown with an example:

Before:

before

Afer:

after

Note, you do need to remove the ; for unused workspace icons:

```
ws-icon-0 = 1;
ws-icon-1 = 2
ws-icon-2 = 3
ws-icon-3 = 4
ws-icon-4 = 5
ws-icon-5 = 6
ws-icon-6 = 7
ws-icon-7 = 8
ws-icon-8 = 9;
```

## Related Issues & Documents
Small fix, didn't create issue.

### DIFF
--- a/src/modules/i3.cpp
+++ b/src/modules/i3.cpp
@@ -55,7 +55,7 @@ namespace modules {
     m_labelseparator = load_optional_label(m_conf, name(), "label-separator", "");
 
     m_icons = factory_util::shared<iconset>();
-    m_icons->add(DEFAULT_WS_ICON, factory_util::shared<label>(m_conf.get(name(), DEFAULT_WS_ICON, ""s)));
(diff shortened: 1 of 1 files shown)


patrick96:
I'm not sure what you are describing is a bug. The extra space comes from the fact that you probably set the label to
```
%name% %icon%
```

And since `%icon%` is empty, you will get that space between `%name%` and `%icon%`.

I am not exactly sure why using the null-byte removes that space.

If you don't want that space and want icons for only some workspaces, you could set the label to `%name%%icon%` and add the space to the icons: `ws-icon-0 = 1; `

Also, you don't need to add an icon definition for everyworkspace, just

```
ws-icon-0 = 1;
ws-icon-1 = 9;
```

is enough.

Issue #2292, opened 2020-12-12
xkeyboard module adds spaces between the indicators (capslock, etc), even if the indicator is empty, and after layout when all indicators are empty.


Filip62:
## Checklist

* [x] I have read the appropriate section in the [contributing
  guidelines](https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md)
* [x] I believe this issue is a problem with polybar itself and not a misconfiguration on my part.
* [x] I have searched for other open and closed [issues](https://github.com/polybar/polybar/issues?q=is%3Aissue) that
  may have already reported this problem.
* [x] I have checked the [known issues](https://github.com/polybar/polybar/wiki/Known-Issues) page for this problem.

## Describe the bug

There is an additional space/margin after the xkeyboard module even though in the config i explicitly set the margin to 0, seems like the xkeyboard module adds spaces between the indicators (capslock, etc), even if the indicator is empty.

**Expected behavior:**

No margins added by the xkeyboard module unless it should because of an active indicator.

**Actual behavior:**

The xkeyboard module adds a space/margin after the layout indicator if if CapsLock is not active.

**Was it working before?**
* Did you also experience this bug in an earlier version of polybar (yes/no/don't know)?
Do not know.
* If no, what was the last version where this worked correctly?
Do not know.

## To Reproduce

```dosini
[bar/bspwm2]
monitor = DVI-I-0
width = 100%
height = 27
;offset-x = 1%
;offset-y = 1%
; radius = 6.0
fixed-center = false

background = ${colors.background}
foreground = ${colors.foreground}

line-size = 3
line-color = #f00

border-size = 0
border-color = #00000000

padding-left = 0
padding-right = 0

module-margin-left = 2
module-margin-right = 2

; font-0 = JetBrainsMono Nerd Font:pixelsize=10;1
font-0 = FreeSans:antialias:pixelsize=11;1
font-1 = Font Awesome 5 Free,Font Awesome 5 Free Solid:style=Solid:pixelsize=10;1

modules-left = bspwm
modules-center = xwindow
modules-right = pulseaudio-control memory cpu temperature xkeyboard date powermenu2

wm-restack = bspwm

;override-redirect = true

scroll-up = bspwm-desknext
scroll-down = bspwm-deskprev

cursor-click = pointer
cursor-scroll = ns-resize

[module/xkeyboard]
type = internal/xkeyboard
blacklist-0 = num lock

format-prefix = "%{T2} "
[...]

PR #2306, opened 2020-12-15
Fix issue #2292

Filip62:

* [ ] Refactor
* [ ] Feature
* [x] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Related Issues & Documents

Fixes [#2292](https://github.com/polybar/polybar/issues/2292)


patrick96 (changes requested):
Thanks for looking into this.

I think this solution only works if all of the indicators are empty (`enabled == 0`)

For example, imagine `m_indicators` had three elements and only the second element was empty, in that case `build` would return true but the output would be:

```
label1__label2
```

(Each underscore represents a call to `builder->space`).
Here, returning false would also not fix this because only trailing spaces are removed by the `get_output` function in `base.inl`.

This is a tricky problem. One solution I could think of is to first iterate over all indicators and add all non-empty labels to a separate vector. We can then iterate over that vector and always add spacing between labels because we know they're not empty.

I think there are also ways to do this in a single iteration but they will probably be a lot more complex than what I described because you have to deal with situations like two consecutive labels being empty.

Also, generally the check for a label being empty is `!label || !*label`, since we know `indicator.second` actually is a valid pointer, you can just use `!*indicator.second` to check if a label is empty.

Filip62:
Actually I think it works.

My initial problem was that a trailing space was added between `<label-layout>` and `<label-indicators>` because even though the indicator was empty, the method `xkeyboard::build` returned `true`, therefore the trailing space was not removed.

Now if at least one of the indicators is not empty only then will `xkeyboard::build` return true, so the space makes sense in that scenario.

I think the calls to `builder->space` work just as expected.
`label-indicator-margin = 1`




2alt

In the current master branch there would be a trailing space at the end of every screenshot.

src/modules/xkeyboard.cpp
```diff
@@ -190,13 +190,17 @@ namespace modules {
       builder->node(m_layout);
     } else if (tag == TAG_LABEL_INDICATOR && !m_indicators.empty()) {
       size_t n{0};
+      size_t enabled{0};
       for (auto&& indicator : m_indicators) {
+        if (!indicator.second) {
```
patrick96, line 195:
```suggestion
        if (!*indicator.second) {
```

src/modules/xkeyboard.cpp
```diff
@@ -190,13 +190,17 @@ namespace modules {
...
+          enabled++;
+        }
         if (n++) {
           builder->space(m_formatter->get(DEFAULT_FORMAT)->spacing);
         }
         builder->node(indicator.second);
       }
-      return n > 0;
+      return enabled > 0 && n > 0;
```
patrick96, line 203:
```suggestion
      return enabled > 0;
```

`n` is always greater or equal to `enabled`

patrick96 (changes requested):
I think there was a misunderstading on my part as to what went wrong in your issue.

The extra space(s) after `<label-layout>` can come from one of two places (or both):

* `format-spacing`
* The fact that the format is set to `format = <label-layout> <label-indicator>`

I originally thought your issue was caused by the first, but it isn't because `format-spacing` defaults to 0, so no spacing is added in your case.

You are actually experiencing the second one and you are completely correct that the `build` function returned `true` even though no indicators were present and it should have returned false.

If `format-spacing` is set to anything other than 0, you still get extra spacing. If you want to, you can of course also tackle that, but I think we can just merge the fix for the issue with the wrong return value.

Could you please also add this fix to the `CHANGELOG.md` file. We have just today started keeping a changelog as we merge stuff, there is a new section about in [`CONTRIBUTING.md`](https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog)

Also `!indicator.second` will always return `false` because this only checks if the pointer actually exists and not if the underlying label is empty, that's why you need `*indicator.second`. Right now the `build` function always returns false because `enabled` is never incremented.
That's also why you *never* get any space after the module even though you set `label-indicator-margin = 1`.
margin is always supposed to be added to both sides of a label (or format) and should never be trimmed.

Filip62:
@patrick96 I think I fixed both issues.
From what I tested, everything worked as I would expect.
I hope it works because the solution feels pretty nice with my lame skills. :)

patrick96 (approved):
Oh wow. I was sure that the solution must be more complex, but this does indeed work. Good job!

PR #2320, opened 2020-12-20
feat: Module Visibility

### LINKED ISSUE #2108: Toggle visibility of a module

I don't want, in my case, the mpd module to show unless I'm wanting to listen to music

It'd be nice to have.  There's got to be other scenarios when you might want to do this.  Like, maybe in case you just want a quick glance at memory usage or cpu usage or something and you don't want it in your face 100% of the time.

Be able to send a polybar-msg to it from, in my case, i3 via a key binding.

You could even maybe create a new module _container_ thing, which you'd list the modules you want grouped together.

```
# Create the following containers for me
[container/systemrelated]
# anything after contains are modules you have defined
contains = cpu memory temp
# Default visibility state - true
visible = false
```
or

```
# Define the containers you want
containers = systemrelated anothercontainer yetanothercontainer

# Then, inside of a module, you could assign it to a container
[module/cpu]
type = custom/script

... other cpu settings

container = systemrelated

... other cpu settings

[module/memory]
type = custom/script

... other memory settings

container = systemrelated

... other memory settings

```
Now, in my case i3, I can setup a keybinding which sends a **cmd** to polybar, which would hide or show the modules in the specified _container_.  If it doesn't find the specified _container_, it'd default to looking for a single module with that name.

None

## Additional context
None

parmort:

* [ ] Refactor
* [X] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update
* [ ] Other

## Description
Creates new IPC commands to toggle the visibility of a specific module. End goal (as expressed in #2108) is to stop processing hidden modules, although I will probably complete an initial implementation to just hide the modules.

```sh
$ polybar-msg [-p PID] cmd hide.mymodule # Hides module mymodule
$ polybar-msg [-p PID] cmd show.mymodule # Shows module mymodule
$ polybar-msg [-p PID] cmd toggle.mymodule # Toggles visibility of mymodule
```

## Related Issues & Documents
Closes #2108

### DIFF
--- a/include/modules/meta/base.inl
+++ b/include/modules/meta/base.inl
@@ -62,7 +62,7 @@ namespace modules {
   }
 
   template <typename Impl>
-  bool module<Impl>::visible(bool value) {
+  bool module<Impl>::set_visible(bool value) {
     m_visible = value;
     return m_visible;
   }
--- a/include/modules/meta/base.inl
+++ b/include/modules/meta/base.inl
@@ -62,6 +62,12 @@ namespace modules {
   }
 
   template <typename Impl>
+  bool module<Impl>::visible(bool value) {
+    m_visible = value;
+    return m_visible;
+  }
+
+  template <typename Impl>
   void module<Impl>::stop() {
     if (!static_cast<bool>(m_enabled)) {
       return;
--- a/include/modules/meta/base.inl
+++ b/include/modules/meta/base.inl
@@ -65,6 +65,7 @@ namespace modules {
   void module<Impl>::set_visible(bool value) {
     m_log.info("%s: Visibility changed (state=%s)", m_name, value ? "shown" : "hidden");
     m_visible = value;
+    broadcast();
   }
 
   template <typename Impl>
--- a/include/modules/meta/base.inl
+++ b/include/modules/meta/base.inl
@@ -62,10 +62,9 @@ namespace modules {
   }
 
   template <typename Impl>
-  bool module<Impl>::set_visible(bool value) {
+  void module<Impl>::set_visible(bool value) {
     m_log.info("%s: Visibility changed (state=%s)", m_name, value ? "shown" : "hidden");
     m_visible = value;
-    return m_visible;
   }
 
   template <typename Impl>
--- a/include/modules/meta/base.inl
+++ b/include/modules/meta/base.inl
@@ -63,6 +63,7 @@ namespace modules {
 
   template <typename Impl>
   bool module<Impl>::set_visible(bool value) {
+    m_log.info("%s: Visibility changed (state=%s)", m_name, value ? "shown" : "hidden");
     m_visible = value;
     return m_visible;
   }
--- a/include/modules/meta/base.inl
+++ b/include/modules/meta/base.inl
@@ -57,6 +57,11 @@ namespace modules {
   }
 
   template <typename Impl>
+  bool module<Impl>::visible() const {
+    return static_cast<bool>(m_visible);
+  }
+
+  template <typename Impl>
   void module<Impl>::stop() {
     if (!static_cast<bool>(m_enabled)) {
       return;
--- a/include/modules/unsupported.hpp
+++ b/include/modules/unsupported.hpp
@@ -30,9 +30,7 @@ namespace modules {
     bool visible() const {                                                              \
       return false;                                                                     \
     }                                                                                   \
-    void set_visible(bool value) {                                                      \
-      (void)value;                                                                      \
-    }                                                                                   \
+    void set_visible(bool) {}                                                           \
     void start() {}                                                                     \
     void stop() {}                                                                      \
     void halt(string) {}                                                                \
--- a/include/modules/unsupported.hpp
+++ b/include/modules/unsupported.hpp
@@ -27,6 +27,12 @@ namespace modules {
     bool running() const {                                                              \
       return false;                                                                     \
     }                                                                                   \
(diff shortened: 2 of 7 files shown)


parmort:
### Todo list
- [ ] Create a new IPC command
- [ ] Have the IPC command receiver (probably the controller?) emit a `module_visibility_change` signal
- [ ] Implement a signal receiver function in the base module to grab the signal and hide itself

patrick96:
Approach looks good. Having the visibility information inside the module should make it easier in the future to actually pause/unpause modules when they are invisible.

Btw. for all methods that you add to `module_interface`, you also need to add them to `unsupported.hpp`.

patrick96 (changes requested):
Good job!

Now that I see your implementation, I think it would be cleaner from a user perspective to be able to change module visibility through an action instead of a command, because then users could easily change module visibility through click actions instead of having to go through `polybar-msg` (which also targets all bars).

But we are not going to do this here, because there is no way for all modules to share an action without adding it to every module separately. I have some ideas about action routing (see [this comment](https://github.com/polybar/polybar/pull/1907#issuecomment-731858072)) which could make this easier in the future.

But it's still a good idea to merge this.

Could you also add the name of the ipc commands to the changelog?

parmort:
Would adding an additional config parameter to modules to determine default visibility be an idea? This would allow modules to be hidden by default.
```ini
[module/mymodule]
visible=true ; defaults true
; OR
hidden=false ; defaults false
```

This would be useful for, as an example, an email notification module which only appears when there are new/unread notifications.

> But we are not going to do this here, because there is no way for all modules to share an action without adding it to every module separately. I have some ideas about action routing (see [this comment](https://github.com/polybar/polybar/pull/1907#issuecomment-731858072)) which could make this easier in the future.

This isn't the right place to have this discussion, but would overloads to the `input` function be an idea to fix this, similar to the signal system? Conservatively adding different types of actions could allow the base module to define default functionality for certain actions, like hiding and showing modules.

src/components/controller.cpp
```diff
@@ -530,6 +530,25 @@ bool controller::forward_action(const actions_util::action& action_triple) {
...
+      continue;
+
+    if (visible == 0) {
+      mod->set_visible(false);
+    } else if (visible == 1) {
+      mod->set_visible(true);
+    } else if (visible == 2) {
+      mod->set_visible(!mod->visible());
+    }
+
```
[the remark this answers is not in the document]
parmort: I'm gonna elect to use `broadcast` because that minimizes the calls to the signal emitter, which would be (slightly) helpful if removing the signal system would be an eventuality.

patrick96:
> This isn't the right place to have this discussion, but would overloads to the input function be an idea to fix this, similar to the signal system?

We already overload the `input` function, so yeah this could work, modules would just have to start calling the super function in their `input` methods.
I have already implemented a proof of concept of my action router idea in #2336. This makes the action code inside modules much cleaner (because they no longer have to differentiate between the different actions themselves). It would also make implementing this a lot cleaner.

include/modules/unsupported.hpp
```diff
@@ -27,6 +27,12 @@ namespace modules {
     bool running() const {                                                              \
       return false;                                                                     \
     }                                                                                   \
+    bool visible() const {                                                              \
+      return false;                                                                     \
+    }                                                                                   \
+    void set_visible(bool value) {                                                      \
+      (void)value;                                                                      \
+    }                                                                                   \
```
patrick96, line 35:
```suggestion
    void set_visible(bool) {                                                      \
    }                                                                                   \
```

include/modules/meta/base.inl
```diff
@@ -56,6 +56,18 @@ namespace modules {
...
+  template <typename Impl>
+  bool module<Impl>::visible() const {
+    return static_cast<bool>(m_visible);
+  }
+
+  template <typename Impl>
+  void module<Impl>::set_visible(bool value) {
+    m_log.info("%s: Visibility changed (state=%s)", m_name, value ? "shown" : "hidden");
+    m_visible = value;
+    broadcast();
```
patrick96, line 68: :+1:

patrick96 (changes requested):
Code looks solid. I just recently moved the `config.cmake` file into its own folder and that caused a merge conflict you need to resolve before we can merge this.

patrick96 (approved):
I went ahead and rebased against master so we can merge this.

I think having a module setting for the visible state is a good idea. Do you want to implement this in a new PR?

parmort:
Yeah, I'll get working on that.

patrick96:
Awesome! Thanks :smiley:

patrick96:
I have now more or less finished my action router idea (#2336) and adding universal actions like the visibility change actions seems quite easy. The only downside is that we can't use `toggle` as the action name for toggling the module visibility because that name is already used for the date module's click action. We would have to come up with a different name, maybe `toggle_visible`.

parmort:
Or we could change the date action to `toggle-format` or something. That would break configs though, so some sort of notification would be necessary, like a pinned issue or a note at the top of the readme.

patrick96:
The action names are the closest thing we have to a public API and I am not very keen on breaking it.

I think people can live with `toggle-visible` :wink:

patrick96:
People seem to like this feature ;)

https://www.reddit.com/r/Polybar/comments/kssoga/the_new_showhide_module_feature_is_amazing/

PR #2333, opened 2020-12-25
Reduce overhead of mpd module

jpalus:

* [ ] Refactor
* [ ] Feature
* [ ] Bug Fix
* [X] Optimization
* [ ] Documentation Update

## Description

By default mpd module causes a lot of wakups and results in quite some network traffic even when mpd is not playing anything. This in turn results in unneeded battery use. This PR introduces proper use of idle/recv_idle along with polling on mpd socket to make sure overhead of mpd module is minimal.

### DIFF
--- a/include/adapters/mpd.hpp
+++ b/include/adapters/mpd.hpp
@@ -93,6 +93,7 @@ namespace mpd {
     void idle();
     bool noidle();
     int recv_idle();
+    int try_recv_idle(int timeout);
 
     unique_ptr<mpdstatus> get_status();
     unique_ptr<mpdstatus> get_status_safe();
--- a/include/adapters/mpd.hpp
+++ b/include/adapters/mpd.hpp
@@ -91,7 +91,8 @@ namespace mpd {
 
     int get_fd();
     void idle();
-    int noidle();
+    bool noidle();
+    int recv_idle();
 
     unique_ptr<mpdstatus> get_status();
     unique_ptr<mpdstatus> get_status_safe();
--- a/include/modules/mpd.hpp
+++ b/include/modules/mpd.hpp
@@ -99,7 +99,7 @@ namespace modules {
     chrono::system_clock::time_point m_lastsync{};
     float m_synctime{1.0f};
 
-    int m_quick_attempts{0};
+    int m_connect_attempts{0};
 
     // This flag is used to let thru a broadcast once every time
     // the connection state changes
--- a/src/adapters/mpd.cpp
+++ b/src/adapters/mpd.cpp
@@ -201,12 +201,21 @@ namespace mpd {
     }
   }
 
-  int mpdconnection::noidle() {
+  bool mpdconnection::noidle() {
+    check_connection(m_connection.get());
+    bool success = true;
+    if (m_idle) {
+      success = mpd_send_noidle(m_connection.get());
+    }
+    return success;
+  }
+
+  int mpdconnection::recv_idle() {
     check_connection(m_connection.get());
     int flags = 0;
-    if (m_idle && mpd_send_noidle(m_connection.get())) {
-      m_idle = false;
+    if (m_idle) {
       flags = mpd_recv_idle(m_connection.get(), true);
+      m_idle = false;
       mpd_response_finish(m_connection.get());
       check_errors(m_connection.get());
     }
--- a/src/adapters/mpd.cpp
+++ b/src/adapters/mpd.cpp
@@ -1,3 +1,5 @@
+#include <poll.h>
+
 #include <cassert>
 #include <csignal>
 #include <thread>
@@ -220,6 +222,20 @@ namespace mpd {
       check_errors(m_connection.get());
     }
     return flags;
+  }
+
+  int mpdconnection::try_recv_idle(int timeout) {
+    struct pollfd pfd = {m_fd, POLLIN, 0};
+
+    int poll_ret = poll(&pfd, 1, timeout);
+
+    if (poll_ret > 0) {
+      return recv_idle();
+    } else if (poll_ret == 0) {
+      return 0;
+    } else {
+      throw mpd_exception("poll() returned error: "s + std::strerror(errno));
+    }
   }
 
   unique_ptr<mpdstatus> mpdconnection::get_status() {
--- a/src/adapters/mpd.cpp
+++ b/src/adapters/mpd.cpp
@@ -197,7 +197,7 @@ namespace mpd {
   void mpdconnection::idle() {
     check_connection(m_connection.get());
     if (!m_idle) {
-      mpd_send_idle(m_connection.get());
+      mpd_send_idle_mask(m_connection.get(), (mpd_idle)(MPD_IDLE_PLAYER | MPD_IDLE_OPTIONS));
       check_errors(m_connection.get());
       m_idle = true;
     }
--- a/src/modules/mpd.cpp
+++ b/src/modules/mpd.cpp
@@ -152,10 +152,10 @@ namespace modules {
 
   void mpd_module::idle() {
     if (connected()) {
-      m_quick_attempts = 0;
+      m_connect_attempts = 0;
       sleep(80ms);
     } else {
-      sleep(m_quick_attempts++ < 5 ? 0.5s : 2s);
+      sleep(++m_connect_attempts < 10 ? m_connect_attempts * 0.5s : 5s);
     }
   }
 
--- a/src/modules/mpd.cpp
+++ b/src/modules/mpd.cpp
@@ -160,9 +160,6 @@ namespace modules {
   }
 
   bool mpd_module::has_event() {
-    bool def =
-        (m_statebroadcasted == (connected() ? mpd::connection_state::DISCONNECTED : mpd::connection_state::CONNECTED));
-
     try {
       if (!m_mpd) {
         m_mpd = factory_util::unique<mpdconnection>(m_log, m_host, m_port, m_pass);
@@ -173,11 +170,11 @@ namespace modules {
     } catch (const mpd_exception& err) {
       m_log.err("%s: %s", name(), err.what());
       m_mpd.reset();
-      return def;
+      return true;
     }
 
-    if (!connected()) {
-      return def;
+    if (!connected() || m_statebroadcasted != mpd::connection_state::CONNECTED) {
+      return true;
     }
 
     if (!m_status) {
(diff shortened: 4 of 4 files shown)


patrick96 (changes requested):
Thanks for contributing :smiley: 

I think I see two main changes here:

* Use an event mask to only update on certain events
* If we don't track the song time, use polling instead of `recv_idle`. The path for `track_time = true` seems to be the same as before.

I understand why the first change reduces the number of unnecessary updates.
But for the second not so much. Why is using `poll` better than waiting for `mpd_recv_idle` to finish?

I have also found that if the mpd service is started while the module is running, the module is not updated and it starts printing the following error repeatedly after a few seconds:

```
error: module/mpd: Cannot send a new command while receiving another response
```

This is probably an event missing in the event mask.

Please also run the files you changed through `clang-format`: https://github.com/polybar/polybar/wiki/Style-Guide#code-formatting

Other than that the code looks good and also seems to work well.

jpalus:
Updated formatting in lines added in this PR (note that there were more changes done by clang-format seemingly unrelated).

Just to be sure we are on the same page let me elaborate a bit more on this PR.

**Before**:

The flow is like (always, whether MPD is doing anything or not and irrespective of user configuration):
1. send IDLE
2. immediately send NOIDLE
3. get any status change (recv_idle)
4. sleep for 80ms 
5. repeat (go to 1.)

The primary issue is that `recv_idle` after `noidle` returns immediately, it's only IDLE followed by `recv_idle` that blocks. This in turn means polybar wakes up ~10x a second doing IDLE/NOILDE cycle (involves network traffic). Noticed it on my laptop when in travel and on battery when network traffic meant additional VPN overhead, all that while MPD was not playing anything.

**After**:

Flow for case in which MPD is playing and user configured elapsed time to be displayed is the same as before.

In all other cases:
1. send IDLE
2. poll on file descriptor for 2 seconds (to have a chance of quitting without hanging waiting on mpd plugin)
3. if data is ready on file descriptor call `recv_idle` to process events, go to 1.
4. otherwise go to 2.

Regarding your question about not calling `noidle` in new case I suppose it's a misconception of how `recv_idle` behaves after NOIDLE.

My assumption was that `noidle` was intentional to keep high resolution tracking of elapsed time and since I'm not using it I wanted to focus on optimizing non-tracking time case.

To answer first question now:
> I understand why the first change reduces the number of unnecessary updates.
> But for the second not so much. Why is using poll better than waiting for mpd_recv_idle to finish?

Limiting events processed by mpd plugin is actually only a small improvement compared to the main one:
* reducing number of wakeups with polling goes from 10 wakeups/sec to 1 wakup/\<average song duration\>
* limiting set of processed events eliminates only wakups related to ie adding/removing songs to queue

As for the issue with:
```
error: module/mpd: Cannot send a new command while receiving another response
```
Do you have exact steps to reproduce? Most likely it means we called IDLE, followed by some other command other than NOIDLE. Which in turns means `noidle()` was not called when it should have been called.

Note that I think "time tracking case" can be improved too, but I wanted to focus on "non-time tracking case" in first iteration to limit chances of regression.

jpalus:
Correction to:
>reducing number of wakeups with polling goes from 10 wakeups/sec to 1 wakup/\<average song duration\>

which is unfair and factually incorrect. That should read:
> reducing number of wakeups with polling goes from 10 wakeups/sec to 1 wakup/2 sec and network traffic from 10/sec to 1/\<average song duration\>

patrick96:
Thanks for the thorough explanation. The documentation in the libmpdclient header file wasn't that clear to me.

If `mpd_recv_idle` blocks while in idle mode, why do we even need to use `poll`? Can't we make the code even simpler?

```cpp
      if (!track_time) {
        m_mpd->idle();
      }
      idle_flags = m_mpd->recv_idle();
```

If I understand correctly, this will wait indefinitely until a new event occurs.

> to have a chance of quitting without hanging waiting on mpd plugin

I guess it's because of that. If we move all that idle stuff into the `idle` function, this should no longer be a problem.
Right now this actually can't happen because the module update and the controller getting the module output run under different locks. Though this is one of the concurrency issues in polybar because it could technically lead to race conditions, but in practice I think this is quite rare because the controller is most likely able to get the module output during the time the module is sleeping.

I leave it up to you if you want to move the idle stuff into the `idle` function. Either is fine with me, but at some point in the future we will have to move it because the `has_event` function shouldn't really block.

Also, just a clarifying question for me: If mpd is in idle mode and an event occurs (matching the mask), it automatically leaves idle mode, correct?

> Do you have exact steps to reproduce?

Sure:

```dosini
[bar/example]
modules-left = mpd

enable-ipc = true

[module/mpd]
type = internal/mpd
format-online = <label-song>
format-offline = offline
label-song = %title%
```

Steps:

* Stop your mpd service
* Start polybar with that configuration
* Start the mpd service
* After a few seconds the error should start to appear

<details>
<summary>This is the log on my machine: </summary>

```
notice: Parsing config file: /home/patrick96/polybar/issues/2333/minimal.ini
* Created ipc channel at: /tmp/polybar_mqueue.2089160
* Loaded monitor HDMI-0 (1920x1080+1920+0)
* Bar geometry: 1920x24+1920+0; Borders: 0,0,0,0
error: module/mpd: Connection refused
* Starting application
* Starting module/mpd
error: module/mpd: Connection refused
* Configured DPI = 96x96
warn: No fonts specified, using fallback font "fixed"
notice: Loaded font "fixed" (name=Noto Sans, offset=0, file=/usr/share/fonts/noto/NotoSans-Regular.ttf)
* Bar window: 0x5600001
* module/mpd: Rebuilding cache
* Redrawing bar window
error: module/mpd: Connection refused
error: module/mpd: Connection refused
error: module/mpd: Connection refused
error: module/mpd: Connection refused
error: module/mpd: Connection refused
error: module/mpd: Cannot send a new command while receiving another response
* module/mpd: Rebuilding cache
error: module/mpd: Cannot send a new command while receiving another response
* module/mpd: Rebuilding cache
error: module/mpd: Cannot send a new command while receiving another response
* module/mpd: Rebuilding cache
error: module/mpd: Cannot send a new command while receiving another response
* module/mpd: Rebuilding cache
```
</details>

The same also happens if I add `<label-time>`: `format-online = <label-song> <label-time>`

I ran polybar through the debugger and it happens in the try-catch block in the `udpate` function. This function runs after `has_event` has finished, so it's not some sort of race condition.

patrick96:
One thing I noticed and now know where it comes from is that polybar now takes a significant amount of time to shut down. I believe that's because it has to wait for the `poll` to finish. We need to somehow wake up the poll (or `mpd_recv_idle`) call when we want to shut down the module.

jpalus:
Sorry for the silence but finally found some time to look into this.

Issue with "error: module/mpd: Cannot send a new command while receiving another response" should be resolved. See last commit with fix for details.

As for `poll()` being in `has_event()` or shutdown taking up to 2 secs it's a result of a compromise that I tried to find while trying to fit current design based on `idle()/has_event()/update()`. I suppose we could send a signal that would interrupt poll but I found it easier to just poll with small timeout like 2 seconds -- if that's too high to your liking let me know. Once `poll` starts using timeout return value becomes very important and unfortunately `idle()` does not return anything. I would have to communicate it in some other way like class field which would be even uglier. Essentially `poll()` translates to `idle until  it has_event ` which does not translate easily into current design.

I really wish polybar was based on some nice epoll wrapper like libevent:
- when disconnected register callback for a timer which attempts to connect
- once connected either register callback for events on MPD file descriptor if not tracking time or timer otherwise

A truly event driven library would be perfect match for MPD module.

patrick96:
I wrote down some of my thoughts about interruptability of modules in https://github.com/polybar/polybar/issues/2337#issuecomment-759636886. Ideally, all modules are interruptable when waiting on any kind of event.

After some research, it seems the best way to do this is to pass an extra file descriptor to `poll` and write to it whenever we want to wake up the module. I think that would be a good idea here as well.

I am planning some changes to the modules in this regard to make them more uniform (not all modules are currently interruptible). Maybe this will also make use of something like libuv or libevent.

But for now, I think passing an fd to `try_recv_idle` that we write to during `wakeup` would let us disable the timeout.

I will try to find some time next week to review this.

jpalus:
Regarding:

> Also, just a clarifying question for me: If mpd is in idle mode and an event occurs (matching the mask), it automatically leaves idle mode, correct?

and to elaborate a bit more about the fix -- from API perspective to fully process idle one needs to:
1. send idle command
2. either wait for event or send idle command
3. receive reply

The important step is last one. To answer your question -- yes event automatically leaves idle, as in you don't have to send noidle command, but you have to receive outcome to complete idle fully.

Skipping step 3 is what caused issue when MPD is started after polybar -- "broadcast" of new status forces `update()` after going into idle. Later song retrieval leaves idle through `check_prerequisites()` but since `noidle()` no longer does `recv_idle` idle was not fully completed and hence error. Now it's worth considering whether not to bring back previous behavior of doing `noidle()/recv_idle()` together -- preferably in new method ie `discard_noidle()` to emphasize  that we are effectively discarding any events that might have occurred. Personally I would lean towards the opposite -- drop `noidle()` from `check_prerequisites*()` and make sure idle is left with all events processed so that we don't loose state integrity. The only place that I would consider valid to do `noidle()/recv_idle()` discarding result is graceful shutdown of MPD module.

patrick96 (changes requested):
> Personally I would lean towards the opposite -- drop noidle() from check_prerequisites*() and make sure idle is left with all events processed so that we don't loose state integrity. The only place that I would consider valid to do noidle()/recv_idle() discarding result is graceful shutdown of MPD module.

I would tend to agree. The module should be built in a way that the connection should be in an idle state only while waiting for events and not while updating the module.

I no longer get the error with these changes and starting/stopping mpd while the bar is running is working.
However, one thing that is not working yet is starting the bar with mpd stopped and starting it while the module is running. The module prints `Connection refused` as long as it is stopped, but then doesn't do or print anything once I start the mpd service.

Regarding event-driven programming. I plan to introduce an event loop into almost all modules using `libuv`. I am currently working rewriting our main loop with libuv in #2384 and after that I want to get started on writing an eventloop for modules.
But this will take a considerable amount of time, so for now I think we should write our own solution for this module. As I mentioned before, we should pass an additional fd to `poll` that we can use to prematurely wake up the module, this would also allow us to increase the timeout.
A way this could work is adding a function to `mpdconnection` that can be called from any thread that just writes to that file descriptor.

PR #2336, opened 2020-12-26
module: Implement action router

patrick96:

* [x] Refactor
* [ ] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
Action implementation inside module becomes much cleaner because each
module just registers action names together with a callback (pointer to
member function) and the action router does the rest.

The implementation is neat because we can just port over modules one-by-one because the input method kept the exact same semantics. Modules just need to stop overwriting the `input` function and start registering their actions to use the new system.

## Related Issues & Documents

This would make it much easier to implement universal module actions like in #2320

### DIFF
--- a/include/modules/alsa.hpp
+++ b/include/modules/alsa.hpp
@@ -42,7 +42,7 @@ namespace modules {
 
     void change_volume(int interval);
 
-    void action_epilogue(const vector<mixer_t> mixers);
+    void action_epilogue(const vector<mixer_t>& mixers);
 
     vector<mixer_t> get_mixers();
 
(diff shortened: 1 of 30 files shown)


patrick96:
A possible optimization could be to only have one action router instance per module type instead of per module instance. In all current modules, the list of actions as well as their function pointers is the same for all instances of each module.

This gives us a constant overhead of one instance per module type while the number of instances in the current approach is the same as the number of modules in the bar.

All these instances could then be a static variable (not static member) defined in `base.inl` and the registration would somehow also need to take place statically (maybe initializer lists)

patrick96:
I have noticed that some modules create a new connection to their adapter when handling actions (mpd, i3). They likely do this because the actions run in the controller thread while the rest of the module runs in the module thread. This disparity likely also creates a host of different concurrency issues related to action handling.

I think in the future we should run the action handling code completely inside the module threads. For this each module could have an event queue and the controller simply pushes all actions into that queue. I will open a separate issue for this.

PR #2345, opened 2021-01-06
renderer: Keep Formatting State in separate component

### LINKED ISSUE #2344: Reduce Renderer Size
Currently, the renderer is responsible for both the rendering itself and maintaining state information about formatting from formatting tags (colors, underlines, fonts, ...).
I want to separate these two responsiblities and make the renderer responsible only for rendering text on the bar.
Another component would then keep track of the currently set formatting context which the renderer would access during rendering.
The renderer would then only be controlled through a small set of rendering primitives like drawing text and drawing offsets.

Besides following the single responsibility principle, this would have many other benefits:

* Most of the renderer code is not really testable because it produces visual output. Keeping that code as small as possible is desirable
* It would allow us to more easily substitute different renderers which would help with testing (mocks) and supporting different display servers (wayland).
* We could implement and test stackable formatting tags (#1623) in a dedicated component instead of directly in the renderer.

I have a few hours right now, I'll see if I can whip up a proof of concept.

patrick96:

* [x] Refactor
* [ ] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update
* [ ] Other:

## Description

This is a first proof of concept for #2344

Right now only the current colors, font and attribute activations are stored.
But the plan is to also keep track of all action areas and alignment so that the
renderer is completely independent from the actual formatting tags.

## Related Issues & Documents

Closes #2344

Maybe we want to write a page about the architecture.

### DIFF
--- a/include/components/bar.hpp
+++ b/include/components/bar.hpp
@@ -10,6 +10,7 @@
 #include "events/signal_fwd.hpp"
 #include "events/signal_receiver.hpp"
 #include "settings.hpp"
+#include "tags/context.hpp"
 #include "utils/math.hpp"
 #include "x11/types.hpp"
 #include "x11/window.hpp"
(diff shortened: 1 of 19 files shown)


patrick96:
I believe this is decent progress for now. The renderer still isn't completely decoupled from `bar.cpp`, there are still some functions that are not in the interface. But now the renderer doesn't have to store information about actions and formatting state anymore, all of this is delegated to dedicated context classes.
This also removes all parsing related signals, all these now go through the renderer interface.

PR #2359, opened 2021-01-19
feat: Rounded bar border

h45h74x:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
This PR exists to change the behaviour when applying a `radius` together with `border-size` with a minimal amount of code changes. Polybar is now able to display a round border (also supporting the functionality added in https://github.com/polybar/polybar/pull/2297) :tada: 

In order to avoid bigger changes, I've implemented it so that the rounded border gets its size and color from the top border configuration (`border-size` / `border-top-size` and `border-color` / `border-top-color` respectively). 
While I am not sure how to implement the configuration in a more user-friendly way without changing the current configuration options, I would be glad to add any changes you suggest.

## Related Issues & Documents
While the idea to implement this came up whilst styling my system, this issue is a good representation: https://github.com/polybar/polybar/issues/1566

Several quick screenshots as visual demonstration (with and without compositor) can be found [here](https://short.sgruber.at/scr_git_polybar_roundedborder)

It would be necessary to remove the incompatibility notice regarding the `radius` and `border-size` configuration options.

### DIFF
--- a/include/cairo/context.hpp
+++ b/include/cairo/context.hpp
@@ -127,8 +127,6 @@ namespace cairo {
     }
 
     context& operator<<(const circle_segment& segment) {
-      // cairo_arc(m_c, arc.x, arc.y, arc.radius, 0, 2 * M_PI);
-
       double degree = M_PI / 180.0;
 
       cairo_new_sub_path(m_c);
--- a/include/cairo/context.hpp
+++ b/include/cairo/context.hpp
@@ -278,15 +278,6 @@ namespace cairo {
       return *this;
     }
 
-    context& stroke(bool preserve = false) {
-      if (preserve) {
-        cairo_stroke_preserve(m_c);
-      } else {
-        cairo_stroke(m_c);
-      }
-      return *this;
-    }
-
     context& mask(cairo_pattern_t* pattern) {
       cairo_mask(m_c, pattern);
       return *this;
--- a/include/cairo/context.hpp
+++ b/include/cairo/context.hpp
@@ -126,6 +126,24 @@ namespace cairo {
       return *this;
     }
 
+    context& operator<<(const circle_segment& segment) {
+      // cairo_arc(m_c, arc.x, arc.y, arc.radius, 0, 2 * M_PI);
+
+      double degree = M_PI / 180.0;
+
+      cairo_new_sub_path(m_c);
+
+      cairo_arc(m_c, segment.x, segment.y, segment.radius, segment.angle_from * degree, segment.angle_to * degree);
+      cairo_line_to(m_c, segment.x, segment.y);
+
+      cairo_arc(m_c, segment.x, segment.y, segment.radius, segment.angle_from * degree, segment.angle_to * degree);
+      cairo_line_to(m_c, segment.x, segment.y);
+
+      cairo_close_path(m_c);
+
+      return *this;
+    }
+
     context& operator<<(const textblock& t) {
       double x, y;
       position(&x, &y);
@@ -258,6 +276,15 @@ namespace cairo {
         cairo_fill_preserve(m_c);
       } else {
         cairo_fill(m_c);
+      }
+      return *this;
+    }
+
+    context& stroke(bool preserve = false) {
+      if (preserve) {
+        cairo_stroke_preserve(m_c);
+      } else {
+        cairo_stroke(m_c);
       }
       return *this;
     }
--- a/include/cairo/context.hpp
+++ b/include/cairo/context.hpp
@@ -116,26 +116,20 @@ namespace cairo {
     }
 
     context& operator<<(const rounded_corners& c) {
-      double d = M_PI / 180.0;
       cairo_new_sub_path(m_c);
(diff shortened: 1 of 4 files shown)


patrick96:
This looks really cool, especially the one where only two opposing corners are rounded :star_struck:

I think taking the top border value is fine. I'm not sure it makes sense to have different border sizes when using rounded corners. A corner is part of two borders, so the two borders should be identical otherwise we can't really draw a rounded corner for it.

We could make it slightly more configurable by using the bottom border for the bottom corners and the top border for the top ones. We then either tell the user in the documentation that the two adjacent borders for a rounded corner should be equal or we even enforce this when loading the config.

Thanks for implementing this. I didn't think it would look this good :tada: 

I'll try to review this tomorrow. But I have exams coming up, so no promises.

h45h74x:
Thanks for the positive feedback!
_(I almost thought this PR would be rejected because of the `wontfix` tag in the issue...)_

> We could make it slightly more configurable by using the bottom border for the bottom corners and the top border for the top ones.

I think that's a great idea, but is cairo able to merge the upper and lower section (thinking of color and line width) together seamlessly?

Also, take your time with the review. I am not in a hurry with implementing this, I already have it on my machine :D

patrick96:
>  (I almost thought this PR would be rejected because of the wontfix tag in the issue...)

At the time I didn't know it would be this easy ;)

(And I didn't want to invest the time to figure it out)

> I think that's a great idea, but is cairo able to merge the upper and lower section (thinking of color and line width) together seamlessly?

I doesn't have to be able to do that. When there is a requirement (or expectation) that the border settings for the two adjacent borders to a rounded corner have to be the same, then we never have to blend stuff together. For example, if someone has the top left and bottom left corner rounded, by this requirement the top border must equal the left border and the left border must equal the bottom border, so no blending required on the left border.

h45h74x:
I've changed the implementation so that the border is way more customizable now.
_(Previously the border was one single subpath in cairo which did not allow for much customisation)_

Circle segments with separate, adjustable radii are used for the corners, the offset and length of the regular horizontal and vertical borders is adjusted accordingly. Also, the color of the corners is now always matching the left / right vertical border color.

There are two more screenshots in my [nextcloud folder](https://short.sgruber.at/scr_git_polybar_roundedborder) now, depicting the result of the following two configurations:

_colors_
```
border-top-color=#0000ff        ; blue
border-bottom-color=#00ffff     ; cyan
border-left-color=#ff0000       ; red
border-right-color=#ffff00      ; yellow
```

polybar_test_01.png
```
border-top-size=10
border-bottom-size=10
border-left-size=10
border-right-size=10

radius-top-left=10
radius-top-right=20
radius-bottom-left=30
radius-bottom-right=0
```

polybar_test_02.png
```
border-top-size=10
border-bottom-size=20
border-left-size=10
border-right-size=10

radius-top-left=10
radius-top-right=20
radius-bottom-left=0
radius-bottom-right=0
```

h45h74x:
Hi, @patrick96. As I said, I'm not in a rush to merge this PR, but did you already have a look at my changes? :D

patrick96:
I did look a bit, just not enough to do a full review. I noticed that you can quite weird looking things (border going to the edge with a rounded edge beneath it), so it might be a good idea to print a warning if the border settings of two borders adjacent to a rounded edge are not the same.

Will do a full review later.

include/cairo/context.hpp
```diff
@@ -126,6 +126,22 @@ namespace cairo {
       return *this;
     }
 
+    context& operator<<(const circle_segment& segment) {
```
patrick96, line 129: Can the `<<` operator for `rounded_corners` also reuse this code? It seems there is a lot of code in common here.

EDIT: Maybe not entirely, since this also does lines.

include/cairo/context.hpp
```diff
@@ -262,6 +278,15 @@ namespace cairo {
       return *this;
     }
 
+    context& stroke(bool preserve = false) {
```
patrick96, line 281: Are you using this anywhere?

patrick96 (commented):
Awesome. Everything looks good now :smiley: 

Just a few small comments to resolve.

I think having a note on the wiki about using the same border setting for adjacent corners should be enough. I can no longer reproduce the truly weird behavior, probably because I was using an older version back then, so I don't think it is necessary to write a bunch of code to give a warning message.

include/cairo/context.hpp
h45h74x: I've removed the duplicate lines that were in there somehow and moved the degree value to a constant to avoid recalculating it every time the method is called.

I also don't think repurposing the `cairo_arc` method for use with `rounded_corners` AND `circle_segment` is necessary, but you tell me :D

patrick96 (approved):
Looks good :)

Thanks a lot, this is really cool :smiley:

h45h74x:
Thank you for letting me contribute :D

Issue #2354, opened 2021-01-16
Escape characters in configs


patrick96:
Currently, everything in the config is read as-is. There is no way to escape characters with special meaning.

This isn't really an issue right now because the config recognizes special characters only if the value starts with `${` and ends with `}` (references).

However, if we want to allow something like `key = foo ${section.key} bar`, we need to have a way to escape literal uses of `${` to not irreparably break existing configs.

I would propose the backslash (`\`) as the escape character because it's the most commonly used escape character.

Having an escape character here would also allow us to implement multiline config options by putting the backslash as the last character in a line (like in bash). This would also make #2207 obsolete.

Unfortunately, just making `\` an escape character may break existing configs that use it as a literal character.
I propose a multi-step process to implement this functionality, to give users time to escape their literal backslashes:

**1. Add warning or error to config parser for single backslashes**

Add an error to the config parser when there is a single backslash in a config value and tell the user to escape their backslashes.
Since the backslash isn't used for escaping anything yet, all single backslashes in a config are literal backslashes. 
We should still treat single backslashes as literal backslashes for backwards compatibility.

Because we tell the user to escape their backslashes, we also need to treat two backslashes (`\\`) as a single literal backslash.
This is the only breaking change here: If a user uses two consecutive literal backslashes, they will now be interpreted as a single literal backslash.
I think the risk of this is quite low.

At this point, the backslash is a proper escape character, it's just not used for anything else than escaping other backslashes.

**2. Using the backslash to create multi-line config values**

Any config value that ends in a backslash is joined together with the next line (the value will **not** contain a newline character).
Here, we need to take care to recognize both `\n` and `\r\n` as valid line-endings.

For users that have now properly escaped their literal backslashes, this should not break anything.
Otherwise it will break configs where there are an odd number of backslashes at the end of the line.

This would make #2207 obsolete

**3+. Introduce constructs with special meanings to config values**

We can now start adding things like config values where we can mix references and text.

Each time we do that, it may break certain configs that used that construct in a literal sense (for example someone using `${` as a literal string).

---

I think we can target 1 for a release and 2 & 3 for the release after that.

The error introduced in 1 should stay and we should also keep interpreting single backslashes that don't escape anything as literal backslashes.

I doubt that many people will be affected by any one of these steps. But I'm sure that there are some people that the backslash as a separator, so we should properly communicate what we're doing.

Filip62:
I will try to implement this.

patrick96:
[...]

PR #2361, opened 2021-01-21
Add initial support for an escape character #2354

Filip62:

* [x] Feature
* [x] Documentation Update
* [ ] Refactor
* [ ] Bug Fix
* [ ] Optimization

## Description

* Introduce an escape character '\\'.
* Warn users about unescaped characters (only backslashes for now).
* For now, treat a single unescaped backslash as a literal backslash character.
* Treat two consecutive backslashes as a single literal backslash character.

## Related Issues & Documents
Implements first step of #2354

See #2354  

### DIFF
--- a/src/components/config_parser.cpp
+++ b/src/components/config_parser.cpp
@@ -257,6 +257,8 @@ std::pair<string, string> config_parser::parse_key(const string& line) {
     throw invalid_name_error("Key", key);
   }
 
+  value = parse_escaped_value(move(value), key);
+
   /*
    * Only if the string is surrounded with double quotes, do we treat them
    * not as part of the value and remove them.
@@ -292,4 +294,25 @@ bool config_parser::is_valid_name(const string& name) {
   return true;
 }
 
+string config_parser::parse_escaped_value(string&& value, const string& key) {
+  string cfg_value = value;
+  bool log = false;
+  auto backslash_pos = value.find('\\');
+  while (backslash_pos != string::npos) {
+    if (backslash_pos == value.size() - 1 || value[backslash_pos + 1] != '\\') {
+      log = true;
+    } else {
+      value = value.replace(backslash_pos, 2, "\\");
+    }
+    backslash_pos = value.find('\\', backslash_pos + 1);
+  }
+  if (log) {
+    // TODO Log filename, and line number
+    m_log.err(
+        "Value '%s' of key '%s' contains one or more unescaped backslashes, please prepend them with the backslash "
+        "escape character.",
+        cfg_value, key);
+  }
+  return move(value);
+}
 POLYBAR_NS_END
--- a/tests/unit_tests/components/config_parser.cpp
+++ b/tests/unit_tests/components/config_parser.cpp
@@ -25,6 +25,9 @@ class TestableConfigParser : public config_parser {
   using config_parser::parse_line;
 
  public:
+  using config_parser::parse_escaped_value;
+
+ public:
   using config_parser::m_files;
 };
 
@@ -64,6 +67,12 @@ vector<pair<pair<string, string>, string>> parse_line_key_list = {
     {{"key", "value"}, " key = value"},
     {{"key", ""}, " key\t = \"\""},
     {{"key", "\""}, " key\t = \"\"\""},
+    {{"key", "\\"}, " key = \\"},
+    {{"key", "\\"}, " key = \\\\"},
+    {{"key", "\\val\\ue\\"}, " key = \\val\\ue\\"},
+    {{"key", "\\val\\ue\\"}, " key = \\\\val\\\\ue\\\\"},
+    {{"key", "\\val\\ue\\"}, " key = \"\\val\\ue\\\""},
+    {{"key", "\\val\\ue\\"}, " key = \"\\\\val\\\\ue\\\\\""},
 };
 
 INSTANTIATE_TEST_SUITE_P(Inst, ParseLineInValidTest, ::testing::ValuesIn(parse_line_invalid_list));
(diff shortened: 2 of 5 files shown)


patrick96 (changes requested):
Thanks :)

I think everything works, good job!

As far as I can tell you are using `positive_offset` and `negative_offset` to keep track of the difference in indices between `cfg_value` and the returned value (and `escaped_value`).

Now, I don't think it's necessary to generate the escaped string for when there are unescaped backslashes. In that case, it would be possible to use the same string object for searching and replacing without having to keep track of index differences.

There are also a few other things that should be in this PR:

* The manpage for the polybar config should be updated to include information about the escape character. The file is in `doc/man/polybar.5.rst` and the changes should go in the `Keys` section.
* New tests for the new function as well as some added testcases for the `ParseLineKeyTest` that include backslashes. The tests belong in `tests/unit_tests/components/config_parser.cpp`
* A changelog entry describing your changes, including breaking changes: https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog

src/components/config_parser.cpp
```diff
@@ -292,4 +294,29 @@ bool config_parser::is_valid_name(const string& name) {
   return true;
 }
 
+void config_parser::is_valid_value(string& value, const string& key) {
```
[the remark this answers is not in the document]
Filip62: I am not sure how exactly should it look,  but I will do at least something.

patrick96:
Let me know once this is ready for review again.

Also, the CI docs fail isn't your fault, there seems to be an issue with an ubuntu mirror.

src/components/config_parser.cpp
```diff
@@ -292,4 +294,23 @@ bool config_parser::is_valid_name(const string& name) {
...
+  bool log = false;
+  auto backslash_pos = value.find_first_of('\\');
+  while (backslash_pos != string::npos) {
+    if (value[backslash_pos + 1] != '\\') {
+      backslash_pos = value.find_first_of('\\', backslash_pos + 1);
+      log = true;
+    } else {
+     value = value.replace(backslash_pos, 2, "\\");
+     backslash_pos = value.find_first_of('\\', backslash_pos + 1);
+    }
```
patrick96, line 308: You have some weird indentation here (2 spaces in the if branch and 1 in the else branch), please use 2 spaces.

Also setting the `backslash_pos` is the same in both branches, you can move it after the if-else block.

src/components/config_parser.cpp
```diff
@@ -292,4 +294,23 @@ bool config_parser::is_valid_name(const string& name) {
   return true;
 }
 
+string config_parser::parse_escaped_value(string&& value, const string& key) {
+  string cfg_value = value;
+  bool log = false;
+  auto backslash_pos = value.find_first_of('\\');
+  while (backslash_pos != string::npos) {
+    if (value[backslash_pos + 1] != '\\') {
```
patrick96, line 302: This can still lead to an out-of-bounds access. Maybe this is the cause of the segfault.

The special case happens only if there is an unescaped backslash at the end of the line, the behavior should be the same as if there was an unescaped backslash anywhere else.

tests/unit_tests/components/config_parser.cpp
```diff
@@ -236,3 +246,34 @@ TEST_F(ParseHeaderTest, throwsSyntaxError) {
   EXPECT_THROW(parser->parse_header("[root]"), syntax_error);
 }
 // }}}
+
+// ParseEscapedValueTest {{{
+
+/**
+ * \brief Class for parameterized tests on parse_escaped_value
+ *
+ * The first element of the pair is the expected key-value pair and the second
```
patrick96, line 255:
```suggestion
 * The first element of the pair is the expected value and the second
```

patrick96 (changes requested):
Code looks good now. Just some minor adjustments.

Also, I'm curious why you are using `find_first_of` instead of `find`

patrick96:
Also, it seems that CI is also having a segfault. I was also able to reproduce it sometimes. Looking at the coredump, it looks like it's something inside the logger, but the actual errors are kind of weird. I suspect that it's because you used string concatenation for the log message.

src/components/config_parser.cpp
```diff
@@ -292,4 +294,23 @@ bool config_parser::is_valid_name(const string& name) {
...
+    } else {
+     value = value.replace(backslash_pos, 2, "\\");
+     backslash_pos = value.find_first_of('\\', backslash_pos + 1);
+    }
+  }
+  if (log == true) {
+    m_log.err("Value '" + cfg_value + "' of key '" + key +
+              "' contains one or more unescaped backslashes, please prepend them with the backslash escape character.");
+  }
+  return move(value);
```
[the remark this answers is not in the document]
Filip62: Clangd tells me that it will be copied if I remove the move call.

Filip62:
> Code looks good now. Just some minor adjustments.
> 
> Also, I'm curious why you are using `find_first_of` instead of `find`

No reason, changed it to `find`.

> Also, it seems that CI is also having a segfault. I was also able to reproduce it sometimes. Looking at the coredump, it looks like it's something inside the logger, but the actual errors are kind of weird. I suspect that it's because you used string concatenation for the log message.

Printf tokens didn't help.

src/components/config_parser.cpp
Filip62: Could you please explain why?
I added a fix if I understand what you meant, but the segfaults are still there.

src/components/config_parser.cpp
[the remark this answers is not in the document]
patrick96: In that case, leave it. I'm sure the compiler know more than me ;)

patrick96 (changes requested):
I think I found the issue with the segfault. It was undefined behavior in test code I wrote.

I pushed a fix in https://github.com/polybar/polybar/pull/2364.

After that is merged we can see if there are any other CI issues with your code.
But other than that, I think we can merge this soon.

src/components/config_parser.cpp
patrick96: You're change is correct. Also, this was not the cause of the segfault.

patrick96:
Can you rebase against master so that your branch includes my fix?

src/components/config_parser.cpp
```diff
@@ -292,4 +294,27 @@ bool config_parser::is_valid_name(const string& name) {
...
 
+string config_parser::parse_escaped_value(string&& value, const string& key) {
+  string cfg_value = value;
+  bool log = false;
+  auto backslash_pos = value.find('\\');
+  while (backslash_pos != string::npos) {
+    if (backslash_pos == value.size() - 1) {
+      log=true;
+    } else if (value[backslash_pos + 1] != '\\') {
+      log = true;
```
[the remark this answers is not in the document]
Filip62: Yeah I was just testing.

Filip62:
If we remove the `backslash_pos == value.size() - 1` check, why would it lead to out of bounds access?
Wouldn't it be just looking at the string termination character?

patrick96 (approved):

patrick96:
Everything looks good now, but I can't merge yet because there is still a merge conflict (though it wasn't there a minute ago). You need to resolve it before I can merge this.

Filip62:
> Everything looks good now, but I can't merge yet because there is still a merge conflict (though it wasn't there a minute ago). You need to resolve it before I can merge this.

I am not sure if I am missing something but `git merge upstream/master` tells me I am up to date.

patrick96:
I think you need to to `git rebase upstream/master`.

patrick96:
I think https://github.com/polybar/polybar/pull/2361/commits/6b21044529e2344861abcba3b57b658efaa921da messed things up a bit

Filip62:
Sorry for breaking things, still get confused by git sometimes.

patrick96:
There was still an issue. I went ahead and just squashed all commits.

patrick96:
Done! Thanks a lot for all your efforts :smiley:

Filip62:
Thanks a lot for all YOUR efforts.

Also, if you don't mind:


> If we remove the `backslash_pos == value.size() - 1` check, why would it lead to out of bounds access?
> Wouldn't it be just looking at the string termination character?

patrick96:
C++ strings don't have a termination character. If you do str[str.size()], you will access one element to the right of the last one.
The `std::string` class stores the string size separately and doesn't need a termination character.

Filip62:
I should've known that, I know the size is stored internally, but no one ever told me that the null termination is not there in C++, I am amazed and embarrassed.

PR #2365, opened 2021-01-27
add reverse-scroll option for xworkspaces module

tchici:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
Added reverse-scroll option for xworkspaces module. The default is to set it to false, that way users do not have to change their configuration files if they want old behavior.

## Related Issues & Documents
None

The new setting should be documented in the xworkspaces section, probably under basic settings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   repository.
 
 ### Added
+- `internal/xworkspaces`: `reverse-scroll` can be used to reverse the scroll
+  direction when cycling through desktops.
 - The backslash escape character (\\).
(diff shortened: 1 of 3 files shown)


patrick96 (changes requested):
Thanks :smiley: 

We recently decided that the `next` action should always go to the right and `prev` always to the left. Because of that the `reverse-scroll` option should be applied when producing the scroll actions in `get_output` (the same way it is done in the i3 and bspwm modules)

patrick96 (changes requested):
Looks good now :)

I forgot to mention this before, please add your changes to the changelog: https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog

patrick96 (approved):
Perfect!

Issue #2337, opened 2020-12-27
Per-module event queue


patrick96:
This is part of a larger effort to rework the main loop (#1778)

The action handler in a module runs in the caller thread (the controller thread) and not in the module thread. This requires us to be very careful about concurrent access and synchronization. Some modules solve this by creating a new connection to their adapter whenever an action is called (i3, mpd). For most other modules it is not clear whether the action code is actually save, it just seems to work (in almost all cases the action code will be called while the module sleeps).

I think we should do three things:

1. Create an event queue inside each module. When the controller wants to deliver an action, it can put that action into the event queue and wake the module (if it is sleeping). After waking up the module works through all received actions (in the module thread). We can either use the lock-free queue we are already using in the controller or just have a lock on the queue.
2. Clearly define the interface of a module and which functions may be called from other threads and which must be called from the module thread.
3. Ensure that all modules adhere to that interface and don't cause concurrency issues.

The general idea is to put as much as possible into the module threads and let the controller communicate with the module through only a narrow interface. If possible, using only atomic primitive variables or small locks around certain variables (e.g. the module output).

patrick96:
I don't think doing this is actually possible.

The problem is waking up the module after something was placed in the queue. This is fairly easy if the module sleeps on the `m_sleephandler` condition variable. But for modules that idle in a different way (for example polling a socket), this is not possible. There we must wait for the module to finishing idling.

Because of that, having an event loop doesn't work for all modules.

To make reasoning about concurrency easier and to make modules less complex, we should strive to guarantee the following:
* The module implementations should never have to acquire locks unless they do their own concurrency separate from the controller-module concurrency.
* Some functions may have restrictions on what they can access inside the class

Ideally, the module implementation should not be aware of concurrency, mutual exclusion should be guaranteed by the base modules. However, this may not be that easy.

Using the same lock for `update` and `input` is required for proper synchronization. At the moment, `get_output` should also use the same lock because modules access local state there. #1752 may help here because it would allow us to do all of `get_output` the same way for all modules, so we can get granular with locks.

But either way, we don't want the controller to stall when `update` is waiting for data (for example from the network).

One way to do this is to split `update` into two functions:
* A collector function where the module collects the raw data
* The actual update function that uses that raw data to set the module state and configure the labels.

We don't need any synchronization between these two functions because they run sequentially in the module thread.

Whatever code is responsible for collecting the data should be able to run concurrently with the `get_output` and `build` function (i.e. use different locks).
[...]

Issue #1778, opened 2019-05-21
Reworking main loop


Lomadriel:
As explained in #1732, there are some concurrency issues in polybar and this issue is a try to address this problem. Nothing is set in stone, feels free to give your opinion! :smiley: 

At the moment a lot of threads could be removed (more or less easily) and removing these threads could be a way to address some concurrency issues. That's why I propose to rework the main loop.

The mainloop should look like this (pseudo-code):
```
while (!stop)
  dispatch_input_events();
  update_modules();
  if (dirty)
      build_output();
      redraw();
  sleep_until_next_update_is_needed();
```

- The `collect_and_dispatch_input_events()` should collect the input events (click, scroll, …), resolves the targets and transmits the events. When a module gets an event, it is added into a queue of event to process.
- The `update_modules` function calls the `update` method of each module, and causes the module to update it internal state (the labels). 
- The `redraw` function simply redraw the bar.
- The `sleep_until_next_update_is_needed()` sleep until an event has been receive by a module or by the controller for the input events. In case there is no event, the controller sleep until the next required update (`internal` property of each module).

In this architecture each module still need to gets the other events by themselves.
When an event is receive by a module, the event can be ignored but should not modify the internal state, the event is then stored into a queue that will be processed during the update. The only variable that should be modified is a flag signaling that the module should be updated. Why a flag and not an event to the controller -> for the separation in multiple SO files and the plugin system.

Also, the `build` and `update` mutexes should be merged into one, since in the current architecture they should be locked at the same time almost every time (this is impossible now because some modules wait into their `has_event()` method). The scope of the should be as "small" as possible. At the moment, some mutexes are locked while no sensible data are accessed or even during a sleep !!! (See the method `poll_events` in `inotify_module.hpp`)

I tried to keep in mind that we also aim to separate every modules in several SO files but if I missed something please let me know. :)

If something is unclear, feel free to ask.
I will try to implement a dummy prototype this week-end.

patrick96:
I think, we need to start working on parts of this.

The first step should probably be to merge the `process_eventqueue` thread into the `read_events` loop so that there is a single controller thread. Because right now there can be up to 3 threads running in a module:
* The `process_eventqueue` thread getting module output or sending a module action
* The main thread in `read_events` sending an ipc hook or xcb event
* The module thread itself

This makes it quite difficult to reason about concurrency.

[...]

PR #2384, opened 2021-02-17
Use libuv for the controller's event loop

patrick96:

* [x] Refactor
* [ ] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

This is a first proof of concept for streamlining our main loop. For now I just replaced the event loop listening for changes from the X server, IPC, signals and config file with a libuv based event loop. This made the code a lot easier and also cleaner than our fd-based loop. It also works with things other than fds, like files, pipes, signals, and even async wake up.

There are still some things that I want to do before merging this:

* [x] Use [`uv_pipe_t`](http://docs.libuv.org/en/v1.x/pipe.html) for monitoring the IPC fifo
* [x] Create separate eventloop component that hides things like allocations from us and provides an easy C++-style interface to libuv for the controller to use (if possible with callbacks to member functions).
* [x] Integrate `process_eventqueue` thread into event queue (either with async or writing to a pipe).
* [x] Handle error states
* [x] Add dependency changes to `CHANGELOG.md`
* [x] Figure out the minimum version of libuv we should use (1.8.0 is the highest version supported by all supported debian and ubuntu versions).
* [x] Close all handles and requests before closing the loop
* [x] Address all TODO comments

I think we should also use this event loop for handling mouse events on the bar window instead of using `taskqueue` in `bar.cpp`, but this can be a separate step after this is merged.

## Related Issues & Documents

This is the first step in reworking our main loop: #1778

I think after this, we can more or less easily implement #675 (but it needs special handling for `include-directory`).

A similar event loop will likely also be added to modules that need it (all, except static): #2337

We need to add libuv >= 1.3 as a dependency

* `libuv1-dev` on ubuntu
* `libuv` on Fedora
* `libuv-devel` on openSUSE


vincentbernat:
Previously, on first start, Polybar was half-frozen: display was updated, but IPC events and X events were not dispatched correctly. I was unable to find the cause. This is not the case anymore. Good work!

patrick96:
Well, that's a nice side effect :)

Thanks for letting me know :+1:

PR #2386, opened 2021-02-18
feat(xworkspaces) Allow to show the nwin less

kronn:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

Show the number of windows only if they exceed the configured threshold.
This way, one can only the number of windows when there are multiple.
With a threshold of 1, the following behavior follows:

  0 → inactive/empty workspace
  1 → active workspace, no number
  2 or more → active workspace and number of windows

The default is 0 to match the current behavior.

The config-key is called window-number-threshold

## Notes

This is a working implementation of an idea. I can imagine several changes and improvements for this, but want to get feedback before taking a deep-dive.
- I did not touch the tests (sadly).
- I recognize the need to document this in the wiki and maybe in the example configuration.
- I am unsure about naming things (and of course cache invalidation and off-by-one errors, but mostly naming things, in this case)
- for all the above reasons, I also did not touch the CHANGELOG.

I am looking and am open for feedback. :-)

## Related Issues & Documents

none AFAIK

### DIFF
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -113,6 +113,7 @@ namespace modules {
     bool m_click{true};
     bool m_scroll{true};
     bool m_revscroll{false};
+    unsigned int m_nwin_threshold{0};
     size_t m_index{0};
 
     // The following mutex is here to protect the data of this modules.
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -44,6 +44,7 @@ namespace modules {
     m_click = m_conf.get(name(), "enable-click", m_click);
     m_scroll = m_conf.get(name(), "enable-scroll", m_scroll);
     m_revscroll = m_conf.get(name(), "reverse-scroll", m_revscroll);
+    m_nwin_threshold = m_conf.get(name(), "window-number-threshold", m_nwin_threshold);
 
     // Initialize ewmh atoms
     if ((m_ewmh = ewmh_util::initialize()) == nullptr) {
@@ -266,6 +267,8 @@ namespace modules {
    */
   void xworkspaces_module::rebuild_desktop_states() {
     std::set<unsigned int> occupied_desks;
+    string nwin;
+
     for (auto&& c : m_clients) {
       occupied_desks.insert(c.second);
     }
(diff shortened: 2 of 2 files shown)


patrick96:
I am wondering a bit about possible use-cases for this feature. In most cases, having the `%nwin%` token be empty, will make the label look weird.

For example:
```dosini
[module/ws]
type = internal/xworkspaces
window-number-threshold = 1
label-active = %name% %nwin%
label-active-padding = 2
label-active-background = #DE935F
label-occupied = %name% [%nwin%]
label-occupied-padding = 2
label-occupied-background = #373B41
label-empty-padding = 2
```

With the threshold set to 0, it would look like this:
normal

However, when setting the threshold to 1, suddenly the brackets are empty and there is an extra space after the workspace name of the active workspace:
weird

It seems to me that this would create bars that look a bit off, but maybe there is a use-case that I haven't thought about.
What is your use-case for this feature?

kronn:
My main use-case goes like this: Depending on the monitor-size, xmonad chooses a different layout. So usually, while in the office an attached to a wider monitor, I have all coding-windows visible on the coding-workspaces. When I detach the laptop and finish working on the train I do not have an external monitor and therefore less screen estate. I then have the coding-windows "behind" each other, only one really visible at the time. I can manually switch the layout to see them all, but most of the time I don't.

Essentially, I know which windows are where if I know how many windows are in a workspace.
However, I find It distracting to see the number "1", since the active-state implies at least one window. 
Therefore, I wanted the threshold.

I wanted to see how easy this extension would be (very) and if there is interest to have this available.

The visual clutter you mentioned is indeed a merge-blocker for me.
I am really unsure how to solve this. If this were done without configuration the resulting implementation could grow rather complex. I would involve parsing the label and deciding whether to add one of more spaces (I think). In order to simplify this, I would (maybe) need to introduce more configuration options (something to wrap the token, so I know what semantically belongs to the `%nwin%`), resulting in more coding to keep the defaults simple again. Am I missing some convention that would make this easier or do you have a recommendation?

patrick96:
That's a pretty interesting use-case that I hadn't thought about.

> I am really unsure how to solve this. If this were done without configuration the resulting implementation could grow rather complex. I would involve parsing the label and deciding whether to add one of more spaces (I think).

polybar can't intelligently deal with empty tokens because it doesn't know what the user intended.

We won't get around further config options if we want to intelligently handle empty tokens.

> Am I missing some convention that would make this easier or do you have a recommendation?

I do have some ideas.

In some way, the user needs to tell us how a token should look when it has and when it doesn't have a value.
For example like this:

```dosini
label-active = %name%%nwin%

token-nwin = " [%token%]"
token-nwin-empty = ""
```

So when replacing the `%nwin%` token, we would first look up how the user defined the token should look.

This still has some cases where this won't work:
* When you want the same token to look different in different labels. For this we may need to also introduce per label token settings like `label-active-token-nwin`.
* Something like `%token1% %token2%%token3%` and you want the space to be gone only when both tokens are empty. I don't think there is a good way to do this.

Doing something like this would be quite a big endeavor I think.

Issue #1752, opened 2019-04-23
Custom labels, ramps, progressbars


patrick96:
This is just a quick brainstorm I wrote out about making labels, ramps etc. more flexible. `sdk` on IRC wanted it converted to an issue so they can track it.
---

Don't restrict what labels, ramps, etc a user can specify

A user should be able to set
```dosini
format = <label-abc> <ramp-abc>

label-abc = Text
ramp-abc-0 = A
ramp-abc-1 = B
ramp-abc-2 = C
```

And polybar will automatically deduce what is a label and what is a ramp and load them from the config accordingly.

The percentage a ramp or progressbar is tied to in a certain module should be obvious, where it is not, we should allow users to specify

This would make lots of code handling label loading in modules redundant. The customizability is then mainly constrained by what formats we allow.

This approach may prove harder to integrate with dynamic labels like `label-mode` in the bspwm module that gets replaced by other labels depending on the state.

albakham:
Would be great to see this!

pbogre:
has this feature been abandoned? i am absolutely looking forward to this :)

PR #2394, opened 2021-02-28
add units support (POINT, PIXEL, SPACE) for polybar

### LINKED ISSUE #1651: Allow `px` units when specified

frebib:
This is an almost direct rebase of #1671

Fixes #1651, #951, #1700 and #1265

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

Disclaimer: I merely rebased this from the original PR. I changed as little as possible to make it work. I expect there will be plenty of style and similar comments in the review, which I intend to fix before this is merged. Created as a draft to collect feedback before final round of reviewing

## Checklist:
* [x] Compile & Run
* [x] clang-format
* [ ] Unit tests for `unit.hpp`
* [ ] Decide if negative spacing is allowed
* [ ] Go through all config options that use extents or percentage_with_offset and see how they deal with negative values (and if they're even allowed).
* [ ] Address the weirdness in `base.inl`: https://github.com/polybar/polybar/pull/1671#discussion_r369192910
* [ ] Community testing
* [ ] Probably some other things

### DIFF
--- a/src/components/builder.cpp
+++ b/src/components/builder.cpp
@@ -161,7 +161,7 @@ void builder::node(const label_t& label) {
     color_close();
   }
 
-  if (!label->m_underline.has_color() || (label->m_margin.right.value > 0 && m_tags[syntaxtag::u] > 0)) {
+  if (label->m_underline.has_color()) {
     underline_close();
   }
   if (!label->m_overline.has_color() || (label->m_margin.right.value > 0 && m_tags[syntaxtag::o] > 0)) {
--- a/src/components/builder.cpp
+++ b/src/components/builder.cpp
@@ -164,7 +164,7 @@ void builder::node(const label_t& label) {
   if (label->m_underline.has_color()) {
     underline_close();
   }
-  if (!label->m_overline.has_color() || (label->m_margin.right.value > 0 && m_tags[syntaxtag::o] > 0)) {
+  if (label->m_overline.has_color()) {
     overline_close();
   }
 
--- a/src/components/builder.cpp
+++ b/src/components/builder.cpp
@@ -7,6 +7,7 @@
 #include "utils/color.hpp"
 #include "utils/string.hpp"
 #include "utils/time.hpp"
+#include "utils/unit.hpp"
 POLYBAR_NS
 
 using namespace tags;
@@ -125,7 +126,7 @@ void builder::node(const label_t& label) {
 
   auto text = label->get();
 
-  if (label->m_margin.left > 0) {
+  if (label->m_margin.left.value > 0) {
     space(label->m_margin.left);
   }
 
@@ -143,13 +144,13 @@ void builder::node(const label_t& label) {
     color(label->m_foreground);
   }
 
-  if (label->m_padding.left > 0) {
+  if (label->m_padding.left.value > 0) {
     space(label->m_padding.left);
   }
 
   node(text, label->m_font);
 
-  if (label->m_padding.right > 0) {
+  if (label->m_padding.right.value > 0) {
     space(label->m_padding.right);
   }
 
@@ -160,14 +161,14 @@ void builder::node(const label_t& label) {
     color_close();
   }
 
-  if (label->m_underline.has_color()) {
+  if (!label->m_underline.has_color() || (label->m_margin.right.value > 0 && m_tags[syntaxtag::u] > 0)) {
     underline_close();
   }
-  if (label->m_overline.has_color()) {
+  if (!label->m_overline.has_color() || (label->m_margin.right.value > 0 && m_tags[syntaxtag::o] > 0)) {
     overline_close();
   }
 
-  if (label->m_margin.right > 0) {
+  if (label->m_margin.right.value > 0) {
     space(label->m_margin.right);
   }
 }
@@ -202,25 +203,27 @@ void builder::node_repeat(const label_t& label, size_t n) {
 /**
  * Insert tag that will offset the contents by given pixels
  */
-void builder::offset(int pixels) {
-  if (pixels == 0) {
+void builder::offset(geometry pixels) {
+  if (pixels.value == 0) {
     return;
   }
-  tag_open(syntaxtag::O, to_string(pixels));
+  tag_open(syntaxtag::O, unit_utils::geometry_to_string(pixels));
 }
 
 /**
  * Insert spaces
  */
-void builder::space(size_t width) {
-  if (width) {
-    m_output.append(width, ' ');
+void builder::space(space_size size) {
+  if (size.value > 0.) {
+    m_output += add_surrounding_tag(size);
   } else {
     space();
   }
 }
 void builder::space() {
-  m_output.append(m_bar.spacing, ' ');
+  if (m_bar.spacing.value > 0) {
+    space(m_bar.spacing);
+  }
 }
 
 /**
@@ -234,7 +237,7 @@ void builder::remove_trailing_space(size_t len) {
   }
 }
 void builder::remove_trailing_space() {
-  remove_trailing_space(m_bar.spacing);
+  remove_trailing_space(static_cast<size_t>(m_bar.spacing.value));
 }
 
 /**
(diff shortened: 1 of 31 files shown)


patrick96:
Awesome! Thanks for putting in all this effort :partying_face: 

The unit tests don't compile yet, probably because the dispatch test still uses the old offsets.

I have some time to review this, this afternoon. I'm very excited about this, this will likely be the biggest change in the next version (unless we can get #1752 in, but I think that would take too long).

frebib:
Thanks for fixing up the tests. I had a go at fixing them but ran into weirdness I didn't understand.

If there's anything else you'd like me to change please holla

patrick96:
Yeah, the tests took some time because the equality operator for `extent_val` needed to be defined in a very particular way.

I'm writing up the review right now, there are only a few things left, I think.

src/components/builder.cpp
```diff
@@ -160,15 +161,15 @@ void builder::node(const label_t& label) {
     color_close();
   }
 
-  if (label->m_underline.has_color()) {
+  if (!label->m_underline.has_color() || (label->m_margin.right.value > 0 && m_tags[syntaxtag::u] > 0)) {
```
patrick96, line 164:
```suggestion
  if (label->m_underline.has_color()) {
```

I don't really know what the second part is supposed to do.

src/components/builder.cpp
```diff
@@ -160,15 +161,15 @@ void builder::node(const label_t& label) {
...
     underline_close();
   }
-  if (label->m_overline.has_color()) {
+  if (!label->m_overline.has_color() || (label->m_margin.right.value > 0 && m_tags[syntaxtag::o] > 0)) {
```
patrick96, line 167:
```suggestion
  if (label->m_overline.has_color()) {
```

src/components/config.cpp
```diff
@@ -208,6 +211,59 @@ unsigned long long config::convert(string&& value) const {
   return v < ULLONG_MAX ? v : 0ULL;
 }
 
+template <>
+spacing_val config::convert(string&& value) const {
+  char* new_end;
+  auto size_value = std::strtof(value.c_str(), &new_end);
```
patrick96, line 217: Also here, we should better use `std::stof`

src/components/config.cpp
```diff
@@ -208,6 +211,59 @@ unsigned long long config::convert(string&& value) const {
...
+
+  string unit = string_util::trim(new_end);
+  if (!unit.empty()) {
+    if (unit == "px") {
+      size.type = spacing_type::PIXEL;
+      size.value = std::trunc(size.value);
+    } else if (unit == "pt") {
+      size.type = spacing_type::POINT;
+    } else {
+      size.value = std::trunc(size.value);
```
patrick96, line 233: This branch should produce a `value_error` since something like `123foo` is invalid and should not be interpreted as 123 spaces.

src/components/config.cpp
```diff
@@ -208,6 +211,59 @@ unsigned long long config::convert(string&& value) const {
...
+    }
+  }
+
+  return size;
```
patrick96, line 237: If we have spaces, the number of spaces is not truncated and may be a floating point number.

src/components/config.cpp
```diff
@@ -208,6 +211,59 @@ unsigned long long config::convert(string&& value) const {
...
+ * The new format is X%:Z, where X is in [0, 100], and Z is any real value
+ * describing a pixel offset. The actual value is calculated by X% * max + Z
+ */
+template <>
+percentage_with_offset config::convert(string&& value) const {
+  size_t i = value.find(':');
+
+  if (i == std::string::npos) {
+    if (value.find('%') != std::string::npos) {
+      return {strtod(value.c_str(), nullptr), {}};
```
patrick96, line 257:
```suggestion
      return {std::stoi(value), {}};
```

src/components/config.cpp
```diff
@@ -208,6 +211,59 @@ unsigned long long config::convert(string&& value) const {
...
+    } else {
+      return {0., convert<extent_val>(move(value))};
+    }
+  } else {
+    std::string percentage = value.substr(0, i - 1);
+    return {strtod(percentage.c_str(), nullptr), convert<extent_val>(value.substr(i + 1))};
```
patrick96, line 263:
```suggestion
    return {std::stoi(percentage), convert<extent_val>(value.substr(i + 1))};
```

src/drawtypes/label.cpp
```diff
@@ -183,14 +183,23 @@ namespace drawtypes {
...
+      const auto parse_or_throw = [&](const string& key, spacing_val default_value) {
+        try {
+          return conf.get(section, key, default_value);
+        } catch (const std::exception& err) {
+          throw application_error(
+              sstream() << "Failed to set " << section << "." << key << " (reason: " << err.what() << ")");
+        }
+      };
+
+      auto value = parse_or_throw(key, spacing_val{});
```
patrick96, line 199:
```suggestion
      auto value = parse_or_throw(key, ZERO_SPACE);
```

patrick96 (changes requested):
I went through all my reviews in the old PR and tried to address them here. I also ran `clang-format` over all changed files.

There are still some things that need to be done:

* We need to decide whether we want to allow negative spacing
* We (I) need to address the weirdness in `base.inl`: https://github.com/polybar/polybar/pull/1671#discussion_r369192910
* We need to go through all config options that use extents or percentage_with_offset and see how they deal with negative values (and if they're even allowed)

src/components/config.cpp
frebib: This seems to produce a warning
```
[ 41%] Building CXX object bin/CMakeFiles/poly.dir/modules/ipc.cpp.o
/home/frebib/nerdcloud/projects/frebib/polybar/src/components/config.cpp: In member function ‘T polybar::config::convert(std::string&&) const [with T = polybar::percentage_with_offset; std::string = std::__cxx11::basic_string<char>]’:
/home/frebib/nerdcloud/projects/frebib/polybar/src/components/config.cpp:257:24: warning: narrowing conversion of ‘std::__cxx11::stoi((*(const string*)(& value)), 0, 10)’ from ‘int’ to ‘double’ [-Wnarrowing]
  257 |       return {std::stoi(value), {}};
      |               ~~~~~~~~~^~~~~~~
/home/frebib/nerdcloud/projects/frebib/polybar/src/components/config.cpp:263:22: warning: narrowing conversion of ‘std::__cxx11::stoi(percentage, 0, 10)’ from ‘int’ to ‘double’ [-Wnarrowing]
  263 |     return {std::stoi(percentage), convert<extent_val>(value.substr(i + 1))};
      |             ~~~~~~~~~^~~~~~~~~~~~
```

patrick96: I think that was just a brainfart on my side. I wanted to use the c++ stdlib functions for number parsing but probably didn't see that the original produced a double. It should be `std::stod`, I'll change that.

mainrs:
Is it possible for non-write access users to even see a merge conflict? I wanted to help out but there doesn’t seem to be a way to see the conflicts.

patrick96:
@mainrs If you don't see them, I guess non-members can't see them in the UI. If you really want to see the conflicts, you can clone frebib's repo locally and try to merge into upstream master.

However, the merge conflicts are not a blocker here. There are some decisions I have to make with regards to semantics, as I mentioned in my review: https://github.com/polybar/polybar/pull/2394#pullrequestreview-602364135
This is in the 3.6.0 milestone and I definitely plan to have this in the next release. I will try to work on this after I have finished #2539.

I appreciate your initiative though :)

There are a bunch of issues tagged with ["help wanted"](https://github.com/polybar/polybar/labels/help%20wanted) that are worth looking at.
From a quick glance, some that might be interesting and easier for first time contributors are:

* #1630 
* #2543 
* #675 
* #2456 

The last one will probably touch quite a few components and may require a lot more time than the other three.

PR #2404, opened 2021-03-26
Add wm-restack=generic option that lowers polybar to the bottom of the stack

### LINKED ISSUE #1657: Set bar to bottom of window stack
The problem is that a window that's under a transparent bar is not visible through the transparent bar. To solve this problem AND just because I think it would look cool, I would like to have the option to stack the bar to the bottom of the window stack. Just like is possible with tint2.

I don't know exactly how many people would like to see this feature. But if I think of all the r/unixporn people I've seen that generate a background image containing status-like information every second, I guess more than just me!

Solution: option like "panel_layer = bottom || top" in the bar config.
Technically, I wouldn't know how to implement this exactly. It might have something to do with "bar::restack_window()" in "src/components/bar.cpp"? XCB_CONFIGURE_WINDOW lets you set the window stacking order. Again, I don't really know.

polybar
tint2

### LINKED ISSUE #2205: Generic/Xmonad-specific window restacker
I'm using xmonad as my wm.
In some use cases, starting polybar after xmonad put polybar above windows, even with the correct configuration on both programs (without xmonad ``avoidStruts`` hook).
In xmobar, the ``option lowerOnStart = True`` is used to fix this.

Solution: Add `lowerOnStart` options support to polybar, which, if set `true`, will start polybar on the bottom of window stack.

ArchLinux
polybar 3.4.3

timakro:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

Previously wm-restack only supported bspwm and i3. Both have a special
top-level window that polybar detects and places itself directly above.
This patch adds wm-restack=generic which simply lowers polybar to the
very bottom of the stack. This option was tested and confirmed to work
with xmonad which doesn't have a special top-level window and therefore
doesn't require special handling like bspwm and i3.

## Related Issues & Documents

Fixes #2205
Closes #1657 

I suggest the following changes to the configuration page on the wiki:

```
; Restack the bar window. Fixes the issue where the
; bar is being drawn on top of fullscreen windows.
;
; Currently supported options:
;   generic (works in xmonad, may work with other WMs)
;   bspwm
;   i3 (requires: `override-redirect = true`)
; wm-restack =
```

### DIFF
--- a/src/components/bar.cpp
+++ b/src/components/bar.cpp
@@ -468,7 +468,7 @@ void bar::restack_window() {
       }
       restacked = true;
     } catch (const exception& err) {
-      m_log.err("Failed to restack bar window (err=%s", err.what());
+      m_log.err("Failed to restack bar window (err=%s)", err.what());
     }
   } else if (wm_restack == "bspwm") {
     restacked = bspwm_util::restack_to_root(m_connection, m_opts.monitor, m_opts.window);
--- a/src/components/bar.cpp
+++ b/src/components/bar.cpp
@@ -458,7 +458,19 @@ void bar::restack_window() {
 
   auto restacked = false;
 
-  if (wm_restack == "bspwm") {
+  if (wm_restack == "generic") {
+    try {
+      auto children = m_connection.query_tree(m_connection.screen()->root).children();
+      if (children.begin() != children.end() && *children.begin() != m_opts.window) {
+        const unsigned int value_mask = XCB_CONFIG_WINDOW_SIBLING | XCB_CONFIG_WINDOW_STACK_MODE;
+        const unsigned int value_list[2]{*children.begin(), XCB_STACK_MODE_BELOW};
+        m_connection.configure_window_checked(m_opts.window, value_mask, value_list);
+      }
+      restacked = true;
+    } catch (const exception& err) {
+      m_log.err("Failed to restack bar window (err=%s", err.what());
+    }
+  } else if (wm_restack == "bspwm") {
     restacked = bspwm_util::restack_to_root(m_connection, m_opts.monitor, m_opts.window);
 #if ENABLE_I3
   } else if (wm_restack == "i3" && m_opts.override_redirect) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   windows per workspace
   ([`#604`](https://github.com/polybar/polybar/issues/604))
 - `internal/backlight`: added `use-actual-brightness` option
+- Added `wm-restack = generic` option that lowers polybar to the bottom of the stack.
+  Fixes the issue where the bar is being drawn on top of fullscreen windows in xmonad.
+  ([`#2205`](https://github.com/polybar/polybar/issues/2205))
 
 ### Changed
 - Slight changes to the value ranges the different ramp levels are responsible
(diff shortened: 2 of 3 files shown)


src/components/bar.cpp
```diff
@@ -458,7 +458,16 @@ void bar::restack_window() {
 
   auto restacked = false;
 
-  if (wm_restack == "bspwm") {
+  if (wm_restack == "generic") {
+    auto children = m_connection.query_tree(m_connection.screen()->root).children();
+    auto lowest = *children.begin();
```
patrick96, line 463: This may be a bit pedantic, but could chidlren ever be empty?

In theory, this shouldn't happen because polybar is a child of the root, so the root must have at least one child.
I'm just wondering if there can't be some kind of weirdness.

src/components/bar.cpp
```diff
@@ -458,7 +458,16 @@ void bar::restack_window() {
...
+    if (lowest != m_opts.window) {
+      const unsigned int value_mask = XCB_CONFIG_WINDOW_SIBLING | XCB_CONFIG_WINDOW_STACK_MODE;
+      const unsigned int value_list[2]{lowest, XCB_STACK_MODE_BELOW};
+      m_connection.configure_window_checked(m_opts.window, value_mask, value_list);
```
patrick96, line 467: This can throw an exception which would crash the bar. Please add a try-catch block that prints an error (and doesn't set `restacked`)

patrick96 (changes requested):
Awesome! This also implemented #1657 I think.

Everything seems to work from my testing and looks like this should work in most WMs. It even works in i3 even though there the polybar window isn't always a direct child of the root.

This just needs one small change and then we're good to go.
Thanks a lot :tada:

src/components/bar.cpp
```diff
@@ -458,7 +458,19 @@ void bar::restack_window() {
...
+    try {
+      auto children = m_connection.query_tree(m_connection.screen()->root).children();
+      if (children.begin() != children.end() && *children.begin() != m_opts.window) {
+        const unsigned int value_mask = XCB_CONFIG_WINDOW_SIBLING | XCB_CONFIG_WINDOW_STACK_MODE;
+        const unsigned int value_list[2]{*children.begin(), XCB_STACK_MODE_BELOW};
+        m_connection.configure_window_checked(m_opts.window, value_mask, value_list);
+      }
+      restacked = true;
+    } catch (const exception& err) {
+      m_log.err("Failed to restack bar window (err=%s", err.what());
```
patrick96, line 471:
```suggestion
      m_log.err("Failed to restack bar window (err=%s)", err.what());
```

patrick96 (approved):

patrick96:
Thanks!

ixzh:
hi can I ask a question: 
with wm-stack=generic, polybar still shown over fullscreen app, I need to open a new window in the current layout, then fullscreen works. I'm using xmonad on arch.

patrick96:
@timakro Can you check this out?

patrick96:
@ixzh Just to check, does polybar print any errors and are you using the latest polybar development version?

ixzh:
> are you using the latest polybar development version

no error message...I'm using polybar 3.5.6.

patrick96:
polybar 3.5.6 doesn't have `wm-restack = generic`, that's an unreleased feature.

ixzh:
> polybar 3.5.6 doesn't have `wm-restack = generic`, that's an unreleased feature.

Ok run from commandline, sawthe warning:
```
warn: Ignoring unsupported wm-restack option 'generic'
```
switched to polybar-git version, fullscreen works, but the following error was given:
```
error: Invalid value for "bar/mainbar-xmonad.tray-background", using default value (reason: "#zz24283b" is an invalid color value.)
```
transparent bg color not supported?

PR #2413, opened 2021-04-10
do not show to long string for keyboard layout name

Dushistov:

* [ ] Refactor
* [x] Feature
* [x] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

In my case layout name "macintosh_vndr/gb". It is too long,
"gb" is enough.

### DIFF
--- a/src/modules/xkeyboard.cpp
+++ b/src/modules/xkeyboard.cpp
@@ -123,7 +123,11 @@ namespace modules {
       m_layout->replace_token("%name%", m_keyboard->group_name(m_keyboard->current()));
       m_layout->replace_token("%variant%", m_keyboard->variant_name(m_keyboard->current()));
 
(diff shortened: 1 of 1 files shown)


parmort:
Perhaps a better way of doing this could be to add a configuration option for the xkeyboard module, instead of universally truncating the layout name. Some people might need/want to see the entire layout name, and committing this would break their configurations.

patrick96:
I agree with @parmort, we shouldn't change existing behavior.

I don't know much about how the different parts of the keyboard layout name are structured and if there is even a defined structure. Maybe you have some more insight here than me.
But I would tend to prefer adding a new token for whatever we would call the thing after the slash instead of controlling it through a new config option.

@Dushistov let me know if you would want to work on something like that. If so, we would need to find a name for the thing after the slash and introduce a new token that contains everything after the slash.

patrick96:
Closing due to inactivity

PR #2429, opened 2021-05-12
The xworkspaces module should wait for EWMH to become available

### LINKED ISSUE #1915: When polybar starts with WM, workspace list is empty
## Describe the issue
I'm using xmonad, and starting `polybar` directly from xmonad when it starts up. This used to work fine, but recently (I suspect after https://github.com/polybar/polybar/pull/882) my list of workspaces is now shown as entirely empty. If I restart `polybar` shortly after, the workspaces appear as normal.

**Was it working before?**
I never experienced this issue with my previous build, which was gfe7dc28b, but am seeing it now with g2b31a3f1.

## To Reproduce
It is surprisingly hard to reproduce this, as it only happens very occasionally at boot. I _believe_ this config is sufficient to reproduce, but have yet to confirm. Otherwise, my full config is [this one](https://github.com/jonhoo/configs/blob/b6d4aee616d234f581766b467a0e772e775cc53f/gui/.config/polybar/config). Note that I _do_ overwrite `monitor`, but not sure whether that would affect any of this.
```dosini
[bar/main]
modules-left = ewmh

[module/ewmh]
type = internal/xworkspaces
pin-workspaces = false

[settings]
screenchange-reload = true
```

Essentially, have your WM run `polybar` on boot with:
```console
$ env "MONITOR=$m" polybar --reload main
```
where `$m` is the monitor you want to use (again, not sure if the monitor part is relevant, but including it for completeness). Reboot and observe the bar. Every few boots, the list of workspaces will be empty. If you then restart polybar, it will be populated again.

## Polybar Log

I forgot to copy the log last time this happened, and am having a hard time reproducing it reliably. Will post the log next time it happens.

## Environment:
* WM: xmonad
* Distro: Arch Linux
* Output of `polybar -vvv`:
```
polybar 3.4.0-24-g2b31a3f1

Build type: Release
Compiler: /usr/bin/c++
```

timakro:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

Before this change if EWMH wasn't available the xworkspaces module was
permanently disabled. When polybar was started alongside the window
manager e.g. from .xinitrc this caused a race condition between polybar
and the window manager and the xworkspaces module may or may not be
displayed. After this change polybar will wait for EWMH to become
available. This change closes #1915, see that issue for more details.

Curiously this only required the removal of the error condition which
used the be raised when EWMH wasn't available. The xworkspaces module
will show up on the bar as soon as the first EWMH event is processed by
the existing event handling code. I can't argue much about the
correctness of this patch but it seems to work flawlessly in my testing
with xmonad. I didn't test any other window managers. Note that removing
the error condition below which checks that _NET_DESKTOP_VIEWPORT is
available might make this work with pin-workspaces=true. I couldn't test
the effects of that change because I only tested with xmonad and xmonad
doesn't support _NET_DESKTOP_VIEWPORT, so I didn't make that change.

## Related Issues & Documents

Closes #1915

### DIFF
--- a/include/modules/xworkspaces.hpp
+++ b/include/modules/xworkspaces.hpp
@@ -99,7 +99,6 @@ namespace modules {
     vector<string> m_desktop_names;
     vector<bool> m_urgent_desktops;
     unsigned int m_current_desktop;
-    string m_current_desktop_name;
 
     /**
      * Maps an xcb window to its desktop number
--- a/src/modules/xworkspaces.cpp
+++ b/src/modules/xworkspaces.cpp
@@ -50,11 +50,6 @@ namespace modules {
       throw module_error("Failed to initialize ewmh atoms");
     }
 
-    // Check if the WM supports _NET_CURRENT_DESKTOP
-    if (!ewmh_util::supports(m_ewmh->_NET_CURRENT_DESKTOP)) {
-      throw module_error("The WM does not support _NET_CURRENT_DESKTOP, aborting...");
-    }
-
     // Check if the WM supports _NET_DESKTOP_VIEWPORT
     if (!(m_monitorsupport = ewmh_util::supports(m_ewmh->_NET_DESKTOP_VIEWPORT)) && m_pinworkspaces) {
       throw module_error("The WM does not support _NET_DESKTOP_VIEWPORT (required when `pin-workspaces = true`)");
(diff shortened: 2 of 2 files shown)


patrick96:
Oh wow, I didn't think it would be this easy, but it makes sense

Thank you very much :smiley: 

The only question now is how to deal with window managers that don't actually support the features that you need. For example window managers that don't support `_NET_CURRENT_DESKTOP` or `_NET_DESKTOP_VIEWPORT`.

For the first property, I think the module would just run but never display anything and wouldn't produce an error. For `_NET_DESKTOP_VIEWPORT`, I'm not totally sure what happens. My guess is that it just displays all tags and doesn't produce an error either.

I need to do some testing on this, but I don't really have time right now. Will look at it at some point in the future.

PR #2463, opened 2021-07-11
feat(ipc): allow receiving arbitrary text on IPC socket

### LINKED ISSUE #2455: Make IPC module accept arbitrary text over the socket
Currently, to extend polybar, we can either use the script module or the IPC module. The script module allows to do anything, but you have to spawn a new process. With the `tail` directive, it's possible to avoid spawning processes at a regular interval. But, a distinct process is still needed (and it eats some memory). With the IPC module, you can trigger an update from another process, but this will run a script to make the update (and it eats CPU and battery).

Something between the two could be useful: extend the IPC module to allow sending arbitrary text to be displayed instead of just a hook number.

I am willing to implement this, but as there is already some issues around the IPC module, I would like to gather some feedback and agreement around such a feature.

See #1127 and #1670 which are somewhat related.

vincentbernat:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
Instead of just allowing hook numbers to be executed, the user can
send arbitrary text and the IPC module will put it in the bar. The IPC
payload format is extended to accept an arbitrary string if the first
character after the module name is ':'.

    polybar-msg hook test :'%{F#4444ff}hello%{F-}'

I did not make a new module as requested by @patrick96 as the modifications to the IPC module are minimal and the actions are reused as is. Tell me what you think.

## Related Issues & Documents
Fix #2455

### DIFF
--- a/src/modules/ipc.cpp
+++ b/src/modules/ipc.cpp
@@ -15,12 +15,12 @@ namespace modules {
   ipc_module::ipc_module(const bar_settings& bar, string name_) : static_module<ipc_module>(bar, move(name_)) {
     size_t index = 0;
 
-    for (auto&& command : m_conf.get_list<string>(name(), "hook")) {
-      m_hooks.emplace_back(std::make_unique<hook>(hook{name() + to_string(++index), command}));
-    }
-
-    if (m_hooks.empty()) {
-      throw module_error("No hooks defined");
+    try {
+      for (auto&& command : m_conf.get_list<string>(name(), "hook")) {
+        m_hooks.emplace_back(std::make_unique<hook>(hook{name() + to_string(++index), command}));
+      }
+    } catch (const key_error& err) {
+      m_log.notice("%s: no hooks defined", name());
     }
 
     if ((m_initial = m_conf.get(name(), "initial", 0_z)) && m_initial > m_hooks.size()) {
@@ -122,6 +122,14 @@ namespace modules {
         m_output.clear();
       }
 
+      broadcast();
+      return;
+    }
+    if (message.rfind(name() + ":", 0) == 0) {
+      m_log.info("%s: Received generic payload (%s)", name(), message);
+      string payload{message.substr(name().length() + 1)};
+      m_output.clear();
+      m_output = payload;
       broadcast();
     }
   }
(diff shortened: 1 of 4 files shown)


patrick96 (approved):
Thank you very much, these changes are indeed very minimal :open_mouth: 

I think the `polybar-msg hook` command is becoming too limited for the ipc module. We have a very capable module action system that would be better suited. It allows us to have named actions per module with optional data.
For example here, we could add a `send` action whose data will be set as the module's content.

If your ipc module is named `ipc`, this could look like this:

```
polybar-msg -p ... action "#ipc.send.abc"
```

The big advantage of this is that this also works for click commands: `click-left = #ipc.send.abc` will deliver this to the correct module on the clicked bar.
I have added those changes locally, I will push them shortly.

src/modules/ipc.cpp
```diff
@@ -15,12 +15,12 @@ namespace modules {
   ipc_module::ipc_module(const bar_settings& bar, string name_) : static_module<ipc_module>(bar, move(name_)) {
     size_t index = 0;
 
-    for (auto&& command : m_conf.get_list<string>(name(), "hook")) {
-      m_hooks.emplace_back(std::make_unique<hook>(hook{name() + to_string(++index), command}));
-    }
-
-    if (m_hooks.empty()) {
-      throw module_error("No hooks defined");
+    try {
```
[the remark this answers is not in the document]
vincentbernat: OK, updated.

src/modules/ipc.cpp
```diff
@@ -123,8 +123,14 @@ namespace modules {
       }
 
       broadcast();
+      return;
```
patrick96, line 126:
```suggestion
```

vincentbernat: This could be left as an optimization? Once it matches the first hook, it cannot match the others.

vincentbernat:
Just to show an example, here is a piece of code to display Network Manager state: https://github.com/vincentbernat/i3wm-configuration/blob/7df5011824cc55ffa9d3ac8b32bb680d63ac083d/bin/i3-companion#L454.

```dosini
[module/network]
type = custom/ipc
hook-0 = cat $XDG_RUNTIME_DIR/i3/network.txt
initial = 1
```

PR #2475, opened 2021-07-28
Added support for hook cycling using prev and next

ijcon40:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Related Issues & Documents

https://github.com/polybar/polybar/issues/2464
https://github.com/polybar/polybar/issues/999

It would be a good idea to include examples of using the cycling and current tokens in the ipc module example config.

### DIFF
--- a/include/modules/ipc.hpp
+++ b/include/modules/ipc.hpp
@@ -3,6 +3,9 @@
 #include "modules/meta/static_module.hpp"
 #include "utils/command.hpp"
 
+#include <mutex>
+
+
 POLYBAR_NS
 
 namespace modules {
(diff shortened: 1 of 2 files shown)


patrick96 (changes requested):
Sorry for the delay and thanks for contributing :)

Since the ipc module now supports [actions](https://polybar.readthedocs.io/en/stable/user/actions.html), it would be better to directly expose `prev` and `next` actions from the module instead of giving the user `%next%` and `%prev%` tokens that they then have to use in `click-left = polybar-msg -p %pid% hook demo %next%`.

When we have actions, we can just do `click-left = #demo.next`

patrick96:
Made obsolete by #2528

PR #2485, opened 2021-08-28
fix(monitor): do not include outputs when monitors are supported

### LINKED ISSUE #2481: `polybar -m` shows physical output when split in two monitors
## Checklist

* [x] I have read the appropriate section in the [contributing
  guidelines](https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md)
* [x] I believe this issue is a problem with polybar itself and not a misconfiguration on my part.
* [x] I have searched for other open and closed [issues](https://github.com/polybar/polybar/issues?q=is%3Aissue) that
  may have already reported this problem.
* [x] I have checked the [known issues](https://github.com/polybar/polybar/wiki/Known-Issues) page for this problem.

## Describe the bug
When running `polybar -m` when using XRandR 1.5 monitors, we get both the monitors and physical outputs. I was wondering if this was done on purpose. I could provide a patch if you think I am right.

**Expected behavior:**
The physical outputs should be hidden if they are covered by monitors.

**Actual behavior:**
The physical outputs are kept in the list.

## To Reproduce

```console
$ xrandr --setmonitor '*'DisplayPort-3-left  1920/444x1440/334+0+0     DisplayPort-3
$ xrandr --setmonitor    DisplayPort-3-right 1520/352x1440/334+1920+0  none
$ xrandr --noprimary
$ polybar -m
DisplayPort-3-left: 1920x1440+0+0 (XRandR monitor, primary)
DisplayPort-3-right: 1520x1440+1920+0 (XRandR monitor)
DisplayPort-3: 3440x1440+0+0
```

## Environment:
* WM: i3
* Distro: Debian unstable
* Output of `polybar -vvv`:

```
polybar 3.5.6-129-gdd6637b39a14

Build type: Release
Compiler: /usr/lib/ccache/c++
Compiler flags:  -O3 -DNDEBUG -Wall -Wextra -Wpedantic -Wsuggest-override
Linker flags: -Wall -Wextra -Wpedantic -Wsuggest-override  -Wall -Wextra -Wpedantic -Wsuggest-override
```

vincentbernat:
Previously, when splitting an output into two monitors, `polybar -m`
would report both the splitted monitors and the output. This was not
caught by the the clone detection as the detection works by removing
monitors contained into another monitors (and monitors were excluded
from that logic) and we want the other way around: outputs covered by
monitors should be ignored.

Instead of trying to detect covered outputs, the solution is quite
simple: when monitors are supported, do not consider outputs. A
monitor can be set primary (and RandR reports primary outputs as
primary monitors). The only information we would miss from monitors
are things like refresh rate and EDID. We don't need that, so we are
fine.

As monitors are only created for connected output (and they are in
this case "active") or disconnected output if they are mapped (and
they are in this case "inactive"), I am a bit unsure if we have
exactly the same behaviour as previously when `connected_only` is set
to `false`.

As some modules require an output, we keep the output in the
`monitor_t` structure and we ensure it is correctly set when using
monitors. A monitor can have 0 or several outputs. We only handle the
0 and 1 cases. When a monitor has more outputs, only the first one is
used. AFAIK, only the xbacklight module needs that and I think we are
fine waiting for a user needing this module and merging monitors.

This last part is buggy. The C++ binding fail to expose the
`outputs()` method to iterate over the outputs of a monitor. This
seems to be a bug in XPP. The field is correctly defined in the RandR
XML file and it works with the Python binding. I attempt to fallback
to C, but the C header does not have room to include this information
and the C++ binding may make a copy, losing the information.

```xml
	<struct name="MonitorInfo">
		<field type="ATOM" name="name" />
		<field type="BOOL" name="primary" />
		<field type="BOOL" name="automatic" />
		<field type="CARD16" name="nOutput" />
		<field type="INT16" name="x" />
		<field type="INT16" name="y" />
		<field type="CARD16" name="width" /> <!-- pixels -->
		<field type="CARD16" name="height" /> <!-- pixels -->
		<field type="CARD32" name="width_in_millimeters" />
		<field type="CARD32" name="height_in_millimeters" />
		<list type="OUTPUT" name="outputs">
		    <fieldref>nOutput</fieldref>
		</list>
	</struct>
```

When not buggy, we shouldn't get "XRandR monitor" in the output of
`polybar -m` unless there is no output attached to the monitor.

Fixes #2481

* [ ] Refactor
* [ ] Feature
* [x] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -126,6 +126,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([`#2367`](https://github.com/polybar/polybar/issues/2367))
 - Warning message regarding T@ in bspwm module
   ([`#2371`](https://github.com/polybar/polybar/issues/2371))
+- `polybar -m` hides physical outputs covered by monitors
+  ([`#2481`](https://github.com/polybar/polybar/issues/2481))
 
 ## [3.5.6] - 2021-05-24
 ### Build
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -126,7 +126,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([`#2367`](https://github.com/polybar/polybar/issues/2367))
 - Warning message regarding T@ in bspwm module
   ([`#2371`](https://github.com/polybar/polybar/issues/2371))
-- `polybar -m` hides physical outputs covered by monitors
+- `polybar -m` used to show both physical outputs and randr monitors, even if the outputs were covered by monitors.
   ([`#2481`](https://github.com/polybar/polybar/issues/2481))
 
 ## [3.5.6] - 2021-05-24
--- a/src/x11/extensions/randr.cpp
+++ b/src/x11/extensions/randr.cpp
@@ -107,56 +107,65 @@ namespace randr_util {
    */
   vector<monitor_t> get_monitors(connection& conn, xcb_window_t root, bool connected_only, bool purge_clones) {
     vector<monitor_t> monitors;
+    bool found = false;
 
-#if WITH_XRANDR_MONITORS
     if (check_monitor_support()) {
-      for (auto&& mon : conn.get_monitors(root, true).monitors()) {
+#if WITH_XRANDR_MONITORS
+      /* Use C, because XPP does not provide access to output info from monitors. */
+      xcb_generic_error_t *err;
+      auto rrmonitors =
+        xcb_randr_get_monitors_reply(
+          conn, xcb_randr_get_monitors(conn, root, true), &err);
+      if (err != NULL) {
+        free(err);
+      } else {
+        for (auto iter = xcb_randr_get_monitors_monitors_iterator(rrmonitors);
+             iter.rem;
+             xcb_randr_monitor_info_next(&iter)) {
+          auto mon = iter.data;
+          auto name = conn.get_atom_name(mon->name).name();
+
+          /* Output is only useful for some plugins. We take the first one. */
+          int randr_output_len = xcb_randr_monitor_info_outputs_length(mon);
+          auto randr_outputs = xcb_randr_monitor_info_outputs(mon);
+          auto output = (randr_output_len >= 1) ? randr_outputs[0] : XCB_NONE;
+          monitors.emplace_back(make_monitor(output, move(name), mon->width, mon->height, mon->x, mon->y, mon->primary));
+          found = true;
+        }
+        free(rrmonitors);
+      }
+#endif
+    }
+    if (!found || !purge_clones) {
+      auto primary_output = conn.get_output_primary(root).output();
+      string primary_name{};
+
+      if (primary_output != XCB_NONE) {
+        auto primary_info = conn.get_output_info(primary_output);
+        auto name_iter = primary_info.name();
+        primary_name = {name_iter.begin(), name_iter.end()};
+      }
+
+      for (auto&& output : conn.get_screen_resources(root).outputs()) {
         try {
-          auto name = conn.get_atom_name(mon.name).name();
-          monitors.emplace_back(make_monitor(XCB_NONE, move(name), mon.width, mon.height, mon.x, mon.y, mon.primary));
+          auto info = conn.get_output_info(output);
+          if (info->crtc == XCB_NONE) {
+            continue;
+          } else if (connected_only && info->connection != XCB_RANDR_CONNECTION_CONNECTED) {
+            continue;
+          }
+
+          auto name_iter = info.name();
+          string name{name_iter.begin(), name_iter.end()};
+
+          auto crtc = conn.get_crtc_info(info->crtc);
+          auto primary = (primary_name == name);
+          if (!std::any_of(monitors.begin(), monitors.end(), [name](const auto &monitor) { return monitor->name == name; })) {
+            monitors.emplace_back(make_monitor(output, move(name), crtc->width, crtc->height, crtc->x, crtc->y, primary));
+          }
         } catch (const exception&) {
           // silently ignore output
         }
-      }
-    }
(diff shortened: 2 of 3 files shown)


vincentbernat:
Unfortunately, I don't know how hard this is to fix the remaining issue (in XPP, I think).

src/x11/extensions/randr.cpp
```diff
@@ -108,55 +108,52 @@ namespace randr_util {
...
 
-#if WITH_XRANDR_MONITORS
     if (check_monitor_support()) {
-      for (auto&& mon : conn.get_monitors(root, true).monitors()) {
+#if WITH_XRANDR_MONITORS
+      for (auto&& mon : conn.get_monitors(root, connected_only).monitors()) {
         try {
           auto name = conn.get_atom_name(mon.name).name();
-          monitors.emplace_back(make_monitor(XCB_NONE, move(name), mon.width, mon.height, mon.x, mon.y, mon.primary));
+          /* TODO: not reliable. The C++ binding does not expose
```
patrick96, line 116: Yeah, I couldn't figure out why xpp doesn't generate this. Though I really don't much experience with xpp internals.

We may not get around rewriting this whole loop using the C API

patrick96 (changes requested):
Nice work!

One thing: I think people should still be able to select xrandr outputs in polybar's `monitor` setting.

> As monitors are only created for connected output (and they are in
this case "active") or disconnected output if they are mapped (and
they are in this case "inactive"), I am a bit unsure if we have
exactly the same behaviour as previously when connected_only is set
to false.

I think this should be fine. The `connected_only` is really only needed for people where polybar for some reason detects a monitor as disconnected (even though it isn't).

> When a monitor has more outputs, only the first one is
used. AFAIK, only the xbacklight module needs that and I think we are
fine waiting for a user needing this module and merging monitors.

I agree.

src/x11/extensions/randr.cpp
vincentbernat: Do you mean that we should get around by rewriting the whole loop using the C API? In this case, I can do that.

patrick96: Yes, that's what I meant. I didn't really formulate it directly enough :smile:

vincentbernat: OK, I have pushed that in a separate commit. I can squash them if you prefer. Be aware that I almost never do C++, so I have put auto everywhere and it compiles dans doesn't crash, so I think I am good.

patrick96: > OK, I have pushed that in a separate commit. I can squash them if you prefer.

No, this is fine. We can squash-merge the PR if things get out of hand.

> Be aware that I almost never do C++, so I have put auto everywhere and it compiles dans doesn't crash, so I think I am good.

Yeah, that works :smile: 

I think the code would be verbose if we had to explicitly add the types everywhere. And with xpp auto is required, because the types are impossible to write down.

patrick96 (changes requested):
Changes look good :+1: 

There are two things still missing:

* Adding these changes in the `CHANGELOG.md`
* Users should be able to select randr outputs as well, even if there are monitors defined. In the example in the issue, users should be able to select `DisplayPort-3-left`, `DisplayPort-3-right`, as well as `DisplayPort-3` for the bar monitor. For this, the behavior of `polybar -m` doesn't have to change but `polybar -M` should also display the outputs.

vincentbernat:
OK, I have added a line of code for also showing physical outputs (the one with `any_of`). Changelog added and I squashed everything.

src/x11/extensions/randr.cpp
```diff
@@ -107,56 +107,64 @@ namespace randr_util {
...
-        monitors.emplace_back(make_monitor(output, move(name), crtc->width, crtc->height, crtc->x, crtc->y, primary));
-      } catch (const exception&) {
-        // silently ignore output
+          auto name_iter = info.name();
+          string name{name_iter.begin(), name_iter.end()};
+
+          auto crtc = conn.get_crtc_info(info->crtc);
+          auto primary = (primary_name == name);
+          if (!std::any_of(monitors.begin(), monitors.end(), [name](const auto &monitor) { return monitor->name == name; }))
+            monitors.emplace_back(make_monitor(output, move(name), crtc->width, crtc->height, crtc->x, crtc->y, primary));
```
patrick96, line 164:
```suggestion
          if (!std::any_of(monitors.begin(), monitors.end(), [name](const auto &monitor) { return monitor->name == name; })) {
            monitors.emplace_back(make_monitor(output, move(name), crtc->width, crtc->height, crtc->x, crtc->y, primary));
          }
```

CHANGELOG.md
```diff
@@ -126,6 +126,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([`#2367`](https://github.com/polybar/polybar/issues/2367))
 - Warning message regarding T@ in bspwm module
   ([`#2371`](https://github.com/polybar/polybar/issues/2371))
+- `polybar -m` hides physical outputs covered by monitors
```
patrick96, line 129:
```suggestion
- `polybar -m` shows physical outputs, even if they are covered by monitors
```

patrick96 (changes requested):
Alright, PR seems to work fine :D 

Just some minor changes and we can merge this.

CHANGELOG.md
vincentbernat: Hum, no, previously, `polybar -m` was showing them, now it is hiding it. `polybar -M` is unchanged.

patrick96: Yes, we generally want to describe what was broken. But maybe being more explicit is clearer here:

```suggestion
- `polybar -m` used to show both physical outputs and randr monitors, even if the outputs were covered by monitors.
```

patrick96 (approved):
Looks good now! :tada:

PR #2488, opened 2021-09-03
[module/memory] Display MiB instead of GiB when GiB value is less than 1GiB

### LINKED ISSUE #2472: [module/memory] Display MiB instead of GiB when GiB value is less than 1GiB
I find more intuitive if the memory/swap values are displayed in MiB instead of GiB when the value is less than 1 GiB, for better precision. 
For example, display `512 MiB` instead of `0.50 GiB`.

Coming from `i3-status`, which does this by default, I find this more appropiate

I think that when values are lower than 1GiB is better to see their MiB value: we have an higher precision and the same number of digits. 

In addition to the already existing label tokens:
```
;   %gb_used%
;   %gb_free%
;   %gb_total%
;   %mb_used%
;   %mb_free%
;   %mb_total%
```
add something like: 
```
;   %mb_gb_used%
;   %mb_gb_free%
;   %mb_gb_total%
```
that show the memory value as described earlier.

I'd like the same options available also for swap.

malta895:

* [ ] Refactor
* [X] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

Memory/swap values can now be displayed in MiB instead of GiB when the value is less than 1 GiB, for better precision.
For example, display 512 MiB instead of 0.50 GiB.

The following tokens have been added to the memory module in order to activate this behavior from config:
```
;   %used%
;   %free%
;   %total%
;   %swap_total%
;   %swap_free%
;   %swap_used%
```

## Related Issues & Documents

Closes #2472 

The documentation of the memory module should be updated with the tokens described earlier, and a description on their behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `POLYBAR_FLAGS` cmake variable can be used to pass extra C++ compiler flags.
 
 ### Added
-- `module/memory`: Display MiB instead of GiB when GiB value is less than 1GiB
+- `internal/memory`: New tokens `%used%`, `%free%`, `%total%`, `%swap_total%`, 
+  `%swap_free%`, and `%swap_used%` that automatically switch between MiB and GiB
+  when below or above 1GiB.
   ([`2472`](https://github.com/polybar/polybar/issues/2472))
 - Option to always show urgent windows in i3 module when `pin-workspace` is active
   ([`2374`](https://github.com/polybar/polybar/issues/2374))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `POLYBAR_FLAGS` cmake variable can be used to pass extra C++ compiler flags.
 
 ### Added
+- `module/memory`: Display MiB instead of GiB when GiB value is less than 1GiB
+  ([`2472`](https://github.com/polybar/polybar/issues/2472))
 - Option to always show urgent windows in i3 module when `pin-workspace` is active
   ([`2374`](https://github.com/polybar/polybar/issues/2374))
 - `internal/xworkspaces`: `reverse-scroll` can be used to reverse the scroll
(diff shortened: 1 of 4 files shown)


src/utils/string.cpp
```diff
@@ -291,6 +291,17 @@ namespace string_util {
     return floating_point(kibibytes / 1024.0 / 1024.0, precision, true, locale) + " GiB";
   }
 
+  /**
+   * Create a GiB string, if the value in GiB is >= 1.0. Otherwise, create a MiB string.
+   */
+  string filesize_gib_mib(unsigned long long kibibytes, size_t precision, const string& locale) {
```
patrick96, line 297: I don't think when using MB, it should display anything after the decimal point (just as it already does with the `%mb_*%` tokens). Either have two `precision` arguments, one for MB and one for GB or just always use `0` for the precision when using MB.

src/utils/string.cpp
```diff
@@ -291,6 +291,17 @@ namespace string_util {
...
+    if(kibibytes / 1024.0 / 1024.0 < 1.0) {
```
patrick96, line 298:
```suggestion
    if(kibibytes  < 1024 * 1024) {
```

patrick96 (changes requested):
Very nice! Thanks a lot for helping out :smiley: 

Please also add these changes to the `CHANGELOG.md` file in the `Added` section. Also see the relevant section in [CONTRIBUTING.md](https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog) for more info.

malta895:
> Please also add these changes to the `CHANGELOG.md` file in the `Added` section. Also see the relevant section in [CONTRIBUTING.md](https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog) for more info.

Done! Sorry I missed that part of contributing :sweat_smile:

CHANGELOG.md
```diff
@@ -52,6 +52,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `POLYBAR_FLAGS` cmake variable can be used to pass extra C++ compiler flags.
 
 ### Added
+- `module/memory`: Display MiB instead of GiB when GiB value is less than 1GiB
+  ([`2472`](https://github.com/polybar/polybar/issues/2472))
```
patrick96, line 56:
```suggestion
- `internal/memory`: New tokens `%used%`, `%free%`, `%total%`, `%swap_total%`, 
  `%swap_free%`, and `%swap_used%` that automatically switch between MiB and GiB
  when below or above 1GiB.
  ([`2472`](https://github.com/polybar/polybar/issues/2472))
```

src/modules/memory.cpp
```diff
@@ -1,5 +1,5 @@
 #include <fstream>
-#include <iomanip>
+ #include <iomanip>
```
patrick96, line 2:
```suggestion
#include <iomanip>
```

patrick96:
Looks good now :)

I also changed some small things myself, will merge this once CI is done.

patrick96 (approved):

PR #2505, opened 2021-09-21
feat(ramp): Allow specifying ramp weights

### LINKED ISSUE #1750: ramp and animation stage weights
The main module I encountered this to report by was battery, but I can see this as a helpful feature in general. I was getting frustrated at writing out the same bit of line 20 times in order to accomplish what I wanted. Eg, I currently have 
```
ramp-capacity-0 = 
ramp-capacity-1 = 
ramp-capacity-2 = 
ramp-capacity-0-foreground = ${color.red}
ramp-capacity-1-foreground = ${color.yellow}
ramp-capacity-2-foreground = ${color.lgreen}

animation-charging-0 = 
animation-charging-1 = 
animation-charging-2 = 
animation-charging-0-foreground = ${color.blue}
animation-charging-1-foreground = ${color.blue}
animation-charging-2-foreground = ${color.blue}
animation-charging-framerate = 750
```
The effect I want to achieve is likely going to be that <= 5% should be red, 5 - 25 yellow, 25 - 85 green, and 85+ a different shade of green (the colors aren't support important, I'm still playing with them). The current status quo for ranges seems to be from issue https://github.com/jaagr/polybar/issues/63 where we would have to define at least 20 different ramp capacities and each one's foreground color. 

There are times when we have a lot of repetitive code that could've been written in exponentially shorter number of lines. In addition, it provides for more... dynamic code. Eg, the battery solution could be written as:
```
for i in range(5):
    ramp-capacity-$i$ = 
    ramp-capacity-0-foreground = ${color.red}
for i in range(5, 25):
    ramp-capacity-$i$ = 
    ramp-capacity-0-foreground = ${color.yellow}
for i in range(25, 85):
    ramp-capacity-$i$ = 
    ramp-capacity-0-foreground = ${color.lgreen}
for i in range(85, 100):
    ramp-capacity-$i$ = 
    ramp-capacity-0-foreground = ${color.green}
```
Now, we'd only have to change one line to change a whole range's values. (this could have been done with 2 loops admittedly for actually 20 ramps, but I did it for uniformity for the example)
I discuss these in the other headings :upside_down_face: 

The easiest solution would be to just write my statements out 20 times and use regex replaces. I've considered possibly a custom script, but that sort of defeats the purpose of various modules (like battery). My other solution was writing a python program or of the sort and having it output the lines we can replace. These are obviously less attractive than actual loop functionality.

## Additional context
I realize this feature could be fatally flawed. It would be best to limit the loops to be `for loops`, iterating over only over ranges. A normal loop like `for(int i = 0; i < 5; i++)` or while loop could hold potential to have flawed logic and fail horribly (eg `for(int i = 0; i > 5; i++)`). These should probably be interpreted statically (ie if it makes sense that the loop creates 5 lines, then the loop *always* has those 5 lines, and not 10 in a different minute or something).

TheDoctor314:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

Adding weight attribute in ramps for more gradient-like behaviour.
It defines the size of the interval of the ramp stage.

## Related Issues & Documents

Fixes #1750

We need to document the fact that ramps can have their weights specified for a particular stage


patrick96 (changes requested):
Nice work!

I imagined this being done by calculating which ramp element to use depending on the percentage and the different weights. But your solution is so much less complex :smile: 

Please also mention this in the `Added` section of the changelog: https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog

patrick96 (approved):
Looks good now :)

Thanks a lot for contributing :tada:

PR #2535, opened 2021-10-14
Handle relative includes

### LINKED ISSUE #2523: How to handle relative includes?
polybar's config format supports including files and directories. However, the path is expected to be an absolute path.

In practice this is not true: I have found that if I put in a local path, polybar will try to access it relative to the current working directory. Normally, programs with similar configuration methods (eg. gitconfig) will interpret the relative path in relation to the file itself, not where the program has been launched from.

1. Is the intent to support inclusion with relative paths, or to support absolute paths only? The wiki [section on inclusion](https://github.com/polybar/polybar/wiki/Configuration#file-inclusion) is written as if only absolute paths are supported.
2. If the intent is to support only absolute paths, what is the reasoning behind this? Relative paths are very useful in organizing configs - for example I version my polybar config in a dotfiles repo which is shared across many different computers. The polybar configs don't necessarily end up under the same path on every computer (due to trivial differences like username or `XDG_HOME`). Thus I am forced to add workarounds and add complexity to my dotfiles.
3. If the intent is to support relative paths, can the wiki section be revised to make this clear, and also explain the relative include behavior? Moreover, in that case is it fair to say that the normal behavior should be to interpret the include paths relative to the config itself, not the working directory?

Sorry if I'm missing anything obvious, but having searched through the issues I couldn't really find anything relevant, and I can't imagine I'm the only one who has this problem.

TheDoctor314:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

We change to the directory of the given config file before parsing.
This allows us to handle relative includes.
Absolute paths are handled as before.

## Related Issues & Documents

Fixes #2523 

For 3.6.0, we need to mention in the wiki that `include-file` and `include-dir` works with relative paths now

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -164,6 +164,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Polybar can now be run without passing the bar name as argument given that
   the configuration file only defines one bar
   ([`#2525`](https://github.com/polybar/polybar/issues/2525))
+- `include-directory` and `include-file` now support relative paths. The paths
+  are relative to the folder of the file where those directives appear.
+  ([`#2523`](https://github.com/polybar/polybar/issues/2523))
 
 ### Fixed
 - `custom/script`: Concurrency issues with fast-updating tailed scripts.


patrick96:
Thanks :)

This means that relative paths that don't appear in the main config file but in an included config file which may be in a different folder are resolved relative to the main config file.

I think users would expect a relative path in a config file to be relative to that file.

patrick96:
Now `expand` no longer works with absolute paths: `expand("/tmp/foo", "/scratch")` gives you `/scratch/tmp/foo`. Also `polybar -c` doesn't work with absolute paths anymore

PR #2552, opened 2021-11-17
Add support for a `tray-foreground` setting.

### LINKED ISSUE #2235: Tray doesn't use icon color from gtk theme
## Describe the issue

I using dark background on polybar, and dark gtk-3 theme that sets all symbolic icons color to white. Also i using icon theme with light symbolic icons by default, but as fallback (if needed icon not found) it using "hicolor" icon theme with default dark icons. On polybar tray i see all that icons in they default color. 

**Expected behavior:**

I expected that polybar will take into account gtk theme and change icons color according to it. For example gnome and cinnamon are doing that for they own status bars. Really it's not so important to polybar get color from gtk-3 theme (but it will be perfect variant), i just need a way to change icon color.

**Actual behavior:**

Polybar just don't changing color for symbolic icons in tray. So icons that my main theme includes are white, other icons are black.

## To Reproduce

It happens even with default polybar config. No matter what background or foreground color i set or what is my current gtk-3 theme.

## Polybar Log

I removed all custom scripts and leaved only basic modules, and got this log on start:
```
polybar|warn:  pulseaudio: using default sink alsa_output.pci-0000_00_05.0.iec958-stereo
polybar|error: tray: Failed to put tray above 0x1000001 in the stack (XCB_MATCH (8))
```
Second error appears only when i (re)start i3 with polybar.

## Screenshots
изображение

## Environment:
* WM: I3
* Distro: Linux mint 20
* Output of `polybar -vvv`:
```
polybar 3.4.3

Build type: None
Compiler: /usr/bin/c++
Compiler flags: -g -O2 -fdebug-prefix-map=/build/polybar-udaYph/polybar-3.4.3=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -Wall -Wextra -Wpedantic 
Linker flags: -Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,-z,now
```

jfly:

* [ ] Refactor
* [ ] Feature
* [x] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

This adds a new `tray-foreground` config option, and uses it instead of
the `tray-background` setting to build up the `_NET_SYSTEM_TRAY_COLORS`
property. I'm pretty sure that the previous behavior was a mistake (this
was introduced in 94298741b6de08e79131006504850820bdff9024 which is a
pretty large diff).  For me and some other people, this results in a
black icon being drawn on top of a black background, which is pretty
useless! I would say that this diff fixes
https://github.com/polybar/polybar/issues/2235. Note: the old code dealt
with `unsigned int` and maxed values out at 0xff. The new code deals
with `uint16_t` and maxes values out at 0xffff. I haven't found the
relevant documentation to justify this change, but from manually
testing, I'm pretty confident this is the right change.

This all matches pretty closely with this code from i3:
[`i3bar/src/xcb.c::init_tray_colors`
code](https://github.com/i3/i3/blob/43e805a00d1835db4fc0730ce5003956824b350d/i3bar/src/xcb.c#L1490-L1522),
which you can see also uses the bar's foreground color and maxes values
out at 0xffff, not 0xff.

Before:

2021-11-16_20-32-03_dalinar

After:

2021-11-16_20-32-21_dalinar

## Related Issues & Documents
Fixes #2235

* [x] This PR requires changes to the Wiki documentation: https://github.com/polybar/polybar/wiki/Configuration#bar-settings needs to mention the new `tray-foreground` setting. I'd say something like:

      ; Foreground color for the tray container
      ; ARGB color (e.g. #f00, #ff992a, #ddff1023)
      ; Will use the bar foreground color by default
      tray-foreground = ${root.foreground}

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,6 +135,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([`#2414`](https://github.com/polybar/polybar/issues/2414))
 - `custom/ipc`: Add `hook`, `next`, `prev`, `reset` actions to the ipc module
   ([`#2464`](https://github.com/polybar/polybar/issues/2464))
+- Added a new `tray-foreground` setting to give hints to tray icons about what
+  color they should be.
+  ([#2235](https://github.com/polybar/polybar/issues/2235))
 
 ### Changed
 - Polybar now also reads `config.ini` when searching for config files.
--- a/include/x11/tray_manager.hpp
+++ b/include/x11/tray_manager.hpp
@@ -61,6 +61,7 @@ struct tray_settings {
   unsigned int spacing{0U};
   unsigned int sibling{0U};
   rgba background{};
+  rgba foreground{};
   bool transparent{false};
   bool detached{false};
 };
--- a/src/x11/tray_manager.cpp
+++ b/src/x11/tray_manager.cpp
@@ -133,8 +133,9 @@ void tray_manager::setup(const bar_settings& bar_opts) {
     m_log.warn("tray-transparent is deprecated, the tray always uses pseudo-transparency. Please remove it.");
   }
 
-  // Set user-defined background color
+  // Set user-defined foreground and background colors.
   m_opts.background = conf.get(bs, "tray-background", bar_opts.background);
+  m_opts.foreground = conf.get(bs, "tray-foreground", bar_opts.foreground);
 
   if (m_opts.background.alpha_i() != 255) {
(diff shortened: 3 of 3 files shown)


patrick96 (approved):
I appreciate you taking the time to look into this. Change look good, though it is indeed weird that we would need 16bit of color information per channel.

I couldn't really test this since I never suffered from this issue I think (it should only happen for GTK3 icons).

PR #2557, opened 2021-12-01
feat(net): improve interface discovery 

hantianjz:
## What type of PR is this?

* [x] Optimization

## Description
This PR is a follow up change to #2025 attempt to improve the adapter/net.cpp:find_interface()
to return a interface name that is currently running if any interface is found, else just return the 
first one found.

This will allow the interface-type to work with multiple net interface
of the same type, and prefer to one that is currently connect to
a network. This works great HW with multiple Ethernet port and user
expect to occasionally swap between the two.


patrick96 (approved):
Thanks :smiley:

PR #2559, opened 2021-12-07
Replace system_clock with steady_clock

### LINKED ISSUE #857: Modules stop updating when system clock moves backwards

### LINKED ISSUE #1932: incorrect time after unhibernating

llenck:

* [ ] Refactor
* [ ] Feature
* [x] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
Where used for timing, replace system_clock with steady_clock so that the updating of modules isn't affected by system clock changes.

The m_updated_at field of the mpd module was removed instead of having
its clock changed because it became unused in commit 645a3142a19747c252d24cc71414d40a8eb5f7dc.

## Related Issues & Documents
This fixes #857 and #1932. Also replaces PR #1725, since we don't need
our own implementation of condition_variable anymore since people who
update their polybar should have GCC 10 by now (which was released in may 2020).

Fixes #857
Fixes #1932
Closes #1725

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -199,6 +199,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     ([`#1915`](https://github.com/polybar/polybar/issues/1915))
 - `internal/network`: The module now properly supports 'altnames' for
   interfaces.
+- Some modules stop updating when system time moves backwards. ([`#857`](https://github.com/polybar/polybar/issues/857), [`#1932`](https://github.com/polybar/polybar/issues/1932))
 
 ## [3.5.7] - 2021-09-21
 ### Fixed
--- a/include/adapters/mpd.hpp
+++ b/include/adapters/mpd.hpp
@@ -162,7 +162,6 @@ namespace mpd {
     mpd_status_t m_status{};
     unique_ptr<mpdsong> m_song{};
     mpdstate m_state{mpdstate::UNKNOWN};
-    chrono::system_clock::time_point m_updated_at{};
 
     bool m_random{false};
     bool m_repeat{false};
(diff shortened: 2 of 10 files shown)


patrick96:
Thanks for refreshing the PR :)

I'm not totally up-to-date on the bug. But what happens if someone builds polybar with gcc < 10. Do they just get the old behavior of the clock not being truly monotonic?

If so, I guess that's fine, we don't lose anything. But if this makes polybar not work properly with older gcc versions, we may have to rethink this. We aim to support any currently supported Ubuntu version (at least standard support), so we should try to make polybar work also on Ubuntu 18.04 with gcc 7.4

patrick96 (approved):
Digging deeper into the original but report, it seems older GCC versions will still have the issue. That's fine.

I'll wait another day or two for you to comment, if you maybe have a different insight.

But otherwise, I think this is ready to merge. Thanks a lot :smiley:

llenck:
Also did the digging (on gcc 7.5), and I came to the same conclusion as you did. Thanks for reviewing :)

patrick96:
Alright, thanks!

PR #2569, opened 2022-01-13
Add mac address option

### LINKED ISSUE #2568: Mac Address in Network Module
I'm unable to easily show my mac address in the internal/network module type

It will allow people to easily see their current mac address without the need for a custom script.

There already exist things like %ifname% and %local_ip% which show information about the current network. How about adding %mac_addr%

Creating a custom script in order to read the output of `/sys/class/net/<interface>/address` however this solution doesn't allow me to put my mac address in my network module's text

## Additional context
In order to get the mac address of a certain interface you can read the directory `/sys/class/net/<interface>/address`

Camerooooon:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
This PR adds the %mac% option to the internal network interface modul

## Related Issues & Documents
Closes #2568

* [X] This PR requires changes to the Wiki documentation (describe the changes) (token should be added to wiki page https://github.com/polybar/polybar/wiki/Module:-network)

### DIFF
--- a/src/adapters/net.cpp
+++ b/src/adapters/net.cpp
@@ -131,13 +131,6 @@ namespace net {
     if (!*m_socketfd) {
       throw network_error("Failed to open socket");
     }
-    
-    m_status.mac = file_util::contents(NET_PATH + m_interface + "/address");
-    if (m_status.mac == "") {
-      throw system_error("No MAC address located");
-    }
-
-
 
     check_tuntap_or_bridge();
   }
@@ -167,6 +160,11 @@ namespace net {
         if (!accumulate || (ifa->ifa_data == nullptr && ifa->ifa_addr->sa_family != AF_PACKET)) {
           continue;
         }
+      }
+
+      m_status.mac = string_util::trim(file_util::contents(NET_PATH + m_interface + "/address"), isspace);
+      if (m_status.mac == "") {
+        m_status.mac = NO_MAC;
       }
 
       struct sockaddr_in6* sa6;
--- a/src/adapters/net.cpp
+++ b/src/adapters/net.cpp
@@ -151,6 +151,11 @@ namespace net {
       return false;
     }
 
+    m_status.mac = string_util::trim(file_util::contents(NET_PATH + m_interface + "/address"), isspace);
+    if (m_status.mac == "") {
+      m_status.mac = NO_MAC;
+    }
+
     for (auto ifa = ifaddr; ifa != nullptr; ifa = ifa->ifa_next) {
       if (ifa->ifa_addr == nullptr) {
         continue;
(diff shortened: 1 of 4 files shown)


patrick96:
That was fast :smiley:

Camerooooon:
CI should pass now, I forgot to commit .hpp file

src/adapters/net.cpp
```diff
@@ -130,6 +131,13 @@ namespace net {
     if (!*m_socketfd) {
       throw network_error("Failed to open socket");
     }
+    
+    m_status.mac = file_util::contents(NET_PATH + m_interface + "/address");
+    if (m_status.mac == "") {
+      throw system_error("No MAC address located");
```
patrick96, line 137: If the file reading fails somehow, we don't want to module to completely crash. Just, set the value to `NO_MAC` in that case.

src/adapters/net.cpp
```diff
@@ -130,6 +131,13 @@ namespace net {
     if (!*m_socketfd) {
       throw network_error("Failed to open socket");
     }
+    
+    m_status.mac = file_util::contents(NET_PATH + m_interface + "/address");
```
patrick96, line 135: I just tried changing the MAC address and the address does not change in polybar (because the network object is never recreated). We have to set the MAC address in the `query` function.

The commands I used for changing the MAC address:

```bash
sudo ip link set dev wlp3s0 down
sudo ip link set dev wlp3s0 address XX:XX:XX:XX:XX:XX
sudo ip link set dev wlp3s0 up
```

patrick96 (changes requested):
I have also found an issue with reading the MAC address. The string returned contains a newline character which can mess up the rendering or throw a warning. You should remove any whitespace at the beginning and end of the address with `string_util::trim(..., isspace)`

src/adapters/net.cpp
Camerooooon: Ok, I've moved the update code back into the query function

src/adapters/net.cpp
```diff
@@ -161,6 +162,11 @@ namespace net {
         }
       }
 
+      m_status.mac = string_util::trim(file_util::contents(NET_PATH + m_interface + "/address"), isspace);
```
patrick96, line 165: This code can be outside of the loop, it doesn't depend on anything in the loop and we may just read the address multiple times for no reason.

patrick96 (changes requested):
Last round of reviews I think.

All the code seems to be working fine, changing the MAC address also works now. Just some minor changes.

Please also add an entry to the changelog in the "Added" section: https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog

src/adapters/net.cpp
Camerooooon: If the code isn't in the loop then the mac address won't update if it changes. So if the code doesn't belong in the loop then where should it be?

patrick96: Put it outside of the loop. For example at the beginning of the function. The `query` function is called in every update cycle, so it will get updated.

Camerooooon: Ah, I see what you mean I was confused and thought you meant to put it outside the query function.

Camerooooon:
Alright all done. Also, Thank you very much for helping me through this @patrick96

patrick96 (approved):
Cool, everything looks good now :smiley: 

Thanks a lot! :rocket:

Issue #1814, opened 2019-06-15
Modules with click actions become non-clickable when offset is used


vaygr:
## Describe the issue

When offset is applied to a module with click actions enabled, the area you can click on becomes smaller. It looks like one module overlaps with the other. And if margin between modules is big enough and offset is big enough, the module may becomes completely non-clickable.

See [this note](https://github.com/polybar/polybar/issues/1320#issuecomment-482873992) for reference.

**Expected behavior:**

Clickable area of the module stays intact.

**Actual behavior:**

There's only a couple of clickable pixels left. If offset is too big, the module becomes 100% non-clickable.

**Was it working before?**

No

* Did you also experience this bug in an earlier version of polybar?
  * no
* If no, what was the last version where this worked correctly?
  * none?

## To Reproduce

*A minimal but complete config with which the problem occurs:*
```dosini
[bar/top]
...
module-margin-left = 1
module-margin-right = 2
...
[module/mic]
type = custom/script
exec = ~/.config/polybar/mic.sh
interval = 2

click-left = ~/.config/polybar/mic.sh toggle &

[module/headset]
type = custom/script
exec = ~/.config/polybar/hs.sh "XX:XX:XX:XX:XX:XX"
interval = 2

click-left = ~/.config/polybar/hs.sh "XX:XX:XX:XX:XX:XX" toggle &

format-offset = -10
...
```

*List any other steps needed to reproduce the issue besides starting polybar with the config you posted above.*

None

## Polybar Log

*Post everything polybar outputs to the terminal when you run it and the issue occurs below*

N/A

## Screenshots

N/A

## Environment:

* WM: Fluxbox
* Distro: Source Mage GNU/Linux
* Output of `polybar -vvv`:

```
polybar 3.3.1

Features: +alsa +curl -i3 +mpd +network(libnl) +pulseaudio +xkeyboard

X extensions: +randr (+monitors) -render -damage -sync +composite +xkb +xrm +xcursor

Build type: Release
Compiler: /usr/bin/clang++
Compiler flags: -Wno-error=unused-command-line-argument -march=corei7-avx -fPIC -DPIC -pipe -Os -Wall -Wextra -Werror -pedantic -pedantic-errors -Wno-error=parentheses-equality -Wno-zero-length-array -stdlib=libc++
Linker flags: -s -lc++ -lc++abi
```

## Additional context
*Add any other context that you think is necessary about the problem here.*

None

Lomadriel:
[...]

Issue #325, opened 2017-01-13
MPD title spaces when stopped


NBonaparte:
When MPD is stopped, `<label-song>` becomes spaces instead of being empty. This started occurring around v2.5.0.

PR #2578, opened 2022-01-23
Add units support (POINT, PIXEL, SPACE)

### LINKED ISSUE #951: Support DPI-dependent units

### LINKED ISSUE #1265: "module-margin-left = -1" causes total system halt, Arch AUR package for Polybar 3.1

### LINKED ISSUE #1651: Allow `px` units when specified

### LINKED ISSUE #1700: Offset tag doesn't work as indicated in the documentation

patrick96:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

This basically just resolves the merge conflicts in #2394, which itself just fixed the conflicts from #1671.

In this PR, I will try to bring this feature across the finishline, as I already discussed with @frebib some months ago on IRC.

Thanks to @Lomadriel and @frebib for all their efforts on this feature.

## Checklist

* [x] Address all not-yet addressed reviews in the other two PRs:
    * [x] https://github.com/polybar/polybar/pull/1671#discussion_r369139886
    * [x] https://github.com/polybar/polybar/pull/1671#discussion_r369141688
    * [x] https://github.com/polybar/polybar/pull/1671#discussion_r369143622
    * [x] https://github.com/polybar/polybar/pull/1671#discussion_r369143731
    * [x] https://github.com/polybar/polybar/pull/1671#discussion_r369176673
    * [x] https://github.com/polybar/polybar/pull/1671#discussion_r369193402
* [x] Unit tests for unit.hpp
* [x] Handle invalid values in config (e.g. `100foo`)
* [x] Decide if negative spacing is allowed -> No. Negative spacing doesn't really make sense (and is also not defined for `spacing_type::SPACE`).
* [x] Go through all config options that use extents or percentage_with_offset and see how they deal with negative values (and if they're even allowed). -> Generally, the pixel value is capped below at 0. Exception is the extent value in the `%{O}` tag.
* [x] Address the weirdness in base.inl: https://github.com/polybar/polybar/pull/1671#discussion_r369192910
  * [x] Check that #325 was not affected.
* [x] Confirm that all bugs are fixed
* [x] ~Check whether #1814 is fixed here~ It is not.
* [x] We should also use percentage with offset for `tray-offset-*`
* [x] Write documentation
* [x] Write changelog

## Related Issues & Documents

Addresses the following feature requests:

Closes #1651 
Closes #951

Fixes the following bugs:

Fixes #1700
Fixes #1265 

Makes the following PRs obsolete:

Closes #1671 
Closes #2394

We need to document that the `%{O}` tag now supports `px` and `pt` as units:

---

### Offset `%{O}`

Inserts a gap of the specified size (can also be negative).

```
%{O10} pushed 10 pixels to the right %{O-5pt} 5 points to the left
```

The value specified can be any [extent](#extent) value.

---

The `format-offset` documentation should also be updated:

```dosini
; Displace the format block horizontally by +/-N pixels or points
format-NAME-offset = N[pt|px]
```

We should also document the different supported formats for spacing and extends somewhere. Probably best to just add a section to the "Configuration" wiki page:

---

## Sizes & Spacing

In many parts of the config, you can specify some kind of size (width, height, border) or spacing (margin, padding). These are very similar but all differ slightly. We differentiate between three types: spacing, extents, and "percentage with offset".

### Spacing

Spacing can be specified as either a number of spaces (no unit), pixels (`px`) or points (`pt`).
For example:

```dosini
; 10 spaces
label-padding = 10
; 20 pixels
label-padding = 20px
; 15 points
label-padding = 15pt
```

Spaces are just added as whitespace characters, the size depends on the active font.
One point is 1/72th of an inch and is translated to a number of pixels according to the specified DPI.

### Extent

Extent values are almost like spacing but don't support spaces, only pixels (no unit or `px`) and points (`pt`).
These are mainly used to described the size of something (and thus can't support spaces)

### Percentage with Offset

A percentage with offset specifies the size of something using a relative percentage together with an extent value that is added to it.
For example:

```dosini
[bar/...]
...
width = 90%:-10pt
```

This specifies a bar width of 90% minus 10 points. What exactly the percentage is relative to depends on the setting. For the bar width, it is relative to the width of the monitor.

Both the percentage value and the extent value can also be specified on their own:

```dosini
width = 100%
height = 12pt
```

---

We need to update the wiki to mention what kind of format is allowed in the following places:

**Spacing**

* `ramp-coreload-spacing` in the CPU module
* `spacing` in the FS module
* `padding` and `module-margin` in the bar section
* label padding and margin
* format spacing, margin, padding

**Extent**

* `format-offset`

**Percentage with offset**

* Struts (in `[global/wm]`)
* Border size
* Bar width/height
* Bar offset (x/y)
* `tray-offset-*`


patrick96:
Alright, this is finally done!!

I want to thank again @Lomadriel and @frebib for doing a lot of heavy lifting for this feature. This will be very useful to a lot of users.

PR #2585, opened 2022-01-30
Stop using deprecated CMake command in uninstall template

Riyyi:

* [ ] Refactor
* [ ] Feature
* [ ] Bug Fix
* [x] Optimization
* [ ] Documentation Update

## Description
The uninstall template is using a deprecated CMake command to remove the files, this patch fixes that.
To confirm, I ran the following command `$ cmake -E help` which outputs:
```
  remove [-f] <file>...     - remove the file(s), use -f to force it (deprecated: use rm instead)
```


patrick96:
The minimum supported cmake version is 3.5 which does not have `cmake -E rm` ([`man cmake`](https://cmake.org/cmake/help/v3.5/manual/cmake.1.html)). It seems this was introduced in cmake 3.17.

After polybar 3.6 is released, we will bump the cmake version to 3.10 or 3.16 (this is what Ubuntu LTS has). So even then, this change would break these versions.

PR #2590, opened 2022-02-06
Add netspeed parameter (#1083)

### LINKED ISSUE #1083: Show total of %upspeed% and %downspeed% as %netspeed%  in network module
I would like to see a single speed value for my wireless connection instead of having to display both speeds, which is possibly similar to %linkspeed% for wired connection. I can't figure out a way to do it currently using the network module, and I don't want to shift to a custom script implementation. A new label token called  `%netspeed% = %upspeed%+%downspeed%` would be nice.

Also, is it possible to display two different prefixes based on whichever is higher (upspeed vs. downspeed) when %netspeed% is used?

prateekmedia:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

Adds netspeed parameter which is basically sum of up and down speeds

## Related Issues & Documents

Fixes #1083, Although this doesn't add dynamic icons of up and down

`netspeed` parameter uses needs to be described in the network module documentation.

### DIFF
--- a/src/modules/network.cpp
+++ b/src/modules/network.cpp
@@ -148,6 +148,7 @@ namespace modules {
 
     auto upspeed = network->upspeed(m_udspeed_minwidth, m_udspeed_unit);
     auto downspeed = network->downspeed(m_udspeed_minwidth, m_udspeed_unit);
+    auto netspeed = network->netspeed(m_udspeed_minwidth, m_udspeed_unit);
 
     // Update label contents
     const auto replace_tokens = [&](label_t& label) {
@@ -158,7 +159,7 @@ namespace modules {
       label->replace_token("%local_ip6%", network->ip6());
       label->replace_token("%upspeed%", upspeed);
       label->replace_token("%downspeed%", downspeed);
-      label->replace_token("%netspeed%", upspeed + downspeed);
+      label->replace_token("%netspeed%", netspeed);
 
       if (m_wired) {
         label->replace_token("%linkspeed%", m_wired->linkspeed());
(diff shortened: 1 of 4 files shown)


src/modules/network.cpp
```diff
@@ -158,6 +158,7 @@ namespace modules {
       label->replace_token("%local_ip6%", network->ip6());
       label->replace_token("%upspeed%", upspeed);
       label->replace_token("%downspeed%", downspeed);
+      label->replace_token("%netspeed%", upspeed + downspeed);
```
patrick96, line 161: This doesn't work. `uspeed` and `downspeed` are strings, this will just show nonsense. For example if I have 1KB download and 5KB upload, this will show `1KB5KB`.

patrick96 (changes requested):

src/modules/network.cpp
prateekmedia: Oh,So we have to create another function for netspeed everywhere?

prateekmedia: @patrick96 Can you check now.

patrick96 (changes requested):
Exactly! Code looks good now, thanks :)

One last thing, please also mention this change in the changelog in the `Added` section (see https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog)

prateekmedia:
@patrick96 I am also thinking to add total used net speed so someone like me can keep track of total used data of the session. Not in this PR but in next PR of course. So what do you think?

patrick96 (approved):

patrick96:
Merged! Thanks a lot :smiley: 

> I am also thinking to add total used net speed so someone like me can keep track of total used data of the session. Not in this PR but in next PR of course. So what do you think?

If you have a use-case for it, sure, go for it. Just note that the number of bytes received or sent we get from `getifaddrs` is only a uint32 so it wraps around after 4GB and isn't really useful to show the total number of bytes in the session.

Issue #1526, opened 2018-11-24
Make tray a module


fakheet:
So, basically, why not just handle the systray as a module, rather than a separate window layered on top? 

Right now tray is kind of wonky because it really only has 3 placement options and they are all meh in terms of aesthetics. And If you want to put the tray somewhere else, you better hope that no icons are added or removed, otherwise it will either overlap with neighboring elements or leave unnecessary gaps. Some people might prefer to have the tray at some particular place (e.g. between battery and clock). 

I don't see how making tray a module could interfere with the design, since tray already scales to bar's size and adheres to some of the rules, but I'm new to polybar so there may be some reason, why it can't be done, that I am unaware of. If so, please correct me.

patrick96:
> just

It's unfortunately not that easy. Currently the tray uses XEMBED which is a protocol to position windows inside other windows, this means each tray icon you see is actually a window that is handled by whatever application provides the icon and polybar doesn't have much control over that window (other than positioning). This is the reason the tray has to be a separate window.

NBonaparte is doing some work on a proper systray module (see #913), which uses a new tray specification. StatusNotifierItem. But the whole endeavour is far from finished.

fakheet:
I see, thank you for your answer.

Issue #314, opened 2017-01-10
systray as a module


eoli3n:
Systray is placed by var ``tray = right|left``

I think that it could be a nice thing to make it as a module. To be able to place it where we want, and to duplicate it (don't think it is possible, just wondering).

jaagr:
it's not possible to duplicate it, but I've started to work on a systray module...

kevlar1818:
@jaagr I'm eagerly awaiting this feature. Any word on its progress?

jaagr:
@kevlar1818 I've been very busy lately so I haven't had much time over for development, but I will try to finish it once I get some bugs fixed. I might also commit my work-in-progress branch for the feature so that others may help complete it.

SuperPrower:
Hi, any updates on systray as module? I understand it's probably is most painful module for you to work on, but making it module will fix so much problems, like actual centering of modules in center or ability to place it somewhere in right, but not in the edge of the screen.

132ikl:
@jaagr Any update on the systray module? It's been quite a while now.

mhham:
Yes, this feature would be a very nice addition to polybar !

kevlar1818:
@jaagr Any updates on this?

Arignir:
Would be a cool feature, yep

zjeffer:
@jaagr Still working on this module? I'd really like this feature :)

Alekaei:
Commenting to show support and need for this feature

patrick96:
I think NBonaparte did some work on this (see #913). This also relies on us finishing #1088 first.

jjzmajic:
Any updates on this in 2020?

chief-wrench:
Came here to vote for this feature.

Masacaai:
Commenting to show support for this feature!

tpurde:
Any news on this?

TaylanTatli:
Please make this possible.

BRA1L0R:
Bump as this would be super useful

ignaudioz:
bump

iPocket:
Bumping to help this feature request get acknowledged :)

Vlek:
Bumping. Is there a bounty on this or should I make a new one?

patrick96:
[...]

Issue #651, opened 2017-07-14
Radius not respected with systrays


sebastiencs:
Hi,

When I set the radius and a tray icon is inserted it remove the radius:
![radius](https://user-images.githubusercontent.com/5273820/28198931-4ebf3590-6894-11e7-9c2d-1b96aec88a9c.gif)

jaagr:
Hi,

The reason it does this is because the tray is actually another window on top of the bar window. To get around this you could offset the tray by the radius value, for example:

~~~ dosini
[bar/mybar]
radius = 5.0
tray-offset-x = ${self.radius}
~~~

patrick96:
This actually doesn't work, even when the tray is offset, it leaves a black box (which is square) where it would be without offset

sebastiencs:
I confirm, it doesn't work, even with the offset:
![polybar](https://user-images.githubusercontent.com/5273820/28212017-2512bc1a-68d2-11e7-9025-4d8204ea7446.jpg)

PixelGGamer:
![screenshot from 2017-07-24 09-36-29](https://user-images.githubusercontent.com/7670940/28512863-f0ced892-7053-11e7-9031-0dfc5a2d4ef0.png)
The only way for now is setting it to center, which is well, ugly. Looking forward to solution

karambaq:
+1

tasmo:
One solution could be handle the tray as a module which made it possible to move the tray in its position before an other module.

ReiLucio:
The only workaround i found was this, works fine only if your systray width is static... but rice comes first :

`padding-right = 13
`
`tray-detached = true
`
`tray-offset-x = -4
`
![selection_999 019](https://user-images.githubusercontent.com/32798365/42407767-02c8c8cc-8198-11e8-9895-665dad084692.png)

gitneeraj:
> The only workaround i found was this, works fine only if your systray width is static... but rice comes first :
> 
> `padding-right = 13 `
> `tray-detached = true `
> `tray-offset-x = -4 `
> ![selection_999 021](https://user-images.githubusercontent.com/32798365/42408042-4de2d0fa-819d-11e8-9978-a94d9691faff.png)

Does not work if more tray icons appear. Like I have slack, zoom, nm etc.
Initially works fine but when dynamically icons are added, they overlap. :(

AlecsFerra:
news?

Fable-git:
[...]

PR #1615, opened 2019-01-19
[WIP] Fix tray restacking issues

patrick96:
**EDIT:** Maybe it's a better idea to just make the tray window a child window of polybar. Then repositioning should get a lot easier.
Alternatively, have a main window and have the bar and the tray as two separate child windows.

---

The restacking issues observed in #1355 and #425 is caused by the WM reparenting the polybar window but not the tray window when `override-redirect = false`. Because of this the two windows are no longer siblings, causing the `XCB_MATCH` error.

Automatic reparenting causes problems with `tray-offset-*` not being applied in at least openbox and maybe other WMs. The reason for that is that openbox reparents the bar window into a window that hast the same dimensions as the bar window, if we also move the tray under that window, the tray cannot be drawn outside of the bar bounds. My solution here is a somewhat breaking change because by default reparenting is turned on. I think this is fine because it's only a breaking change in WMs that make the bar a child of a window with tight bounds and in configs that use the tray offset to move the tray outside of the bar's bounds. There is no other way to fix two linked issues in those WMs.

To not totally break people's configs that may want to move the tray outside of the bar bounds on those WMs, I have introduced a new key in the bar section:
```dosini
; Make sure that the bar and tray windows are siblings
; This option generally doesn't need to be modified unless you have issues with
; the tray offset not working.
; If set to false, you may get `tray: Failed to put tray above ...` errors and
; the tray may appear on top of fullscreen windows.
tray-reparent = true
```

Fixes #1355 
Fixes #425
Fixes #1813 
Fixes #1995

This fix could also work for #898 
This fix could also work for #2460
This fix could also work for #2433
This fix could also work for #1995
This fix could also work for #789
This fix could also work for #1537
Closes #1657
Fixes #2035

PR #2595, opened 2022-02-16
Tray position like a module

raffael0:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
This PR fixes the long-standing issue of only having 3 positioning options for the system tray. It does this by adding a new module which acts as placeholder. The actual tray window is then placed onto that tray module. Resizing of the tray and repositioning of the module is handled via signals.

These changes should also make it easier to completely break the tray handling out of the main bar.cpp into the new tray module.

## Related Issues & Documents
Fixes:  #1526
Fixes: #314

Since this PR adds a complete new option for tray positioning this has to be reflected in the wiki

## TODO / Issues
* [ ] define how the tray positioning will be handled in the config. Will tray-position be deprecated?
* [x] figure out how to get the position where a tag will be placed in dispatch.cpp. The current approach only works on the left and in the center ([current implementation](https://github.com/raffael0/polybar/blob/e112604c94e47da0990e5d8c8d4387e8b03f18b9/src/tags/dispatch.cpp#L146))

I'd love some input on these issues


src/modules/tray.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include "modules/meta/base.inl"
+
+POLYBAR_NS
+namespace modules {
+  template class module<tray_module>;
+
+  tray_module::tray_module(const bar_settings& bar_settings, string name_)
+      : event_module<tray_module>(bar_settings, move(name_)) {
+    m_formatter->add(DEFAULT_FORMAT, "", {});
```
patrick96, line 15: To be consistent, we should also use `format = <tray>` for the tray. This here just adds a pseudo format that doesn't really do anything (since it has no tags).

```suggestion
    m_formatter->add(DEFAULT_FORMAT, TAG_TRAY, {TAG_TRAY});
```

And `TAG_TRAY` could be defined as `<tray>`.

In that case, you also don't need `get_output` and can just check for `TAG_TRAY` in `tray_module::build`. See how the script module does this: https://github.com/polybar/polybar/blob/ab915fb724546e7abfcb35cbaf863ce2bca613ec/src/modules/script.cpp#L114-L122

patrick96 (changes requested):
Wow that's really cool :smiley: 

I appreciate you working on this. The tray has been a pain point for a long time and I have been dreading working on it forever.
This is surprisingly little code for what I thought would be a major undertaking.

Please let me know what your commitment for this PR is. Are you prepared to go through revisions until this is in a stable state? Or do you feel like this a "this is my code, take it or leave it" situation? Or something in between?
I don't want to push you until you lose interest, so it's good to know what kind of expectations you have here.
Of course feel free to at any point say that you no longer can/want to invest more time here.
We can just as well release this as an experimental, but mostly working, feature and improve it step by step.

---

While testing this, I found some issues, all of which happen using the following config:

```dosini
[bar/example]
modules-left = text tray text

enable-ipc = true

tray-position = adaptive

width = 50%
offset-x = 50%

[module/text]
type = custom/text
content = A

[module/tray]
type = internal/tray
```

* The `tray_manager` uses the position received from `tray_pos_change` as the absolute position for the tray window. However, the dispatcher only sends the x position relative to the bar window. This is an issue if you have two monitors and place the bar on the right one or when the bar has some `offset-x`.
* The tray module does not receive the `tray_width_change` signal and thus never updates its width and doesn't produce a proper offset tag. This causes modules that are placed after the tray to appear immediately after the previous module (and the tray overlaps them)
I'm pretty sure that's because the module is not attached to the signal handler and thus doesn't receive any signals.

I think these two are the most important points to get the module into a experimental state that we can release to users.

There are also some conceptual things and edge cases to consider:

* If you set `tray-position` to anything but `adaptive` but still have a tray module, the tray module will display the tray, but the bar will still cut off parts of the bar in `bar::parse`. We should enforce that if `tray-position = adaptive`, there must be exactly one tray module. And if it is anything else, no tray module must exist. We could also ditch `tray-position = adaptive` and just check for the existence of a tray module.
* If there is too much text on the bar, any excess text will just drop off the bar on the right. We have to decide (and properly handle) what happens when tray is shoved off the right side of the bar like that. My idea is that in that case the tray just stays at the right edge of the bar and never disappears. This can also easily be enforced by the `tray_manager` when it handles the `tray_pos_change` signal.
* We also have to think about how we want to move all the tray configurations that are currently into the bar section into the tray module. But this doesn't have priority at all yet, we can easily tackle this once the tray is stable.

These points are not yet important if we just release the tray module as an experimental feature, but we will need to address them to get to a stable state.

---

>  define how the tray positioning will be handled in the config. Will tray-position be deprecated?

In order to not break existing configs, we need to still support `tray-position`. Depending on whether we merge this as an experimental feature or try to get a stable implementation in this PR, we can already tell people to switch and that all the tray settings in the bar section will be deprecated. But leave it as-is for now.

> figure out how to get the position where a tag will be placed in dispatch.cpp. The current approach only works on the left and in the center

I think the solution here is to not immediately move the tray but only emit the tray position signal once the rendering is finished. Either the dispatch or the renderer need to keep track in which alignment and which relative position in that alignment the tray has to appear and at the end of the render cycle calculate the actual position.

raffael0:
Hi, thanks for your quick and detailed response!

> Please let me know what your commitment for this PR is. Are you prepared to go through revisions until this is in a stable state? Or do you feel like this a "this is my code, take it or leave it" situation? Or something in between?

I'd ultimately like to see this become stable and I'm committed to continue working on it. Since I don't really understand some parts of the code works(modules, formatter etc.) I'll need some help in these areas and definitely appreciate any suggestions. The next few months will be quite busy for me so it may be that I'll need to take a break for a while but I'll definitely let you know before.  

> The tray_manager uses the position received from tray_pos_change as the absolute position for the tray window. However, the dispatcher only sends the x position relative to the bar window. This is an issue if you have two monitors and place the bar on the right one or when the bar has some offset-x

I also noticed this error and the tray positioning has been the most annoying issue I faced. Moving the tray position signal to the renderer sounds like a good idea. I never liked it in the dispatch. 
Am I correct that you propose moving it into renderer::end? 
Do I get the absolute and relative position like this?
In renderer:
```cpp 
 double absolute_x = m_rect.x + block_x(alignment);

```
in dispatch:
```cpp
double current_relative_x = renderer.get_x(*m_ctxt);
```

For now I'll look into the problem with the tray_width_change signal and incorporate your suggestions.

patrick96:
> I'd ultimately like to see this become stable and I'm committed to continue working on it. Since I don't really understand some parts of the code works(modules, formatter etc.) I'll need some help in these areas and definitely appreciate any suggestions.

Just let me know if anything is unclear. Our code documentation isn't that good and a lot of code (especially the base module code) has been here for a long time (even before I started working on polybar) and its behavior isn't always obvious for me either.

> The next few months will be quite busy for me so it may be that I'll need to take a break for a while but I'll definitely let you know before.

That's completely fine :)
There is no rush to get this merged. I'd rather we take our time than do it sloppily.

---

The tray positioning is terrible. It's the main reason why I got stuck in my last big tray PR #1615.
The "proper" solution would be to make the tray window a child window of the bar window. Then we only have to position it relative to the parent and we also a avoid all the issues in #425.
However, that is probably quite a big change (bigger than what you did already), so I'm not sure if it's a good idea to try to force into this PR as well.
For now, let's keep the current system.

Unfortunately, neither of your proposed solutions work, both of them are still relative to the bar window. `m_rect` is relative to the bar window but only includes the inner area of the bar without the borders.

I think the following could work:
We store the tray position in the tag context (maybe as an `std::pair<alignment, int>`). The integer value is the x-position relative to the alignment. The dispatcher can easily find get that using `renderer.get_x`.

Once the dispatcher is finished, it calls into the renderer and the renderer can calculate the position relative to the inner area of the bar like this:

```cpp
double absolute_x = block_x(ctxt.tray.first) + ctxt.tray.second;
```

It then sends this to the tray_manager and the tray manager is responsible for calculating the absolute position. The tray gets passed the `bar_settings` which are necessary to get the absolute position in `setup`. However, I think it makes more sense to pass it into the constructor and also store it as a private field (`const bar_settings&`).
Because then, the tray manager can easily compute the absolute position when it receives the `tray_pos_change` signal like this:

```cpp
m_opts.orig_x = m_bar_opts.inner_area(true).x + evt.cast();
```

src/modules/tray.cpp
patrick96: You can completely remove `get_output`. There is already a `get_output` defined for all modules in `base.inl` which takes care of invoking the `build` function.

raffael0:
- [x] Tray Width error: should be fixed
- [x] Tray positioning: Implemented you suggested fix. works with your example config
- [ ] Testing:  Since I added a new method to the `renderer`, I need to mock that in the [MockRenderer](https://github.com/raffael0/polybar/blob/185e271e835965c0198b2888c99b82591636c3e4/tests/unit_tests/tags/dispatch.cpp#L29) however my added Mock doesn't compile. Could you take a look at it?

patrick96 (changes requested):
This is coming along nicely :)

I have not had any problems when testing it out. The reviews are mainly cosmetic/architectural ones.

Please also run `clang-format` over all files that you have changed so that we have consistent formatting ([Style Guide](https://github.com/polybar/polybar/wiki/Style-Guide#code-formatting)).
You can run clang-format for all changed files using this command:

```
git diff --pretty="" --name-only "$(git merge-base HEAD upstream/master)..HEAD" | xargs clang-format -i --style=file
```

Note that this will completely mess up non-C++ files (cmake etc.), so just don't commit those ;)

After this round of reviews this should be in a state where we can merge this as an experimental feature. After that, we should think about the three optional points in my first review.

What do you prefer? Should we merge this as an experimental feature or first bring it into a stable state in this PR?

patrick96:
This may also fix #651, but I haven't tested it yet.

raffael0:
Hi,
just wanted to give an update on my progress. I'm on the break now I mentioned at the beginning so I currently don't have time to work on this. I'll  probably return next month. Maybe I can squeeze in the latest requested changes in the coming weeks to release it as an experimental feature.

patrick96:
Thanks for letting me know :)

There is no need to rush here, take your time

Luekuu:
Hi Polygon Team, I'm new to the polygon/linux and github world. How i can "download" / "test" this tray-modul? 
Any beta files or so? 
Regards, Lukas

patrick96:
@Luekuu This is very much not ready to use for end users. If you really want to, you can clone @raffael0's fork and [compile](https://github.com/polybar/polybar/wiki/Compiling) the [`working/variable-tray`](https://github.com/raffael0/polybar/tree/working/variable-tray) branch from source.

patrick96:
Will review in depth later.

Your latest commit messed up the formatting in a bunch of files.
Please also revert the formatting changes in the cmake files.

doc/conf.py
```diff
@@ -60,7 +61,7 @@ def get_version(root_path):
 else:
   # In all other builds conf.py is configured with cmake and put into the
   # build folder.
-  doc_path = '@doc_path@'
+  doc_path = '/home/raffael/Documents/Schule/21-22/ANDERE_PROJEKTE/polybar/doc'
```
patrick96, line 64: Please revert all changes in this file

include/components/renderer_interface.hpp
```diff
@@ -33,6 +33,8 @@ class renderer_interface {
    */
   virtual double get_alignment_start(const alignment align) const = 0;
 
+  virtual void apply_tray_position(unique_ptr <polybar::tags::context> &context) = 0;
```
patrick96, line 36:
```suggestion
  virtual void apply_tray_position(const tags::context& context) = 0;
```

It doesn't really make sense to pass a reference to a unique_ptr, passing a const-reference here is much cleaner (that's what the dispatcher does everywhere else anyway).

raffael0:
Sorry about that. I ran clang-format but must have messed something up. I should have removed all of the wrong formatting now.

raffael0:
Sorry about that. I have no idea how that happened. Should be fixed now.

patrick96:
> Sorry about that. I have no idea how that happened. Should be fixed now.

Maybe your IDE auto-formatted the files.

patrick96 (changes requested):
Alright, looks good now :smiley: 

Things to do now to merge this as an experimental feature:

* [x] Add a changelog entry: https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog
* [ ] Addressing the remaining reviews (mainly cosmetic)
* [ ] Remove the `bar_settings` argument from `tray_manager::setup`
* [ ] Resolve merge conflicts. Let me know if you need help with that.

After merging, we need to take these steps to make this feature stable

* [ ] Automatically hide the tray window when no tray position is set in the renderer
* [ ] Consider removing `tray-position = adaptive` again and just activate the tray when a tray module exists. The existence of a tray module should override any value for `tray-position`
* [ ] Fix the tray window being pushed off the bar. The tray window should always be fully within the bar window¨.
* [ ] Also support applicable tray settings in the tray module (`tray-maxsize`, `tray-background`, `tray-foreground`, `tray-padding`, `tray-scale`, and maybe `tray-offset-y`)
* [ ] Emit deprecation warnings for `tray-position` and all other tray settings in the bar section.$

These can be implemented step-by-step in separate PRs. I'll create an issue for them once we merge this.

raffael0:
Hi. these commits should address all your points:
- Added changelog entry(I'm not sure about the wording here)
-  implemented the suggestions
- removed the bar_settings
- resolved the merge conflicts(I rebased my branch onto master. the tray new handling still works and the tests pass after rebasing)

patrick96 (changes requested):
Alright, last rounds of reviews :)

tests/unit_tests/tags/dispatch.cpp
```diff
@@ -53,6 +54,10 @@ class FakeRenderer : public renderer_interface {
     return 0;
   };
 
+  void apply_tray_position(const polybar::tags::context& context) override {
```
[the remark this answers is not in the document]
raffael0: thanks. Didn't know about that :)

CHANGELOG.md
```diff
@@ -18,7 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `internal/pulseaudio`: `reverse-scroll` option ([`#2664`](https://github.com/polybar/polybar/pull/2664))
 - `custom/script`: Repeat interval for script failure (`interval-fail`) and `exec-if` (`interval-if`) ([`#943`](https://github.com/polybar/polybar/issues/943), [`#2606`](https://github.com/polybar/polybar/issues/2606), [`#2630`](https://github.com/polybar/polybar/pull/2630))
-- `custom/text`: Loads the `format` setting, which supports the `<label>` tag, if the deprecated `content` is not defined ([`#1331`](https://github.com/polybar/polybar/issues/1331), [`#2673`](https://github.com/polybar/polybar/pull/2673), [`#2676`](https://github.com/polybar/polybar/pull/2676))
+- `custom/text`: Loads the `format` setting, which supports the `<label>` tag, if the deprecated `content` is not
+  defined ([`#1331`](https://github.com/polybar/polybar/issues/1331)
+  , [`#2673`](https://github.com/polybar/polybar/pull/2673), [`#2676`](https://github.com/polybar/polybar/pull/2676))
```
patrick96, line 23:
```suggestion
- `custom/text`: Loads the `format` setting, which supports the `<label>` tag, if the deprecated `content` is not defined ([`#1331`](https://github.com/polybar/polybar/issues/1331), [`#2673`](https://github.com/polybar/polybar/pull/2673), [`#2676`](https://github.com/polybar/polybar/pull/2676))
```

raffael0:
Is this ready to be merged or am I missing an open issue?

patrick96 (approved):
> Is this ready to be merged or am I missing an open issue?

Everything's good, I was just busy.

patrick96:
Awesome! Very happy to have this in polybar now :smiley: 

Thanks a lot @raffael0 for all your work!

I opened #2689 to track the next steps. If you want to work on some of that, feel free ;)

I'll be off for a week, so I won't be able to review anything.

raffael0:
Great!
I'm also quite busy in coming weeks but I'll probably help out with the tasks still remaining

Vlek:
@raffael0, hello! I am trying to use this. However, there does not appear to be any documentation for it.

I have tried adding it to my bar and using `internal/tray`, but my polybar output gives an error saying that it is an unknown module. I made sure I'm using the latest version, 3.6.3 which appears to be after when this was merged to the master branch. Am I using this incorrectly? Is it not fully available yet?

raffael0:
Hi,
this feature is targeted to be released in version 3.7. It was however merged into master and you can use it from there(I think polybar-git in the AUR should be on master or you can [build from source](https://github.com/polybar/polybar/wiki/Compiling)). We're still working on some issues to stabilize the module(See  #2689). Until then the feature is still experimental and the interface(eg. tray-position=adaptive) will probably change a bit until it's released.

PR #2596, opened 2022-02-18
feat(script): add formatting for script failure

### LINKED ISSUE #2588: script module: add formatting for script failure

I would love to be able to configure a script module to use different formatting and foreground/background colors if a command/script fails i.e. returns an exit code other than 0.

Well, I figured being able to use different formatting and colors for a failed command (e.g. background turns bright red if a command specified in `exec = ` returns a non-zero exit code) would be a very welcome feature as it increases visibility, I suppose.

I'm thinking of something like so:
```
format-success-prefix =
format-success-suffix =
format-success-foreground =
format-success-background =
format-failure-prefix =
format-failure-suffix = 
format-failure-foreground =
format-failure-background =
```
So for example if the command I run in `exec = ` succeeds, then the output will be shown with the formatting specified by the `format-success-*` configurations. Likewise if the command fails, then `format-failure-*` will be used to format the output on the bar.

As dumb as it sounds, I have tried putting `exec = tput setaf 1; echo "Hello"` in the script module. I was totally surprised to find out that it didn't work. I could not think of other alternatives, honestly, but I might not have RTFM-ed enough.

indev29:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [x] Documentation Update

## Description
_The behavior is not changed unless the feature is used. Old configuration files should work the same way._

This PR adds formatting for script failure (when exited with non-zero exit status):

```ini
format-fail = <label-fail>
; and all the others format-fail-*

label-fail = %output%
; and all the others label-fail-*
```

Falls back to `format`, if `format-fail` isn't specified.

The feature is disabled for `tail=true` scripts, since they either run indefinitely (so exit status doesn't have a change to be used) or stop for an `interval` that should be relatively short compared to the lifetime of a script. In both scenarios, exit status is received too late for it to be useful so there's no way to implement meaningful formatting for those scripts.
 
## Related Issues & Documents
Closes #2588 

The following should be added to [Modules/script/Additional Formatting](https://github.com/polybar/polybar/wiki/Module:-script#additional-formatting):

```ini
; Format for script failure (when exited with non-zero exit status)
; If not specified, 'format' is used instead
; Available tags:
;   <label-fail> (default)
format-fail = <label-fail>
format-fail-background = #900
format-fail-foreground = #fff
format-fail-padding = 4

; Available tokens:
;   %output%
; Default: %output%
label-fail = %output:0:15:...%
```

### DIFF
--- a/include/adapters/script_runner.hpp
+++ b/include/adapters/script_runner.hpp
@@ -12,8 +12,8 @@ POLYBAR_NS
 class script_runner {
  public:
   using interval = std::chrono::duration<double>;
-  script_runner(std::function<void(void)> on_update, string exec, string exec_if, bool tail, interval interval,
-      const vector<pair<string, string>>& env);
+  script_runner(std::function<void(void)> on_update, const string& exec, const string& exec_if, bool tail,
+      interval interval, const vector<pair<string, string>>& env);
 
   bool check_condition() const;
   interval process();
--- a/include/adapters/script_runner.hpp
+++ b/include/adapters/script_runner.hpp
@@ -12,8 +12,8 @@ POLYBAR_NS
 class script_runner {
  public:
   using interval = std::chrono::duration<double>;
-  script_runner(std::function<void(void)> on_update, const string& exec, const string& exec_if, bool tail,
-      interval interval, const vector<pair<string, string>>& env);
+  script_runner(std::function<void(void)> on_update, string exec, string exec_if, bool tail, interval interval,
+      const vector<pair<string, string>>& env);
 
   bool check_condition() const;
   interval process();
@@ -24,13 +24,14 @@ class script_runner {
 
   int get_pid() const;
   int get_counter() const;
+  int get_exit_status() const;
 
   string get_output();
 
   bool is_stopping() const;
 
  protected:
-  bool set_output(const string&&);
+  bool set_output(string&&);
 
   interval run_tail();
   interval run();
@@ -52,6 +53,7 @@ class script_runner {
   std::atomic_int m_counter{0};
   std::atomic_bool m_stopping{false};
   std::atomic_int m_pid{-1};
+  std::atomic_int m_exit_status{0};
 };
 
 POLYBAR_NS_END
--- a/include/modules/script.hpp
+++ b/include/modules/script.hpp
@@ -32,7 +32,6 @@ namespace modules {
 
     const bool m_tail;
     const script_runner::interval m_interval{0};
-    const bool m_has_format_fail;
 
     script_runner m_runner;
 
--- a/include/modules/script.hpp
+++ b/include/modules/script.hpp
@@ -16,6 +16,8 @@ namespace modules {
     void stop() override;
 
     string get_output();
+    string get_format() const;
+
     bool build(builder* builder, const string& tag) const;
 
     static constexpr auto TYPE = "custom/script";
@@ -24,16 +26,20 @@ namespace modules {
     bool check_condition();
 
    private:
-    static constexpr const char* TAG_LABEL{"<label>"};
+    static constexpr auto TAG_LABEL = "<label>";
+    static constexpr auto TAG_LABEL_FAIL = "<label-fail>";
+    static constexpr auto FORMAT_FAIL = "format-fail";
 
     const bool m_tail;
     const script_runner::interval m_interval{0};
+    const bool m_has_format_fail;
 
     script_runner m_runner;
 
     map<mousebtn, string> m_actions;
 
     label_t m_label;
+    label_t m_label_fail;
   };
 }  // namespace modules
 
--- a/src/adapters/script_runner.cpp
+++ b/src/adapters/script_runner.cpp
@@ -11,7 +11,7 @@
 
 POLYBAR_NS
 
-script_runner::script_runner(std::function<void(void)> on_update, string exec, string exec_if, bool tail,
+script_runner::script_runner(std::function<void(void)> on_update, const string& exec, const string& exec_if, bool tail,
     interval interval, const vector<pair<string, string>>& env)
     : m_log(logger::make())
     , m_on_update(std::move(on_update))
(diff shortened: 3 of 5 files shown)


src/adapters/script_runner.cpp
```diff
@@ -74,7 +78,7 @@ bool script_runner::is_stopping() const {
  *
  * Returns true if the output changed.
  */
-bool script_runner::set_output(const string&& new_output) {
+bool script_runner::set_output(string&& new_output) {
```
patrick96, line 81: I don't even know why this code ever compiled before. Looking at it again, `std::move(new_output)` should have produced a warning since it only takes a non-const rvalue reference.

include/modules/script.hpp
```diff
@@ -24,16 +26,20 @@ namespace modules {
...
 
    private:
-    static constexpr const char* TAG_LABEL{"<label>"};
+    static constexpr auto TAG_LABEL = "<label>";
+    static constexpr auto TAG_LABEL_FAIL = "<label-fail>";
+    static constexpr auto FORMAT_FAIL = "format-fail";
 
     const bool m_tail;
     const script_runner::interval m_interval{0};
+    const bool m_has_format_fail;
```
patrick96, line 35: This is not needed, you can use `m_formatter->has_format` to check if the fail format exists.

include/adapters/script_runner.hpp
```diff
@@ -12,8 +12,8 @@ POLYBAR_NS
 class script_runner {
  public:
   using interval = std::chrono::duration<double>;
-  script_runner(std::function<void(void)> on_update, const string& exec, const string& exec_if, bool tail,
-      interval interval, const vector<pair<string, string>>& env);
+  script_runner(std::function<void(void)> on_update, string exec, string exec_if, bool tail, interval interval,
```
patrick96, line 15: I don't agree with using rvalues here. You don't gain anything since you still have a copy (now it happens when calling the constructor instead of during the constructor).

Even when using rvalue references, from a performance standpoint, the `script_runner` is only executed when the module starts up and the two strings are generally no more than 100 bytes.

Unless you there is a good reason for this, I'd rather keep using const references.

src/modules/script.cpp
```diff
@@ -25,7 +26,12 @@ namespace modules {
     // Setup formatting
     m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL});
     if (m_formatter->has(TAG_LABEL)) {
-      m_label = load_optional_label(m_conf, name(), "label", "%output%");
+      m_label = load_optional_label(m_conf, name(), TAG_LABEL, "%output%");
+    }
+
+    m_formatter->add(FORMAT_FAIL, TAG_LABEL_FAIL, {TAG_LABEL_FAIL});
```
patrick96, line 32: There is an `add_optional` function in `m_formatter` that makes it easier to have optional formats. Have a look at how the CPU module does this:

https://github.com/polybar/polybar/blob/ab915fb724546e7abfcb35cbaf863ce2bca613ec/src/modules/cpu.cpp#L23-L24

https://github.com/polybar/polybar/blob/ab915fb724546e7abfcb35cbaf863ce2bca613ec/src/modules/cpu.cpp#L94-L100

patrick96 (changes requested):
Works very nicely, thanks a lot! :smiley: 

I mainly have some "cosmetic" nitpicks, otherwise everything looks great.

Please also add an entry in the changelog in the `Added` section: https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog

src/adapters/script_runner.cpp
indev29: I believe copy constructor was used in this case, since it's a better match then non-const rvalue version, thus no warning (a suitable candidate was found). Clang-tidy warns about it though, that's how i discovered this in the first place.

include/adapters/script_runner.hpp
indev29: Well, i think it's kind of a personal preference. Passing an argument by-value offers clearer function interface by explicitly stating that argument will be copied when passed to the function. Const reference, on the other hand, only _suggests_, that it will and doesn't really specify the nature of interaction with the argument.

Performance-wise, when copy of the argument is desired inside a function, passing by value and then moving is better than passing by const reference, since if rvalue is passed, binding to const reference will always copy instead of just moving. Although, as you noted, it doesn't really matter in this case.

```c++
void f1(std::string s);
void f2(const std::string s);

auto s = std::string();
f1(std::move(s)); // move
f2(std::move(s)); // copy
```

The reason I made this change is because I tend to generally follow this principle, mostly because of clearer interface, than performance reasons. Didn't think much there, just an automatic reaction. I'll roll this one back to const references.

src/adapters/script_runner.cpp
patrick96: That makes sense. I really need to add clang-tidy to the CI to catch this sort of thing.

include/adapters/script_runner.hpp
patrick96: Thanks for the explanation :)

I guess it really is a personal preferences. I would rather not expose the fact that a copy is made in the interface because if the copy becomes unnecessary in the future, it changes the interface and you may have to update all users.

src/adapters/script_runner.cpp
```diff
@@ -14,9 +14,9 @@ POLYBAR_NS
 script_runner::script_runner(std::function<void(void)> on_update, const string& exec, const string& exec_if, bool tail,
     interval interval, const vector<pair<string, string>>& env)
     : m_log(logger::make())
-    , m_on_update(on_update)
-    , m_exec(exec)
-    , m_exec_if(exec_if)
+    , m_on_update(std::move(on_update))
+    , m_exec(std::move(exec))
+    , m_exec_if(std::move(exec_if))
```
patrick96, line 19: These can also be undone since these are references.

patrick96 (changes requested):
Alright, just one minor change, then we can merge this :)

patrick96:
Thanks a lot!

Made it just in time for the 3.6 release ;)

Issue #2606, opened 2022-02-25
`custom/script`: repeat inverval for script failure


indev29:
## Why does polybar need this feature?
Now that polybar has options for formatting in case of script failure, it makes sense to add other options for failure handling.

## Describe the solution you'd like
`interval-fail` option, which sets the repeat interval in case of the last script execution failed (exited with non-zero status). If not specified, falls back to the `interval` option.

## Describe alternatives you've considered

1. Module scripts running under `systemd` outputting their results to the temporary files. Polybar modules are just simple scripts that grab the output.
2. Module scripts running as `tail`, managing their lifetime manually.  

These approaches require a lot of work to get running, so having polybar support instead would be much more convenient.

indev29:
I could also implement this one.

indev29:
`interval-if` could also be added (#943).   
With `interval-fail` it becomes unclear which one (`interval` or `interval-fail`) should be used for `exec-if`.

patrick96:
Yeah, that's a good idea :+1:

indev29:
Closed with #2630

PR #2630, opened 2022-03-06
feat(script): add repeat interval for script failure and exec-if

### LINKED ISSUE #943: interval option for exec-if
I would like to have an additional parameter to the `custom/script` module that allows me to set an interval for the `exec-if` command that is different from the standard interval (a backoff timer for when the command fails).

One of my use cases wowuld be [my script for showing information about the train I'm in](https://gist.github.com/Nervengift/6a49faecfc561297ddfe71f32a1c996c): While I want the information to be refreshed every few seconds, most of the time I'm not on a train and so the check should not run too often.

My implementation idea would be:

* add a `exec-if-backoff` parameter (or maybe `interval-exec-if`, `interval-backoff`?)
* if the `exec-if` command fails and `exec-if-backoff` is present, wait for the configured backoff time (instead of `interval`) before runing it again
* when `exec-if` succeeds, execute `exec` then wait for `interval`

indev29:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
Added repeat interval for script failure (`interval-fail`) and `exec-if` (`interval-if`), both falling back to `interval` if not specified. The implementation is straightforward, current configurations retain their behaviour.

## Related Issues & Documents
Closes #943, #2606

The following should be added to [Modules/script/Basic Settings](https://github.com/polybar/polybar/wiki/Module:-script#basic-settings):

```ini
; Interval used when the `exec` command returns with a non-zero exit code 
; If not defined, interval is used instead
interval-fail = 300

; Seconds to sleep between `exec-if` invocations
; If not defined, interval is used instead
interval-if = 180
```

## Additional notes

One thing I've discovered is that the whole exit-code mechanism is not very reliable due to the fact that script invocations and output updates are not synchronized in any way. Here's a sample config to describe the problem:

```ini
[bar/bar]
modules-right = test

[module/test]
type=custom/script

format = <label>
format-background = #090
format-foreground = #fff
format-padding = 4

format-fail = <label-fail>
format-fail-background = #900
format-fail-foreground = #fff
format-fail-padding = 4

label = %output%
label-fail = %output%

exec=echo %counter% $(date +%S) ; ((%counter% % 2))

interval=3
interval-fail=5
```

The script prints it's `%counter%` with date in seconds and fails every other time. Expected output is `green-red-green-red...`.
Here's an example of the output I've got:
```
1 00 (green)
2 03 (green)
3 08 (red)
4 11 (green)
5 16 (green)
6 19 (red)
7 24 (red)
```

As you can see `2` was considered "successful" by formatting, but applied the correct inteval, because it's actually controlled by `script_runner`. What actually happened is this:
```c++
set exit_status: 0 // exec at ~00
// 1
get_output
  get exit_status: 0
// 2
get_output
  get exit_status: 0
set exit_status: 1  // exec at ~03
// 3
get_output
  get exit_status: 1
set exit_status: 0 // exec at ~08
```
The same is for `4-5` and `6-7`.

My point is, it may be confusing for users if their scripts will go for (probably longer) `interval-fail` while being displayed as successful. It's also not an expected behavior really and performs poorly with "flickering" scripts like the one in the example. At the same time I'm not sure If we could do something about it.

Sorry for not noticing this before #2596.


patrick96 (approved):
Looks good! Thanks :smiley:

indev29:
> What do you think about this? Do you want to tackle this?

I think I should, yeah. I'll make a separate PR for these changes.

patrick96:
@indev29 If you work on it, use the `hotifx/3.6.2` branch as the starting point so that we can more easily get it into the next hotfix release.

There has already been a potential report of this issue (even without the intervals): #2650

PR #2664, opened 2022-03-24
feat(pulseaudio): define invert-scroll option

itsfarseen:

* [ ] Refactor
* [X] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
When we enable natural scrolling option in libinput, it sends scroll down even when we swipe up on the touchpad.  
This makes the pulseaudio module feel weird.  
This option fixes that.

Add the following to the wiki under Pulseaudio:

```ini
# Reverses the increment/decrement on scroll event. Set this to true if you are using natural scrolling option on your touchpad.
reverse-scroll = false
```

### DIFF
--- a/src/modules/pulseaudio.cpp
+++ b/src/modules/pulseaudio.cpp
@@ -26,6 +26,7 @@ namespace modules {
 
     auto sink_name = m_conf.get(name(), "sink", ""s);
     bool m_max_volume = m_conf.get(name(), "use-ui-max", true);
+    m_reverse_scroll = m_conf.get(name(), "reverse-scroll", false);
 
     try {
       m_pulseaudio = std::make_unique<pulseaudio>(m_log, move(sink_name), m_max_volume);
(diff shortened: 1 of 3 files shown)


src/modules/pulseaudio.cpp
```diff
@@ -26,6 +26,7 @@ namespace modules {
 
     auto sink_name = m_conf.get(name(), "sink", ""s);
     bool m_max_volume = m_conf.get(name(), "use-ui-max", true);
+    m_invert_scroll = m_conf.get(name(), "invert-scroll", false);
```
patrick96, line 29: Please call the option `reverse-scroll` to be consistent with other modules that also have such an option.

patrick96 (changes requested):
Nice. I didn't know that that's how natural scrolling was implemented, but that actually makes a lot of sense.

Please also add your changes to the changelog: https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog

itsfarseen:
@patrick96, I have renamed the field and added changelog entry.

patrick96 (approved):
Looks good now, thanks!

I went a head and resolved the merge conflicts.

PR #2667, opened 2022-04-01
fix(script): output timing inconsistencies

### LINKED ISSUE #2650: [Bug]: label-fail (from script module) works on random reloads
### Checklist

### Steps to reproduce

I believe label-fail part is still buggy, it seems to work randomly and fails to execute randomly.

1. If you randomly restart polybar, it seems to toggle between working / not working 

### Minimal config

```dosini
[module/restock]
type = custom/script
interval = 300
exec = ~/.scripts/stock.sh
 
label = %output%
label-font = 15
format =  <label>
format-background = ${colors.red}
format-foreground = ${colors.black}
format-padding = 1
format-font = 3
 
label-fail = %output%
label-fail-font = 15
label-fail-foreground = ${colors.black}
format-fail =  <label-fail>
format-fail-background = ${colors.bg-alt}
format-fail-foreground = ${colors.dark-red}
format-fail-padding = 1
format-fail-font = 3
```

### Polybar log

_No response_

### Expected behavior

The label-fail condition doesnt execute everytime, In order to fix it I have to restart polybar randomly.

### Actual behavior

see expected behavior

i3-gaps

### Linux Distribution

Arch Linux

### Polybar version

```text
polybar 3.6.1

Build type: Release
Compiler: /usr/bin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection -Wp,-D_GLIBCXX_ASSERTIONS -O3 -DNDEBUG -Wall -Wextra -Wpedantic -Wsuggest-override
Linker flags: -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -Wall -Wextra -Wpedantic -Wsuggest-override  -Wall -Wextra -Wpedantic -Wsuggest-override
```

_No response_

indev29:

* [ ] Refactor
* [ ] Feature
* [x] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
This PR fixes timing inconsistencies of `script` module output, first described at #2630.  
Module now receives updates from corresponding runner callback, instead of directly calling `get_*` methods of the runner.

Synchronization had to be moved to the module itself. I don't quite like the way I did it here as it still leaves a possibility for inconsistencies to occur (e.g. if runner calls update handler during `get_output` execution).  
The alternative would be to hold lock for the whole execution of `get_output`. It heavily relies on (only) `module::get_output` calling `get_format`, since explicit locking inside `get_format` would lead to deadlock. Although this implementation would eliminate any inconsistencies, I find it unstable architecture-wise. Maybe I've missed some other obvious solutions here.

## Related Issues & Documents
Closes #2650 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Positioning in awesomeWM ([`#2651`](https://github.com/polybar/polybar/pull/2651))
 - `internal/xworkspaces`: The module sometimes crashed polybar when windows were closed. ([`#2655`](https://github.com/polybar/polybar/pull/2655))
 - Mouseover error when only one cursor is defined ([`#2656`](https://github.com/polybar/polybar/pull/2656))
+- Timing inconsistencies with `custom/script` output ([`#2650`](https://github.com/polybar/polybar/issues/2650), first described at [`#2630`](https://github.com/polybar/polybar/pull/2630))
 
 ## [3.6.1] - 2022-03-05
 ### Build
--- a/include/adapters/script_runner.hpp
+++ b/include/adapters/script_runner.hpp
@@ -11,9 +11,18 @@ POLYBAR_NS
 
 class script_runner {
  public:
+  struct data {
+    int counter{0};
+    int pid{-1};
+    int exit_status{0};
+    string output;
+  };
+
+  using on_update = std::function<void(const data&)>;
   using interval = std::chrono::duration<double>;
-  script_runner(std::function<void(void)> on_update, const string& exec, const string& exec_if, bool tail,
-      interval interval, const vector<pair<string, string>>& env);
+
+  script_runner(on_update on_update, const string& exec, const string& exec_if, bool tail, interval interval,
+      const vector<pair<string, string>>& env);
 
   bool check_condition() const;
   interval process();
@@ -22,16 +31,11 @@ class script_runner {
 
   void stop();
 
-  int get_pid() const;
-  int get_counter() const;
-  int get_exit_status() const;
-
-  string get_output();
-
   bool is_stopping() const;
 
  protected:
   bool set_output(string&&);
+  bool set_exit_status(int);
 
   interval run_tail();
   interval run();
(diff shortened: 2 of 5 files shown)


patrick96 (changes requested):
> Synchronization had to be moved to the module itself. I don't quite like the way I did it here as it still leaves a possibility for inconsistencies to occur (e.g. if runner calls update handler during get_output execution).

Yes, that's definitely a problem.

> The alternative would be to hold lock for the whole execution of get_output. It heavily relies on (only) module::get_output calling get_format, since explicit locking inside get_format would lead to deadlock. Although this implementation would eliminate any inconsistencies, I find it unstable architecture-wise. Maybe I've missed some other obvious solutions here.

I also don't think locking the whole `get_output` is a good idea here. However, many other modules also rely on `module::get_output` calling `get_format` and `build`, but I agree that that's not very clean.

I think a good way to tackle this is to copy the data struct into a local variable in the beginning of `get_output` and only use that struct. This resolves all the possible inconsistencies within `get_output`, but not the one with `get_format`.

For `get_format`, we could create a private member variable `m_exit_code` that we set at the beginning of `get_output`. This way `get_format` always relies on the same base-data as `get_output`.

This would also remove the need for individual `get_*` methods for all struct members.

The `xworkspaces` module is an example of another module that sets a member variable in `get_output` that is later used in a method called by `module::get_output` (in this case, the `build` method).

indev29:
> For `get_format`, we could create a private member variable `m_exit_code` that we set at the beginning of `get_output`. This way `get_format` always relies on the same base-data as `get_output`.

Do you mean something like [this](https://github.com/polybar/polybar/pull/2667/commits/558e739eb1429629b7b6aa3952f075f43b0670ce)? I guess we rely on `module::get_output` and don't make it atomic?

patrick96 (approved):
Yes, that's exactly what I had in mind :)

> I guess we rely on module::get_output and don't make it atomic?

Yes, the only thing that doesn't run in the main thread in the script module is `handle_runner_update` and there we properly lock everything.

Issue #2663, opened 2022-03-24
[Bug]: All modules wait for possible double-click events once one module has a double click event set


zack-baker:
### Checklist

- [X] I have read the appropriate section in the [contributing guidelines](https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md)
- [X] I believe this issue is a problem with polybar itself and not a misconfiguration on my part
- [X] I have searched for other open and closed [issues](https://github.com/polybar/polybar/issues?q=is%3Aissue) that may have already reported this problem
- [X] I have checked the [known issues](https://github.com/polybar/polybar/wiki/Known-Issues) page for this problem.
- [X] I have followed the [debugging guide](https://github.com/polybar/polybar/wiki/Debugging-your-Config) to narrow down the problem to a minimal config.

### Steps to reproduce

1. Start polybar with the provided minimal config
2. Click on the xworkspaces module to change workspace   

### Minimal config

```dosini
[bar/main]
modules-center = clicker double-clicker

[module/clicker]
type = internal/xworkspaces

[module/double-clicker]
type = custom/script
label = "|DBLCLICKER|"
exec = echo "a"
double-click-left = notify-send 'Double Click'
```

### Polybar log

```text
polybar|notice:  Parsing config file: /home/zack/.config/polybar/config.ini
polybar|trace: config_parser: Parsing /home/zack/.config/polybar/config.ini
polybar|info:  Loaded monitor HDMI-0 (1920x1080+0+0)
polybar|info:  Configured DPI = 96x96
polybar|info:  Bar geometry: 1920x24+0+0; Borders: 0,0,0,0
polybar|trace: bar: Attach X event sink
polybar|trace: bar: Attach signal receiver
polybar|trace: controller: Setup user-defined modules
polybar|info:  Starting application
polybar|trace: controller: Main thread id = 1
polybar|info:  Starting module/clicker
polybar|info:  Starting module/double-clicker
polybar|info:  Entering event loop (thread-id=1)
polybar|trace: bar: Create renderer
polybar|trace: renderer: Get TrueColor visual
polybar|trace: renderer: Allocate colormap
polybar|trace: renderer: Allocate output window
polybar|info:  script_runner: Invoking shell command: "echo "a""
polybar|trace: renderer: Allocate window pixmaps
polybar|trace: renderer: Allocate graphic contexts
polybar|trace: renderer: Allocate alignment blocks
polybar|trace: renderer: Allocate cairo components
[...]

PR #2695, opened 2022-04-22
fix: check if double click action associated

Zaphoood:

* [ ] Refactor
* [ ] Feature
* [x] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

When a module is clicked, only wait for the double click interval if a double click action is associated with that module. Otherwise trigger the click right away.
The check is button specific, so if the user performs a _right_ click on an area that has an action for a double _left_ click, the right click is triggered immediately.

## Related Issues & Documents

Fixes: #2663

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `custom/text`: Loads the `format` setting, which supports the `<label>` tag, if the deprecated `content` is not defined ([`#1331`](https://github.com/polybar/polybar/issues/1331), [`#2673`](https://github.com/polybar/polybar/pull/2673), [`#2676`](https://github.com/polybar/polybar/pull/2676))
 - Added experimental support for positioning the tray like a module
 
+### Fixed
+- Waiting for double click interval on modules that don't have a double click action ([`#2663`](https://github.com/polybar/polybar/issues/2663), [`#2695`](https://github.com/polybar/polybar/pull/2695))
+
 ## [3.6.2] - 2022-04-03
 ### Fixed
 - `format-offset` being ignored ([`#2643`](https://github.com/polybar/polybar/pull/2643))
--- a/src/components/bar.cpp
+++ b/src/components/bar.cpp
@@ -415,19 +415,12 @@ void bar::parse(string&& data, bool force) {
 
   m_renderer->end();
 
-  const auto check_dblclicks = [&]() -> bool {
-    if (m_action_ctxt->has_double_click()) {
-      return true;
+  m_dblclicks.clear();
+  for (auto&& action : m_opts.actions) {
+    if (static_cast<int>(action.button) >= static_cast<int>(mousebtn::DOUBLE_LEFT)) {
+      m_dblclicks.insert(action.button);
     }
-
-    for (auto&& action : m_opts.actions) {
-      if (static_cast<int>(action.button) >= static_cast<int>(mousebtn::DOUBLE_LEFT)) {
-        return true;
-      }
-    }
-    return false;
-  };
-  m_dblclicks = check_dblclicks();
+  }
 }
 
 /**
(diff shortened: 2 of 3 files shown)


src/components/bar.cpp
```diff
@@ -818,9 +810,27 @@ void bar::handle(const evt::button_press& evt) {
...
+      has_dblclick = m_dblclicks.at(mousebtn::DOUBLE_MIDDLE) or
+                     m_action_ctxt->has_action(mousebtn::DOUBLE_MIDDLE, pos) != tags::NO_ACTION;
+      break;
+    case mousebtn::RIGHT:
+      has_dblclick = m_dblclicks.at(mousebtn::DOUBLE_RIGHT) or
+                     m_action_ctxt->has_action(mousebtn::DOUBLE_RIGHT, pos) != tags::NO_ACTION;
+      break;
+    default:
+      break;
+  }
```
patrick96, line 829: Could you maybe simply this using the `mousebtn_get_double` function? That way, we don't need the whole switch statement here. Also please use `||` instead of `or`.

src/components/bar.cpp
```diff
@@ -415,19 +415,11 @@ void bar::parse(string&& data, bool force) {
...
-  const auto check_dblclicks = [&]() -> bool {
-    if (m_action_ctxt->has_double_click()) {
-      return true;
-    }
-
-    for (auto&& action : m_opts.actions) {
-      if (static_cast<int>(action.button) >= static_cast<int>(mousebtn::DOUBLE_LEFT)) {
-        return true;
-      }
+  for (auto&& action : m_opts.actions) {
```
patrick96, line 418: Note to self: Global bar click actions should somehow be integrated into the action context at some point. It would make a lot of code here cleaner.

patrick96 (changes requested):
Very nice! Thanks for your help :smiley: 

Please also add this bugfix to the [changelog](https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog).

src/components/bar.cpp
Zaphoood: Thanks for the feedback, I applied the changes you suggested

src/components/bar.cpp
```diff
@@ -818,9 +810,13 @@ void bar::handle(const evt::button_press& evt) {
     }
   };
 
+  bool has_dblclick = false;
+  mousebtn double_btn = mousebtn_get_double(btn);
+  has_dblclick = m_dblclicks.at(double_btn) || m_action_ctxt->has_action(double_btn, pos) != tags::NO_ACTION;
```
patrick96, line 815: This will throw an exception if `mousebtn_get_double` returns `NONE` (this happens for scroll events). It may be easier to have `m_dblclicks` be a `std::set<mousebtn>` and just check if the set contains the given double click button

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `custom/text`: Loads the `format` setting, which supports the `<label>` tag, if the deprecated `content` is not defined ([`#1331`](https://github.com/polybar/polybar/issues/1331), [`#2673`](https://github.com/polybar/polybar/pull/2673), [`#2676`](https://github.com/polybar/polybar/pull/2676))
 - Added experimental support for positioning the tray like a module
 
+### Fixed
+- Waiting for double click interval on modules that don't have a double click action ([`#2695`](https://github.com/polybar/polybar/pull/2695))
```
patrick96, line 25: Please also add a link to the original issue https://github.com/polybar/polybar/issues/2663

patrick96 (changes requested):

src/components/bar.cpp
```diff
@@ -415,19 +415,12 @@ void bar::parse(string&& data, bool force) {
...
-    if (m_action_ctxt->has_double_click()) {
-      return true;
-    }
-
-    for (auto&& action : m_opts.actions) {
-      if (static_cast<int>(action.button) >= static_cast<int>(mousebtn::DOUBLE_LEFT)) {
-        return true;
-      }
+  m_dblclicks.clear();
+  for (auto&& action : m_opts.actions) {
```
Zaphoood, line 419: As far as I can tell, `m_opts.actions` is only set once in `bar::bar` so we could put this loop there. Tough I'm not sure if `m_opts.actions` might change somwhere else

patrick96: Doesn't really matter. Leave it here for now.

patrick96 (approved):
Looks good!

PR #2714, opened 2022-05-05
fix: display invalid tags as they are instead of hiding them

### LINKED ISSUE #2713: [Bug]: invalid format tags don't throw exceptions
### Checklist

### Steps to reproduce

1. Run polybar using the provided minimal config

### Minimal config

```dosini
[bar/i3]
modules-center = date

[module/date]
type = internal/date
date = %Y
format = <this-is-not-a-valid-label>
```

### Polybar log

```text
polybar|notice:  Parsing config file: ../tests/bad-config.ini
polybar|warn:  No monitor specified, using "HDMI-0"
polybar|info:  Loaded monitor HDMI-0 (1920x1080+0+0)
polybar|info:  Configured DPI = 96x96
polybar|info:  Bar geometry: 1920x24+0+0; Borders: 0,0,0,0
polybar|notice:  Loading module 'date' of type 'internal/date'
polybar|notice:  Loaded 1 modules
polybar|info:  Starting application
polybar|info:  Starting module/date
polybar|info:  Entering event loop (thread-id=1)
polybar|warn:  No fonts specified, using fallback font "fixed"
polybar|notice:  Loaded font "fixed" (name=Noto Sans, offset=0, file=/usr/share/fonts/noto/NotoSans-Regular.ttf)
polybar|info:  Bar window: 0x3200001
polybar|info:  Disabling tray manager (reason: missing `tray-position`)
polybar|info:  module/date: Rebuilding cache
polybar|info:  Redrawing bar window
polybar|notice:  Received signal(2): Interrupt
polybar|info:  Eventloop finished
polybar|info:  module/date: Stopping
polybar|info:  Deconstruction of module/date took 0 ms.
polybar|info:  Waiting for spawned processes to end
polybar|info:  Reached end of application...
```

### Expected behavior

An exception should be thrown, something like `<this-is-not-a-valid-label> is not a valid format tag for "format"`

### Actual behavior

No exception is thrown

i3 version 4.20.1 (2021-11-03)

### Linux Distribution

Manjaro 5.15.32-1

### Polybar version

```text
polybar 3.6.2

Build type: Release
Compiler: /usr/bin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection -Wp,-D_GLIBCXX_ASSERTIONS -flto=auto -O3 -DNDEBUG -Wall -Wextra -Wpedantic -Wsuggest-override
Linker flags: -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -flto=auto -Wall -Wextra -Wpedantic -Wsuggest-override  -Wall -Wextra -Wpedantic -Wsuggest-override
```

I'm pretty sure the reason for this behavior is in `src/modules/meta/base.cpp`, at line 95:
```
format->value = move(value);
```
`value` is then later used to check for invalid tags, but it's value is now abritrary since it has been moved, which is why invalid tags aren't detected. Copying the value of `value` instead of moving should fix that.

Zaphoood:

* [ ] Refactor
* [ ] Feature
* [x] Bug Fix
* [ ] Optimization
* [ ] Documentation Update
* [ ] Other

## Description

If a formatting string contains an invalid/unsupported tag, it is displayed on the bar as-is in order to make the user aware that there's a mistake in their config.
Before, invalid tags would just be hidden.

## Related Issues & Documents

Fixes #2713

### DIFF
--- a/include/modules/meta/base.inl
+++ b/include/modules/meta/base.inl
@@ -251,8 +251,27 @@ namespace modules {
       }
 
       string tag = value.substr(start, end - start + 1);
-      bool tag_built = CONST_MOD(Impl).build(&tag_builder, tag);
-      string tag_content = tag_builder.flush();
+      bool tag_built{false};
+      string tag_content;
+      if (format->tags.find(tag) != format->tags.end()) {
+        tag_built = CONST_MOD(Impl).build(&tag_builder, tag);
+        tag_content = tag_builder.flush();
+        cursor = end + 1;
+      } else {
+        tag_built = false;
+        /*
+         * Append remaining non-tag content between invalid tag and next tag,
+         * since some whitespace would get cut off by `ltrim` during the next iteration
+         */
+        size_t next = value.find('<', end + 1);
+        if (next != string::npos) {
+          non_tag += value.substr(start, next - start);
+          cursor = next;
+        } else {
+          non_tag += tag;
+          cursor = end + 1;
+        }
+      }
 
       /*
        * Remove exactly one space between two tags if the second tag was not built.
@@ -274,8 +293,6 @@ namespace modules {
         m_builder->node(tag_content);
         has_tags = true;
       }
-
-      cursor = end + 1;
     }
 
     if (cursor < value.size()) {
--- a/src/modules/meta/base.cpp
+++ b/src/modules/meta/base.cpp
@@ -86,7 +86,7 @@ namespace modules {
   // }}}
   // module_formatter {{{
 
-  void module_formatter::add_value(string&& name, string&& value, vector<string>&& tags, vector<string>&& whitelist) {
+  void module_formatter::add_value(string&& name, string&& value, std::set<string>&& tags) {
     const auto formatdef = [&](const string& param, const auto& fallback) {
       return m_conf.get("settings", "format-" + param, fallback);
     };
@@ -104,7 +104,7 @@ namespace modules {
     format->margin = m_conf.get(m_modname, name + "-margin", formatdef("margin", format->margin));
     format->offset = m_conf.get(m_modname, name + "-offset", formatdef("offset", format->offset));
     format->font = m_conf.get(m_modname, name + "-font", formatdef("font", format->font));
-    format->tags.swap(tags);
+    format->tags = move(tags);
 
     try {
       format->prefix = load_label(m_conf, m_modname, name + "-prefix");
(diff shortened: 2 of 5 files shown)


include/modules/meta/base.inl
```diff
@@ -249,31 +249,18 @@ namespace modules {
...
-      if (tag_built) {
-        if (has_tags) {
-          // format-spacing is added between all tags
-          m_builder->spacing(format->spacing);
-        }
-
-        m_builder->node(tag_content);
-        has_tags = true;
-      }
+      m_builder->node(tag_built ? tag_content : tag);
```
patrick96, line 263: Unfortunately we can't rely on the return value of `build` for this, I should have mentioned this in the original issue because I already tried that.

There are some modules that return `false` for valid tags (e.g. the xkeyboard module for `label-indicator` if there are no indicators. In that case displaying `<label-indicator>` is an error.

Try the following xkeyboard module:

```dosini
[module/xkeyboard]
type = internal/xkeyboard
blacklist-0 = num lock
blacklist-1 = scroll lock
blacklist-2 = caps lock
format = <label-indicator>
label-layout = %name%
```

We have to check `format->tags` to see if a tag is valid.

include/modules/meta/base.inl
```diff
@@ -249,31 +249,18 @@ namespace modules {
...
-
-      if (tag_built) {
-        if (has_tags) {
-          // format-spacing is added between all tags
-          m_builder->spacing(format->spacing);
-        }
-
-        m_builder->node(tag_content);
-        has_tags = true;
-      }
```
patrick96, line 276: The logic here should not be changed. Any change here that doesn't preserve the exact behavior may add or remove spacing in all of people's modules.

patrick96 (changes requested):
We unfortunately really need to use `format->tags` to check validity. I wish we could simplify the spacing calculations but it could potentially break a lot of configs.

include/modules/meta/base.inl
Zaphoood: I see, I didn't catch that. In that case, it would be convenient to have `format->tags` be a set instead, or is there something that relies on it being a vector?

patrick96: I don't think anything else uses it. Feel free to use whatever data structure you want (or completely throw it away and do something else :wink:)

src/modules/meta/base.cpp
```diff
@@ -1,6 +1,7 @@
 #include "modules/meta/base.hpp"
 
 #include <utility>
+#include <iterator>
```
patrick96, line 4: This may not be necessary

include/modules/meta/base.inl
```diff
@@ -250,9 +250,16 @@ namespace modules {
...
+      bool tag_built{false};
       string tag = value.substr(start, end - start + 1);
-      bool tag_built = CONST_MOD(Impl).build(&tag_builder, tag);
-      string tag_content = tag_builder.flush();
+      if (format->tags.find(tag) != format->tags.end()) {
+        tag_built = CONST_MOD(Impl).build(&tag_builder, tag);
+        tag_content = tag_builder.flush();
+      } else {
+        tag_built = true;
+        tag_content = tag;
```
patrick96, line 261: In this case, it should probably set `tag_built = false` and add the tag name to `non_tag`.

Because right now, an invalid tag also applies `format->spacing`.
For example if you have

```dosini
format = <label> <invalid>
format-spacing = 3
label = Foo
```

This will be rendered as `Foo    <invalid>` instead of `Foo <invalid>`

patrick96 (changes requested):

PR #2716, opened 2022-05-06
feat(xkeyboard): add %shortname% token

ckiee:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
This takes a string like `English (US)` or `Hebrew` that is produced
by `%name%` and shortens it into a `en` or `he`. This is the same behavior
that used to occur before some upstream changes made `%layout%` useless
on some systems. `%variant%` is *not* a valid replacement due to some layouts 
seemingly not having a variant. (I do not usually touch this, it just annoyed me enough to go patch it out)

## Related Issues & Documents
A quick search didn't find any. 

  - Need to add to the available tokens list for `label-layout` [here](https://github.com/polybar/polybar/wiki/Module:-xkeyboard#additional-formatting).
* [ ] (I don't know) This PR requires changes to the documentation inside the git repo (please add them to the PR).

### DIFF
--- a/src/modules/xkeyboard.cpp
+++ b/src/modules/xkeyboard.cpp
@@ -112,6 +112,9 @@ namespace modules {
     if (m_layout) {
       m_layout->reset_tokens();
       m_layout->replace_token("%name%", m_keyboard->group_name(m_keyboard->current()));
+      auto short_group_name = m_keyboard->group_name(m_keyboard->current());
+      std::transform(short_group_name.begin(), short_group_name.end(), short_group_name.begin(), ::tolower);
+      m_layout->replace_token("%shortname%", short_group_name.substr(0, 2));
       m_layout->replace_token("%variant%", m_keyboard->variant_name(m_keyboard->current()));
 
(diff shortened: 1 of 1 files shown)


patrick96:
Do you have any references for those upstream changes that broke `%layout%`?

I am wondering if there are better ways to solve this. Because just lowercasing the first two letters in `%name%` doesn't work in general. For me, `%name%` is `German (Switzerland)` and `%shortname%` would produce `ge` even though you would expect either `de` or `ch`.

Since %layout% is useless for you, what kind of output does it produce for the keyboard layouts you tried on your system?

ckiee:
@patrick96 So.. I restarted my machine since making this PR and now `%layout%` exhibits the same results as `%shortname%`. I guess not restarting the xserver after upgrading my system made some version mismatch.

patrick96:
Oh no. This shows that turning it off and on again still solves the most problems :sweat_smile:

Issue #2689, opened 2022-04-15
Stabilize tray module


patrick96:
With #2595 being merged, there are a few steps to make the tray module stable.
The goal is to completely deprecate positioning the tray without a module (and all `tray-*` settings in the bar section).

* [x] Automatically hide/show the tray window depending on whether a tray position (`%{Pt}` tag) is set in the renderer. This is needed to deal with hiding modules.
* [x] Remove `tray-position = adaptive` again and just activate the tray when a tray module exists (or one of the other positions is set). The existence of a tray module should override any value for `tray-position`. In particular, if a tray module exists *and* `tray-position` is set, a warning or error should be printed.
* [x] Fix the tray window being pushed off the bar if there is too much stuff on the bar. The tray window should always be fully within the bar window.
* [x] Rework all tray settings in module (see below)
* [x] Write module documentation on RTD
  * [x] Write migration docs on RTD.
* [x] Emit deprecation warnings for `tray-position` and all other tray settings in the bar section. #3002

It's probably easiest to implement each of these in a separate PR to keep complexity low. The order shouldn't matter except that the last one should be done last.

If we don't make the tray stable before the 3.7 release, polybar should also emit a warning whenever the tray module is used so that people are aware that the tray module may break again in the future.

---

For those wanting to test it out right now, I'll also add a minimal example for how to add the tray module. All of this is experimental and any config options may change before the tray becomes stable.

```dosini
[bar/example]
modules-left = tray

[module/tray]
type = internal/tray
format = <tray>
```

---

We could now also completely change the tray behavior if the module is used without it being a breaking change. In the `tray_manager` we could have a flag `in_module` that can be used to differentiate between behavior for the old tray or the module.
We should consider taking advantage of this, especially for formatting.

* [x] There should be `tray-spacing`, which determines the spacing between the icons (`tray-padding` kind of does this already but the padding is also added at the beginning and end of the tray window.
* [x] `tray-padding` should now be applied to the left and right of each individual icon (just as what happens for regular labels). Only supports points and pixels
* [x] `tray-maxsize` and `tray-scale` should be merged into a single option. They both affect `client_height`, while `tray-maxsize` also increases spacing. This can be made cleaner. There should probably be just a single `tray-size` option (percentage with offset relative to bar height, defaults to 66%) and the final icon height and width is `min(bar_heigh, tray-size)`.
* [x] Use of `tray-background` should be discouraged. Make it clear that it will only affect tha background of individual icons, not the spacing in between.

[...]

PR #2726, opened 2022-05-22
tray: removed tray-module = adaptive option.

raffael0:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
This PR removes the `tray-position` option `adaptive` which was temporarily added in #2595.
It's part of a series of changes outlined in #2689 . Now the tray module is automatically activated when it's added as a module

## Related Issues & Documents

Part of #2689

### DIFF
--- a/src/components/controller.cpp
+++ b/src/components/controller.cpp
@@ -594,6 +594,9 @@ size_t controller::setup_modules(alignment align) {
       auto type = m_conf.get("module/" + module_name, "type");
 
       if (type == tray_module::TYPE) {
+        if (!m_tray_module_name.empty()) {
+          throw module_error("Multiple trays defined. Using tray `" + m_tray_module_name + "`");
+        }
         m_tray_module_name = module_name;
       }
 
--- a/src/components/controller.cpp
+++ b/src/components/controller.cpp
@@ -267,7 +267,7 @@ void controller::read_events(bool confwatch) {
     }
 
     if (!m_writeback) {
-      m_bar->start();
+      m_bar->start(m_tray_module_name);
     }
 
     /*
@@ -592,6 +592,10 @@ size_t controller::setup_modules(alignment align) {
 
     try {
       auto type = m_conf.get("module/" + module_name, "type");
+
+      if (type == tray_module::TYPE) {
+        m_tray_module_name = module_name;
+      }
 
       if (type == ipc_module::TYPE && !m_has_ipc) {
         throw application_error("Inter-process messaging needs to be enabled");
--- a/src/x11/tray_manager.cpp
+++ b/src/x11/tray_manager.cpp
@@ -75,17 +75,23 @@ void tray_manager::setup(string tray_module_name) {
   try {
     position = conf.get(bs, "tray-position");
   } catch (const key_error& err) {
-    return m_log.info("Disabling tray manager (reason: missing `tray-position`)");
   }
 
-  if (position == "left") {
+  if (!position.empty() && !tray_module_name.empty()) {
+    m_log.warn("Tray position is manually defined(`tray-position`) and also set by tray module");
+  }
+  if (position.empty() && tray_module_name.empty()) {
+    return m_log.info("Disabling tray manager (reason: tray module not defined)");
+  }
+
+  if (!tray_module_name.empty()) {
+    m_opts.adaptive = true;
+  } else if (position == "left") {
     m_opts.align = alignment::LEFT;
   } else if (position == "right") {
     m_opts.align = alignment::RIGHT;
   } else if (position == "center") {
     m_opts.align = alignment::CENTER;
-  } else if (position == "adaptive") {
-    m_opts.adaptive = true;
   } else if (position != "none") {
     return m_log.err("Disabling tray manager (reason: Invalid position \"" + position + "\")");
   } else {
--- a/src/x11/tray_manager.cpp
+++ b/src/x11/tray_manager.cpp
@@ -67,7 +67,7 @@ tray_manager::~tray_manager() {
   deactivate();
 }
 
-void tray_manager::setup() {
+void tray_manager::setup(string tray_module_name) {
   const config& conf = config::make();
   auto bs = conf.section();
   string position;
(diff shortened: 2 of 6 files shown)


src/components/controller.cpp
```diff
@@ -593,6 +593,10 @@ size_t controller::setup_modules(alignment align) {
     try {
       auto type = m_conf.get("module/" + module_name, "type");
 
+      if (type == tray_module::TYPE) {
+        m_tray_module_name = module_name;
+      }
```
patrick96, line 598: If we find a tray module and the tray module name is already set, we should print an error telling the user there can only be one tray module

patrick96: Doesn't really matter which tray module is then picked in the end, whatever is easier (but do tell the user in the error message).

Also, not starting superfluous modules would be a good idea as well.

src/x11/tray_manager.cpp
```diff
@@ -67,25 +67,31 @@ tray_manager::~tray_manager() {
...
   } catch (const key_error& err) {
-    return m_log.info("Disabling tray manager (reason: missing `tray-position`)");
   }
 
-  if (position == "left") {
+  if (!position.empty() && !tray_module_name.empty()) {
+    m_log.warn("Tray position is manually defined(`tray-position`) and also set by tray module");
+  }
+  if (position.empty() && tray_module_name.empty()) {
+    return m_log.info("Disabling tray manager (reason: tray module not defined)");
```
patrick96, line 84: No need to change it, I'm just wondering. What is the reason behind calling return on the log call? Just to save one line?

src/x11/tray_manager.cpp
```diff
@@ -67,25 +67,31 @@ tray_manager::~tray_manager() {
...
 
   try {
     position = conf.get(bs, "tray-position");
   } catch (const key_error& err) {
-    return m_log.info("Disabling tray manager (reason: missing `tray-position`)");
   }
 
-  if (position == "left") {
+  if (!position.empty() && !tray_module_name.empty()) {
+    m_log.warn("Tray position is manually defined(`tray-position`) and also set by tray module");
```
patrick96, line 81: In that case, we should tell the user that `tray-position` is ignored.

src/x11/tray_manager.cpp
```diff
@@ -67,25 +67,31 @@ tray_manager::~tray_manager() {
...
+  }
+  if (position.empty() && tray_module_name.empty()) {
+    return m_log.info("Disabling tray manager (reason: tray module not defined)");
+  }
+
+  if (!tray_module_name.empty()) {
+    m_opts.adaptive = true;
```
patrick96, line 88: We now have to mutually exclusive tray settings: `adaptive` and `align`.
Whenever one is set, the other is not and they kind of control the same thing (tray position).

In my opinion it would be cleaner to have a single setting for both. For example a new `tray_pos` enum with `NONE`, `LEFT`, `CENTER`, `RIGHT`, and `MODULE`.

patrick96 (changes requested):
Looks good so far :)

Comments are mainly about user-facing error messages. And more architectural comment.

src/x11/tray_manager.cpp
raffael0: Looking at it again this doesn't really make sense. Wouldn't a error be better here since we're terminating the module(eg. via a module_error?)

patrick96: Hmm, before the tray would print an error if `tray-position` was empty or not set. I guess it would be better to just use `none` as the default value for `tray-position` and then we could just use the big if-else-if block just below to determine whether the tray should run or not. Then we wouldn't even need this check here.

But also, not having a tray is in itself not an error, so info logging is fine unless the user specified a wrong value for `tray-position`.

src/x11/tray_manager.cpp
```diff
@@ -67,25 +67,30 @@ tray_manager::~tray_manager() {
...
   const config& conf = config::make();
   auto bs = conf.section();
-  string position;
+  string position = "none";
 
   try {
     position = conf.get(bs, "tray-position");
   } catch (const key_error& err) {
-    return m_log.info("Disabling tray manager (reason: missing `tray-position`)");
   }
```
patrick96, line 78: This can be replaced with `string position = conf.get(bs, "tray-position", "none");`

src/x11/tray_manager.cpp
```diff
@@ -67,25 +67,30 @@ tray_manager::~tray_manager() {
...
 
-  if (position == "left") {
-    m_opts.align = alignment::LEFT;
+  if (!position.empty() && !tray_module_name.empty()) {
+    m_log.warn(
+        "Tray position is manually defined(`tray-position`) and also set by tray module. `tray-position` will be "
+        "ignored");
```
patrick96, line 83:
```suggestion
    m_log.warn("The tray position is manually defined (`tray-position`) and also set by the tray module (%s). `tray-position` will be ignored", tray_module_name);
```

src/x11/tray_manager.cpp
```diff
@@ -67,25 +67,30 @@ tray_manager::~tray_manager() {
...
 
   try {
     position = conf.get(bs, "tray-position");
   } catch (const key_error& err) {
-    return m_log.info("Disabling tray manager (reason: missing `tray-position`)");
   }
 
-  if (position == "left") {
-    m_opts.align = alignment::LEFT;
+  if (!position.empty() && !tray_module_name.empty()) {
```
patrick96, line 80: This will trigger whenever a tray module exists since `tray-position` will almost never be empty. I think it should check whether position is empty and that is it not `none`.

patrick96 (changes requested):

src/x11/tray_manager.cpp
```diff
@@ -67,25 +67,26 @@ tray_manager::~tray_manager() {
   deactivate();
 }
 
-void tray_manager::setup() {
+void tray_manager::setup(const string& tray_module_name) {
   const config& conf = config::make();
   auto bs = conf.section();
-  string position;
+  string position = conf.get(bs, "tray-position", string("none"));
```
patrick96, line 73:
```suggestion
  string position = conf.get(bs, "tray-position", "none"s);
```

We use `""s` for string literals

patrick96 (changes requested):

patrick96 (approved):

patrick96:
Thanks a lot! :smiley:

PR #2748, opened 2022-06-19
fix(tag parser): use %% to get a literal percent sign (%)

xi:
As described in the lemonbar docs: https://github.com/LemonBoy/bar#formatting

* [ ] Refactor
* [ ] Feature
* [X] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

The [lemonbar documentation](https://github.com/LemonBoy/bar#formatting) says that you can

> Use `%%` to get a literal percent sign (`%`).

That was not implemented in polybar.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Waiting for double click interval on modules that don't have a double click action ([`#2663`](https://github.com/polybar/polybar/issues/2663), [`#2695`](https://github.com/polybar/polybar/pull/2695))
 - Error reporting for deprecated config values ([`#2724`](https://github.com/polybar/polybar/issues/2724))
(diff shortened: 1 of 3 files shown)


patrick96:
This is going to break a lot of things. Take for example this config:

```dosini
[bar/example]
modules-left = text

[module/text]
type = custom/text
format = Text%
```

For this, polybar will produce the string `%{l}Text%%{PR}` which is then rendered as `Text%{PR}`, even though we would expect it to render as `Text%`.

The literal percentage sign in `format` unintentionally escapes formatting tags that follow the text module.

I'm not sure adding literal percentage signs like this is possible at all in a backwards-compatible way.

vincentbernat:
The issue is that it is not clear how to avoid some arbitrary code execution. Suppose I have a Bluetooth device named `%{A1:rm -rf ~:}Headset%{A}`. How to secure this string? I was doing %-escape, but I noticed this didn't work as expected. I am now replacing `%` with `%` with zero-width-space appended to it.

Issue #2572, opened 2022-01-17
Hardware settings auto-detection


patrick96:
I already elaborated on this in https://github.com/polybar/polybar/issues/2405#issuecomment-809272880 but it deserves its own issue(s)

If possible, any module that requires some kind of hardware information (e.g. network interface name) should auto-detect these and choose a reasonable default if the user doesn't specify anything.

First, I'll present the current state. We can update this table as this is implemented in more modules.

|| Module | Notes |
|-|--------|-------|
| :heavy_check_mark:  | `internal/alsa` | Uses default soundcard and `Master` mixer |
| :white_large_square: | `internal/backlight` | |
| :white_large_square: | `internal/battery` | |
| :white_large_square: | `internal/fs` | |
| :heavy_check_mark: | `internal/mpd` | No detection, but uses the standard port used by MPD |
| :heavy_check_mark: | `internal/network` | `interface-type` has to be defined |
| :heavy_check_mark: | `internal/pulseaudio` | Uses default sink |
| :heavy_check_mark: | `internal/temperature` | No detection, but uses thermal zone 0 by default |
| :heavy_check_mark: | `internal/xbackligh` | Uses bar monitor by default |

## Backlight

Here, we should just scan `/sys/class/backlight/` for a directory that has the required files.
If multiple match, just chose the first one and print a notice.

Documentation: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/ABI/testing/sysfs-class-backlight

## Battery

Scan `/sys/class/power_supply/` and chose one that has the necessary files.
If multiple match, just chose the first one and print a notice.

A battery must at least have a file `type` that contains `Battery` to be considered.

There is also a `present` file that tells us if the battery is present.
Though from my own testing, the entire directory disappears if the battery is removed.
If this file is considered, we could also have `format-absent` that is used if that file is `0`, the folder doesn't exist or no battery was found.

The `type` file identifies the type of device, we should only consider `Battery`.

Documentation: 

* https://www.kernel.org/doc/html/latest/power/power_supply_class.html
* https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/ABI/testing/sysfs-class-power

## Filesystem

Scan `/proc/self/mountinfo` and use all entries whose mount source (column 10) starts with `/dev/`

Alternatively, we could always use `/` as the default.

Documentation: `man 5 proc`

Zaphoood:
[...]

PR #2752, opened 2022-06-22
feat(temperature): Add `zone-type` setting

xphoniex:

* [x] Feature
* [x] Bug Fix

## Description

CPU's `thermal-zone` changes on every boot on my machine, sometimes it's `1` and other times it's `2`. Others also have had this issue and resorted to scripts updating the `thermal-zone` before starting `polybar`.

This can simply be fixed using a `zone-type` config option:

If you look at the content in `/sys/class/thermal/thermal_zone%zone%/type`, you'll get the name of the sensor and you can easily deduct zone from that. For example for Intel CPUs, you should set `zone-type = x86_pkg_temp` and you'll be set.

## Related Issues & Documents

Ref #2572

### DIFF
--- a/src/modules/temperature.cpp
+++ b/src/modules/temperature.cpp
@@ -16,11 +16,32 @@ namespace modules {
   temperature_module::temperature_module(const bar_settings& bar, string name_)
       : timer_module<temperature_module>(bar, move(name_)) {
     m_zone = m_conf.get(name(), "thermal-zone", 0);
+    m_zone_type = m_conf.get(name(), "zone-type", ""s);
     m_path = m_conf.get(name(), "hwmon-path", ""s);
     m_tempbase = m_conf.get(name(), "base-temperature", 0);
     m_tempwarn = m_conf.get(name(), "warn-temperature", 80);
     set_interval(1s);
     m_units = m_conf.get(name(), "units", m_units);
+
+    if (!m_zone_type.empty()) {
+      bool zone_found = false;
+      vector<string> zone_paths = file_util::glob(PATH_THERMAL_ZONE_WILDCARD);
+      vector<string> available_zones;
+      for (auto &z: zone_paths) {
+        string zone_file = z + "/type";
+        string z_zone_type = string_util::strip_trailing_newline(file_util::contents(zone_file));
+        available_zones.push_back(z_zone_type);
+        if (z_zone_type == m_zone_type) {
+          m_path = z + "/temp";
+          zone_found = true;
+          break;
+        }
+      }
+
+      if (!zone_found) {
+        throw module_error("zone-type '" + m_zone_type +  "' was not found, available zone types: " + string_util::join(available_zones, ", "));
+      }
+    }
 
     if (m_path.empty()) {
       m_path = string_util::replace(PATH_TEMPERATURE_INFO, "%zone%", to_string(m_zone));
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `custom/text`: Loads the `format` setting, which supports the `<label>` tag, if the deprecated `content` is not defined ([`#1331`](https://github.com/polybar/polybar/issues/1331), [`#2673`](https://github.com/polybar/polybar/pull/2673), [`#2676`](https://github.com/polybar/polybar/pull/2676))
 - Added experimental support for positioning the tray like a module
 - `internal/backlight`: `scroll-interval` option ([`#2696`](https://github.com/polybar/polybar/issues/2696), [`#2700`](https://github.com/polybar/polybar/pull/2700))
+- `internal/temperature`: Added `zone-type` setting ([`#2572`](https://github.com/polybar/polybar/issues/2572), [`#2752`](https://github.com/polybar/polybar/pull/2752)) by [@xphoniex](https://github.com/xphoniex)
 
 ### Changed
 - `internal/fs`: Use `/` as a fallback if no mountpoints are specified ([`#2572`](https://github.com/polybar/polybar/issues/2572), [`#2705`](https://github.com/polybar/polybar/pull/2705))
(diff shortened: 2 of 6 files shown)


patrick96:
I'm not around right now, so can't really review this. Will review once I'm back and have time. But general idea looks good. Except that you just try the first 10 zones instead of listing files in the folder and iterating through those.

xphoniex:
> I'm not around right now, so can't really review this. Will review once I'm back and have time. But general idea looks good. Except that you just try the first 10 zones instead of listing files in the folder and iterating through those.

Okay, cool.

Re iterations, rewrote it to use `INT_MAX` ([here](https://github.com/polybar/polybar/compare/eb369e41f99b73a63f0a85d28277d05c91ed79d5..5b96d0f01dc08d9d434b6fc5baaa76e0d472b92e)), this is simpler than listing the files (no allocation needed, and no need to then extract zone # from the file string).

I have a few useful contributions which I can open PR for once this is merged.

patrick96:
I'm not the biggest fan of iterating until no file is found. Is it guaranteed that all `thermal_zoneX` appear in order without gaps?

Also, the error message for when no zone is found is very confusing:

```
The file '/sys/class/thermal/thermal_zone1/type' does not exist
```

The error message should tell the user which zone type it failed to find (and maybe even list all available zone types).

---

> I have a few useful contributions which I can open PR for once this is merged.

Please open a [discussion topic](https://github.com/polybar/polybar/discussions/categories/ideas) for those. That way we can see how those will fit into polybar.

xphoniex:
> I'm not the biggest fan of iterating until no file is found. Is it guaranteed that all `thermal_zoneX` appear in order without gaps?
> 
> Also, the error message for when no zone is found is very confusing:
> 
> ```
> The file '/sys/class/thermal/thermal_zone1/type' does not exist
> ```
> 
> The error message should tell the user which zone type it failed to find (and maybe even list all available zone types).
> 

There's [this](https://github.com/torvalds/linux/blob/268db333b561c77dee3feb6831806412293b4a7e/tools/lib/thermal/thermal.c#L34-L45) in kernel, but still not sure about gaps as there are ways to unregister sensors.

I suppose you've seen this issue before so I implemented what you suggested, check the [diff](https://github.com/polybar/polybar/compare/5b96d0f01dc08d9d434b6fc5baaa76e0d472b92e..6ced7cead38b9754ecd75c3628009a9aa9b2bf69) please.

xphoniex:
reminder

src/modules/temperature.cpp
```diff
@@ -16,12 +17,42 @@ namespace modules {
...
+        available_zones += " " + z_zone_type;
+        if (z_zone_type == m_zone_type) {
+          m_zone = z;
+          zone_found = true;
+          break;
+        }
+      }
+
+      if (!zone_found) {
+        throw module_error("zone-type '" + m_zone_type +  "' was not found, available zone types:" + available_zones);
```
patrick96, line 52:
```suggestion
        throw module_error("zone-type '" + m_zone_type +  "' was not found, available zone types: " + available_zones);
```

src/modules/temperature.cpp
```diff
@@ -16,12 +17,42 @@ namespace modules {
...
+        }
+      }
+      sort(zone_list.begin(), zone_list.end());
+
+      bool zone_found = false;
+      string available_zones;
+      for (auto &z: zone_list) {
+        string zone_file = string_util::replace(PATH_THERMAL_ZONE_TYPE, "%zone%", to_string(z));
+        string z_zone_type = string_util::strip_trailing_newline(file_util::contents(zone_file));
+        available_zones += " " + z_zone_type;
```
patrick96, line 43: Minor: Instead of manually constructing the string, we could store the types in a vector and finally construct the string using. `string_util::join(available_zones, ", ")`

src/modules/temperature.cpp
```diff
@@ -16,12 +17,42 @@ namespace modules {
...
 
+    if (!m_zone_type.empty()) {
+      vector<int> zone_list;
+      vector<string> thermal_list = file_util::list_files("/sys/class/thermal/");
+      for (auto &thermal: thermal_list) {
+        string zone_str = regex_replace(thermal, std::regex("thermal_zone([0-9]+)"), string("$1"), std::regex_constants::format_no_copy);
+        if (!zone_str.empty()) {
+          zone_list.push_back(stoi(zone_str));
+        }
+      }
```
patrick96, line 35: Optional, but could make this code cleaner: Instead of extracting the zone number, we could `file_util::glob` for `/sys/class/thermal/thermal_zone*` and just add `/type` for getting the type and `/temp` to get `m_path`, the zone number isn't really needed for the module itself.

Pick whatever solution you like better.

patrick96 (changes requested):
Please also mention this feature in the [¢hangelog](https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog).

src/modules/temperature.cpp
xphoniex: I kept the zone number as an effort to keep behavior consistent for when there are multiple sensors with the same name. We know we'll always pick the one with lowest number, which depending on hardware and kernel may or may not be random.

Do you still want me to change this?

patrick96: Oh that's a good point. I think sorting the paths should be sufficient to keep consistency (even if it doesn't guarantee that the lowest number is picked).

It's up to you, whichever solution you like better

src/modules/temperature.cpp
xphoniex: https://github.com/polybar/polybar/pull/2752/files#diff-4ecfc5580e380c6cc659596ee5bb12d8f8df8ceea72dae9559e1494503d54691R42

src/modules/temperature.cpp
xphoniex: https://github.com/polybar/polybar/pull/2752/files#diff-4ecfc5580e380c6cc659596ee5bb12d8f8df8ceea72dae9559e1494503d54691R42

src/modules/temperature.cpp
xphoniex: https://github.com/polybar/polybar/pull/2752/files#diff-4ecfc5580e380c6cc659596ee5bb12d8f8df8ceea72dae9559e1494503d54691R28

xphoniex:
> Please also mention this feature in the [¢hangelog](https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog).

https://github.com/polybar/polybar/blob/0ba355760950b0b5932f46c82cac8858eb3b55a9/CHANGELOG.md

patrick96:
Looks very good now. I'll not be at my computer for a few weeks, so I'll have to do the final review + testing once I get back.

xphoniex:
bump to keep the PR in my **Recent activity**

patrick96 (approved):

patrick96:
Thanks!

PR #2784, opened 2022-08-11
Kelvin option for module internal/temperature and more accurate calculation of the temperature in °F

Quantenzitrone:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
implements temperature in Kelvin for the temperature module

## Related Issues & Documents

implements idea #2774

    - just 2 lines for the formatting options of internal/temperature, I can do that, if you want

### DIFF
--- a/src/modules/temperature.cpp
+++ b/src/modules/temperature.cpp
@@ -53,20 +53,24 @@ namespace modules {
   bool temperature_module::update() {
     m_temp = std::strtol(file_util::contents(m_path).c_str(), nullptr, 10) / 1000.0f + 0.5f;
     int temp_f = floor(((1.8 * m_temp) + 32) + 0.5);
+    int temp_k = m_temp + 273;
 
     string temp_c_string = to_string(m_temp);
     string temp_f_string = to_string(temp_f);
+    strung temp_k_string = to_string(temp_k);
 
     // Add units if `units = true` in config
     if(m_units) {
       temp_c_string += "°C";
       temp_f_string += "°F";
+      temp_k_string += "K";
     }
 
     const auto replace_tokens = [&](label_t& label) {
       label->reset_tokens();
       label->replace_token("%temperature-f%", temp_f_string);
       label->replace_token("%temperature-c%", temp_c_string);
+      laber->replace_token("%temperature-k%", temp_k_string);
 
       // DEPRECATED: Will be removed in later release
       label->replace_token("%temperature%", temp_c_string);
--- a/src/modules/temperature.cpp
+++ b/src/modules/temperature.cpp
@@ -51,9 +51,10 @@ namespace modules {
   }
 
   bool temperature_module::update() {
-    m_temp = std::strtol(file_util::contents(m_path).c_str(), nullptr, 10) / 1000.0f + 0.5f;
-    int temp_f = floor(((1.8 * m_temp) + 32) + 0.5);
-    int temp_k = m_temp + 273;
+    int temp = std::strtol(file_util::contents(m_path).c_str(), nullptr, 10);
+    int m_temp = temp / 1000 + 0.5;
+    int temp_f = (temp * 1.8) / 1000 + 32.5;
+    int temp_k = (temp + 273150) / 1000 + 0.5;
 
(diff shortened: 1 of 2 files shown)


src/modules/temperature.cpp
```diff
@@ -53,20 +53,24 @@ namespace modules {
...
 
     string temp_c_string = to_string(m_temp);
     string temp_f_string = to_string(temp_f);
+    string temp_k_string = to_string(temp_k);
 
     // Add units if `units = true` in config
     if(m_units) {
       temp_c_string += "°C";
       temp_f_string += "°F";
+      temp_k_string += "K";
```
patrick96, line 66: Should probably include the degree sign

Quantenzitrone: [kelvin has no degree sign](https://en.wikipedia.org/wiki/Kelvin)

patrick96: Well TIL. Maybe add a comment in the code (doesn't require a link)

src/modules/temperature.cpp
```diff
@@ -53,20 +53,24 @@ namespace modules {
   bool temperature_module::update() {
     m_temp = std::strtol(file_util::contents(m_path).c_str(), nullptr, 10) / 1000.0f + 0.5f;
     int temp_f = floor(((1.8 * m_temp) + 32) + 0.5);
+    int temp_k = m_temp + 273;
```
patrick96, line 56: Should be 273.15

src/modules/temperature.cpp
Quantenzitrone: i dont think it is reqired since it is cleary defined like that

src/modules/temperature.cpp
Quantenzitrone: no, m_temp and temp_k are integers.
the result would be exactly the same

patrick96: My bad, I assumed it's float because we do some floating point operations on the original temperature readings.

I'm reviewing on my phone so I don't have all the context. Will continue reviewing once I'm at my computer again in a week or so, otherwise I'll make more of these mistakes 😅

But I do think we should try to make an accurate conversion, we do in theory have precise enough temperatures for it to matter

src/modules/temperature.cpp
Quantenzitrone: Well, it doesn't have to be explained for °C to have a degree sign, so why should it be needed to explain for Kelvin not having a degree sign? What do Temperatures have to do with Angles in the first place? Even angles don't need a degree sign if you use Radians.

Quantenzitrone:
I tested my changes on Linux Mint, and it works perfectly.

My main distro is NixOS, but I haven't figured out how to build stuff there yet.

Quantenzitrone:
ok i have no idea what this means lol

patrick96 (changes requested):

patrick96:
> ok i have no idea what this means lol

What what means?

Quantenzitrone:
> > ok i have no idea what this means lol
> 
> What what means?

the codecov thingy

patrick96:
> the codecov thingy

I see. That reports how this PR affects our code coverage. Doesn't really matter here since we don't have any tests for module behavior yet.

src/modules/temperature.cpp
```diff
@@ -51,22 +51,27 @@ namespace modules {
   }
 
   bool temperature_module::update() {
-    m_temp = std::strtol(file_util::contents(m_path).c_str(), nullptr, 10) / 1000.0f + 0.5f;
-    int temp_f = floor(((1.8 * m_temp) + 32) + 0.5);
+    float temp = float(std::strtol(file_util::contents(m_path).c_str(), nullptr, 10)) / 1000.0;
```
patrick96, line 54: Nice. Computations on floats make much more sense here :)

patrick96 (approved):

Quantenzitrone:
thanks :)

patrick96:
Thank your for contributing and also cleaning up some of our code ;)

Issue #2022, opened 2020-02-27
Fall back to unnamed format definitions for missing named ones


BVollmerhaus:
## Is your feature request related to a problem?
All of the modules in my bar look identical – they have the same background, padding, icon (prefix) font and so on. Thus, I have created a generic section in my config file with all those properties that I can just inherit in every module:

```
[common-format]
format-background = ${color.module-bg}
format-padding = ${layout.module-padding}

format-prefix-font = ${layout.icon-font}
format-prefix-foreground = ${color.icon-fg}
format-prefix-padding-right = ${layout.icon-padding}
```

```
[module/memory]
type = internal/memory
inherit = common-format
format-prefix = ﬘
label = %gb_used%
```

This works well for many simple modules, such as `cpu` or `memory`, but fails with more complex modules that use named properties in the form of `format-NAME-...` or `format-NAME-prefix-...`. These named properties don't automatically fall back to the unnamed ones if they are not defined, which necessitates a lot of duplication in the config. Instead of just inheriting from my `common-format` section, I have to redefine all the formats again, just with the `-NAME-` in between.

```
[module/network]
type = internal/network

format-connected-background = ${color.module-bg}
format-connected-padding = ${layout.module-padding}

format-connected-prefix-font = ${layout.icon-font}
format-connected-prefix-foreground = ${color.icon-fg}
format-connected-prefix-padding-right = ${layout.icon-padding}

format-disconnected-background = ${color.module-bg}
format-disconnected-padding = ${layout.module-padding}

format-disconnected-prefix-font = ${layout.icon-font}
format-disconnected-prefix-foreground = ${color.icon-fg}
format-disconnected-prefix-padding-right = ${layout.icon-padding}
```

## Why does Polybar need this feature?
I'd imagine that reusing the same formats across multiple, if not all, modules is a very common use-case. At the moment, this requires extensive duplication, even if all values are identical.

## Describe the solution you'd like
It would probably be the most straightforward for all named formats to automatically fall back to their unnamed counterpart if the named one is not defined.

## Describe alternatives you've considered
Maybe even allow defining those formats in the bar section itself and then use them for all modules that don't define their own. This would still require the named formats to fall back to the unnamed ones but would provide an even easier way for global formats without an additional section to inherit from.

patrick96:
[...]

PR #2810, opened 2022-09-01
Added format-x support for ipc_module, tested with demo config

### LINKED ISSUE #2775: Native styling support for custom IPC modules
### Discussed in https://github.com/polybar/polybar/discussions/2771

<div type='discussions-op-text'>

<sup>Originally posted by **beyond9thousand** July 21, 2022</sup>
Would love to be able to assign different colors to different hooks in a custom IPC module</div>

---

That could probably be done by supporting `format-X` for every `hook-X` and have `format-X` inherit from `format`.

PRs welcome :)

madhavpcm:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [x] Documentation Update

## Description

Users can now assign separate formats for different hooks inside the IPC module.

## Related Issues & Documents

Closes #2775 

Changes to documentation
-----
In wiki > Modules > ipc
Examples,
```
[module/demo]
type = custom/ipc
hook-0 = echo foobar
hook-1 = date +%s
hook-2 = whoami
format= <label>
format-0-foreground= ${color.foreground}
format-1-foreground= ${color.foreground}
format-2-foreground= ${color.foreground}
; Based on colors defined by user
format-0-background= ${color.shade1}
format-1-background= ${color.shade2}
format-2-background= ${color.shade3}
initial = 1
click-left = "#demo.hook.0"
click-right = "#demo.hook.1"
double-click-left = "#demo.hook.2"

```

### DIFF
--- a/src/modules/ipc.cpp
+++ b/src/modules/ipc.cpp
@@ -110,7 +110,7 @@ namespace modules {
   string ipc_module::get_format() const {
     if (m_current_hook != -1 && (size_t)m_current_hook < m_hooks.size()) {
       string format_i = "format-" + to_string(m_current_hook);
-      if (m_conf.has(m_name, format_i)) {
+      if (m_formatter->has_format(format_i)) {
         return format_i;
       } else {
         return DEFAULT_FORMAT;
--- a/src/modules/ipc.cpp
+++ b/src/modules/ipc.cpp
@@ -116,6 +116,7 @@ namespace modules {
         return DEFAULT_FORMAT;
       }
     }
+    return DEFAULT_FORMAT;
   }
   /**
    * Output content retrieved from hook commands
--- a/src/modules/ipc.cpp
+++ b/src/modules/ipc.cpp
@@ -61,11 +61,9 @@ namespace modules {
     }
     m_formatter->add(DEFAULT_FORMAT, TAG_OUTPUT, {TAG_OUTPUT});
 
-
-    for(size_t i=0; i < m_hooks.size(); i++){
-        string format_i= "format-" + to_string(i) ;
-        format_i= "format-" + to_string(i) ;
-        m_formatter->add(format_i, TAG_OUTPUT, {TAG_OUTPUT});
+    for (size_t i = 0; i < m_hooks.size(); i++) {
+      string format_i = "format-" + to_string(i);
+      m_formatter->add_optional(format_i, {TAG_OUTPUT});
     }
   }
 
@@ -110,13 +108,14 @@ namespace modules {
   }
 
   string ipc_module::get_format() const {
-      if( m_current_hook != -1 && (size_t)m_current_hook < m_hooks.size()){
-        return "format-" + to_string(m_current_hook);
+    if (m_current_hook != -1 && (size_t)m_current_hook < m_hooks.size()) {
+      string format_i = "format-" + to_string(m_current_hook);
+      if (m_conf.has(m_name, format_i)) {
+        return format_i;
+      } else {
+        return DEFAULT_FORMAT;
       }
-      else{
-        throw module_error("Initial hook out of bounds: '" + to_string(m_initial + 1) +
-                           "'. Defined hooks go from 1 to " + to_string(m_hooks.size()) + ")");
-      }
+    }
   }
   /**
    * Output content retrieved from hook commands
@@ -165,7 +164,8 @@ namespace modules {
       }
     } catch (const std::invalid_argument& err) {
       m_log.err(
-          "%s: Hook action received '%s' cannot be converted to a valid hook index. Defined hooks goes from 0 to %zu.",
+          "%s: Hook action received '%s' cannot be converted to a valid hook index. Defined hooks goes from 0 to "
+          "%zu.",
           name(), data, m_hooks.size() - 1);
     }
   }
--- a/src/modules/ipc.cpp
+++ b/src/modules/ipc.cpp
@@ -59,8 +59,14 @@ namespace modules {
     for (auto& hook : m_hooks) {
       hook->command = pid_token(hook->command);
     }
-
     m_formatter->add(DEFAULT_FORMAT, TAG_OUTPUT, {TAG_OUTPUT});
+
+
+    for(size_t i=0; i < m_hooks.size(); i++){
+        string format_i= "format-" + to_string(i) ;
+        format_i= "format-" + to_string(i) ;
+        m_formatter->add(format_i, TAG_OUTPUT, {TAG_OUTPUT});
+    }
   }
 
   /**
@@ -103,6 +109,15 @@ namespace modules {
     return m_builder->flush();
   }
 
+  string ipc_module::get_format() const {
+      if( m_current_hook != -1 && (size_t)m_current_hook < m_hooks.size()){
+        return "format-" + to_string(m_current_hook);
+      }
+      else{
+        throw module_error("Initial hook out of bounds: '" + to_string(m_initial + 1) +
+                           "'. Defined hooks go from 1 to " + to_string(m_hooks.size()) + ")");
+      }
+  }
   /**
    * Output content retrieved from hook commands
    */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `custom/text`: The `content` setting and all its properties are deprecated in favor of `format` with the same functionality. ([`#2676`](https://github.com/polybar/polybar/pull/2676))
 
 ### Added
+-  Added support for format-i for each hook-i defined in ipc module ([`#2775`](https://github.com/polybar/polybar/issues/2775),  [`#2810`](https://github.com/polybar/polybar/pull/2810))  by [@madhavpcm](https://github.com/madhavpcm).
 - `internal/temperature`: `%temperature-k%` token displays the temperature in kelvin ([`#2774`](https://github.com/polybar/polybar/discussions/2774), [`#2784`](https://github.com/polybar/polybar/pull/2784))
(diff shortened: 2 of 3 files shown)


src/modules/ipc.cpp
```diff
@@ -103,6 +109,15 @@ namespace modules {
     return m_builder->flush();
   }
 
+  string ipc_module::get_format() const {
+      if( m_current_hook != -1 && (size_t)m_current_hook < m_hooks.size()){
+        return "format-" + to_string(m_current_hook);
+      }
+      else{
+        throw module_error("Initial hook out of bounds: '" + to_string(m_initial + 1) +
```
patrick96, line 117: This crashes the module if `initial` is not set, or set to 0.

We guarantee that `-1 <= m_current_hook < m_hooks.size()`. So, we can simply return `DEFAULT_FORMAT` if `has_hook()` returns false and otherwise return the constructed format name.

src/modules/ipc.cpp
```diff
@@ -59,8 +59,14 @@ namespace modules {
     for (auto& hook : m_hooks) {
       hook->command = pid_token(hook->command);
     }
-
     m_formatter->add(DEFAULT_FORMAT, TAG_OUTPUT, {TAG_OUTPUT});
+
+
+    for(size_t i=0; i < m_hooks.size(); i++){
+        string format_i= "format-" + to_string(i) ;
+        format_i= "format-" + to_string(i) ;
```
patrick96, line 67:
```suggestion
```

patrick96 (changes requested):
Thanks :)

I have unfortunately noticed that this breaks any existing config that does formatting:

```dosini
[module/ipc]
type = custom/ipc

hook-0 = echo foobar

format-background = #f00
format-foreground = #fff
```

The current behavior for this is that the module will display `foobar` with white text and red background.
However, with this PR, those settings are ignored (since we are now using `format-0`) and the module displays `foobar` with the bar background and foreground colors.

This makes things more difficult than I first thought. At the very least, we should fall back to `format` if `format-N` is not defined.

---

Please also add a changelog entry for your changes: https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog

And run `clang-format` on the files you modified: https://polybar.readthedocs.io/en/latest/dev/style-guide.html#code-formatting

madhavpcm:
Working on it asap.

src/modules/ipc.cpp
madhavpcm: I think we need to keep some sort of mapping (like unordered_map) for hook-x to the internal index we store.

patrick96: A simple fallback could also be implemented by adding the `format-N` using `m_formatter->add_optional` and then in `get_format` first checking `m_formatter->has_format` and if it doesn't exist, fall back to the default format.

patrick96:
> Working on it asap.

There is really no need for stressing about completing this. Take your time, we're not in a hurry :)

src/modules/ipc.cpp
madhavpcm: > A simple fallback could also be implemented by adding the `format-N` using `m_formatter->add_optional` and then in `get_format` first checking `m_formatter->has_format` and if it doesn't exist, fall back to the default format.

I tried this, add_optional is not adding the format to the list, in addition , I noticed that this config
```
hook-0 = echo foobar
hook-1 = date +%s
;hook-2 = whoami
hook-3 = pwd
format= <label>
format-0-foreground= ${color.foreground}
format-1-foreground= ${color.foreground}
;format-2-foreground= ${color.foreground}

format-0-background= ${color.shade1}
format-1-background= ${color.shade2}
;format-2-background= ${color.shade3}
initial = 1
click-left = "#demo.hook.0"
click-right = "#demo.hook.1"
;double-click-left = "#demo.hook.2"
```
doesnt detect hook-3, because hook-2 stuff is commented. ( is this a rule in configs where you define hook-x in increasing order from 0 to n).  Im trying to figure out why m_conf.has(m_modname, name) is failing.

patrick96: > I tried this, add_optional is not adding the format to the list,

Yeah, the disadvantage of that is that it only detects a format if the key `format-N` is defined. If you only define `format-N-foreground`, it doesn't see it.

And yes, it's a rule that lists like this have to be complete. So for `hook-3` to be loaded, `hook-0` up to `hook-2` have to exist as well.

madhavpcm: > Yeah, the disadvantage of that is that it only detects a format if the key `format-N` is defined. If you only define `format-N-foreground`, it doesn't see it.

If I add format-N in the config, there is no output from the IPC module. In this case, neither ```add()``` nor ```add_optional()``` works. If I search for format-n in  ipc::constructor with ```has_format ("format-n")```, the result turns out negative.

patrick96: Then there is probably something else going wrong. If the setting `format-0` exists, both `add("format-0")` and `add_optional("format-0")` will load the format and `has_format("format-0")` will return true.

madhavpcm: > Then there is probably something else going wrong. If the setting `format-0` exists, both `add("format-0")` and `add_optional("format-0")` will load the format and `has_format("format-0")` will return true.

polybar|error: Failed to get contents for "module/demo" (err: Format "format-0" has not been added)
, throw undefined_format is being thrown. This is what happens when I define format-0 in config, ```has_format()``` fails, 
image.

patrick96: How are you adding the formats to the formatter?

> My doubt is, why does m_formats only have "format", shouldnt it have the other keywords of module/demo

`m_formats` only contains the formats added with `add` or `add_optional`.

`m_conf` is the thing that contains all config options

madhavpcm: > How are you adding the formats to the formatter?
In this case, I commented any addition of format. But even if I add, with either add or add_optional (after defining format-n in config), the module is not visible. I guess this error was due to the fact that I wasn't adding the format.

patrick96:
I think configuration 2 above is the way to go here. There is no good way to make it work without defininig `format-X` since its difficult to tell whether any other format properties are defined.

As to why the module doesn't show anything for that, I don't know. If you upload the latest changes, I may be able to take a look.

> Works with colors of the main bar, hence what I observed is that we should only call add() if there exists format-x in the config.
For this I have to traverse elements in m_conf->m_sections(m_name) and check if format-x is there, as of now I didnt find any function which does this

The thing you're looking for is `m_conf.has`. But what you are describing is exactly what `add_optional` does: it checks if the format exists and only adds it if it does.

madhavpcm:
> As to why the module doesn't show anything for that, I don't know. If you upload the latest changes, I may be able to take a look.

Latest changes demonstrate that if format-X is defined, the module just doesnt output anything (as if it crashed). Comment format-0, so it will reach hook-0 output, from there if you trigger any other hook it hides itself.

madhavpcm:
> > As to why the module doesn't show anything for that, I don't know. If you upload the latest changes, I may be able to take a look.
> 
> Latest changes demonstrate that if format-X is defined, the module just doesnt output anything (as if it crashed). Comment format-0, so it will reach hook-0 output, from there if you trigger any other hook it hides itself.
However if you change add_optional to add, and then place this in get format instead
 ```cpp  
  if (m_conf.has(m_name, format_i)) {
        return DEFAULT_FORMAT;
      } else {
        return format_i;
      }
```
I notice that, if format is defined, or format-x is defined, and get_format returns the defined format in interest, then the module fails.

**Comment by patrick96** [comment]
Ah, I see what the issue is. `<label>` doesn't exist in the ipc module. Try using `format-X = <output>`

File: src/modules/ipc.cpp [hunk 3]
```diff
@@ -103,6 +107,17 @@ namespace modules {
...
   }
 
+  string ipc_module::get_format() const {
+    if (m_current_hook != -1 && (size_t)m_current_hook < m_hooks.size()) {
+      string format_i = "format-" + to_string(m_current_hook);
+      if (m_conf.has(m_name, format_i)) {
+        return format_i;
+      } else {
+        return DEFAULT_FORMAT;
+      }
```
patrick96, line 117:
```suggestion
      if (m_formatter->has_format(format_i)) {
        return format_i;
      }
```

madhavpcm: Sure, il follow up , thanks again!

madhavpcm:
```
format= <output>
format-background = #ff2
format-0 = <output>
format-1 = <output>
format-2 = <output>

format-0-foreground= ${color.foreground}
format-1-foreground= #f00
format-2-foreground= ${color.foreground}
; Based on colors defined by user
format-0-background= ${color.shade1}
format-2-background= ${color.shade3}
```
in this particular config,I want say the partial config of format defined here and partial config of a format-x, to be used in the output. However  format-1-background is not defined, its falling back to bars default color. Is it possible to mix things up ? I tried look around the render function calls , didnt notice at first glance.

madhavpcm:
Also, since <output> is deprecated in some other modules, I can also work on <label> support for IPC module.

patrick96:
> Is it possible to mix things up ?

No. When any property on a format is not defined, it falls back to how that property is defined in the [`[settings]` section](https://github.com/polybar/polybar/wiki/Configuration#application-settings). I think at some point we'll want to have some better fallback mechanisms for formats (there is already an issue about that: #2022), but that would be a bigger undertaking.

> Also, since output is deprecated in some other modules, I can also work on label support for IPC module.

That would be cool. I think the ipc module is the only one that doesn't have any labels. But wait until we merged this PR.

patrick96 (changes requested):
Alright, this seems to be working now :smiley: 

Plase also add a changelog entry and then we can merge this. https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog

patrick96 (approved):

PR #2841, opened 2022-10-14
<label> support for IPC module

madhavpcm:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [x] Documentation Update

<label> support for IPC module, 
label=%output%
label-fail=%output%

are now supported in the IPC module

### DIFF
--- a/src/modules/ipc.cpp
+++ b/src/modules/ipc.cpp
@@ -60,20 +60,33 @@ namespace modules {
     for (auto& hook : m_hooks) {
       hook->command = pid_token(hook->command);
     }
-    m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL});
 
+    m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL});
+    m_formatter->add_optional("output", {});
+
+    if (m_formatter->has_format("output")) {
+      m_conf.warn_deprecated(name(), "output", "format");
+
+      if (m_formatter->get("output")->value.empty()) {
+        throw module_error(name() + ".output is empty or undefined");
+      }
+
+      m_format = "output";
+    } else {
+      m_format = DEFAULT_FORMAT;
+
+      if (m_formatter->has(TAG_LABEL)) {
+        m_label = load_optional_label(m_conf, name(), TAG_LABEL, "%output%");
+      }
+
+      m_formatter->add_optional(FORMAT_FAIL, {TAG_LABEL_FAIL});
+      if (m_formatter->has(TAG_LABEL_FAIL)) {
+        m_label_fail = load_optional_label(m_conf, name(), TAG_LABEL_FAIL, "%output%");
+      }
+    }
     for (size_t i = 0; i < m_hooks.size(); i++) {
       string format_i = "format-" + to_string(i);
       m_formatter->add_optional(format_i, {TAG_LABEL});
-    }
-
-    if (m_formatter->has(TAG_LABEL)) {
-      m_label = load_optional_label(m_conf, name(), TAG_LABEL, "%output%");
-    }
-
-    m_formatter->add_optional(FORMAT_FAIL, {TAG_LABEL_FAIL});
-    if (m_formatter->has(TAG_LABEL_FAIL)) {
-      m_label_fail = load_optional_label(m_conf, name(), TAG_LABEL_FAIL, "%output%");
     }
   }
 
@@ -123,20 +136,29 @@ namespace modules {
       if (m_formatter->has_format(format_i)) {
         return format_i;
       } else {
-        return DEFAULT_FORMAT;
+        return m_format;
       }
     }
-    return DEFAULT_FORMAT;
+    return m_format;
   }
   /**
    * Output content retrieved from hook commands
    */
   bool ipc_module::build(builder* builder, const string& tag) const {
-    if (tag == TAG_LABEL) {
-      builder->node(m_label);
-      return true;
-    } else if (tag == TAG_LABEL_FAIL) {
-      builder->node(m_label_fail);
+    if (m_label) {
+      if (tag == TAG_LABEL) {
+        builder->node(m_label);
+        return true;
+      } else if (tag == TAG_LABEL_FAIL) {
+        builder->node(m_label_fail);
+      }
+    } else {
+      if (tag == "<output>") {
+        builder->node(m_output);
+        return true;
+      } else {
+        return false;
+      }
     }
     return false;
   }
@@ -197,8 +219,10 @@ namespace modules {
       m_current_hook = -1;
 
       m_output.clear();
+
       if (m_label) {
         m_label->reset_tokens();
+        m_label->replace_token("%output%", m_output);
       }
 
       broadcast();
@@ -254,11 +278,14 @@ namespace modules {
       m_output.clear();
     }
 
+    update_output();
+  }
+
+  void ipc_module::update_output() {
     if (m_label) {
       m_label->reset_tokens();
       m_label->replace_token("%output%", m_output);
     }
-
     broadcast();
   }
 } // namespace modules
--- a/src/modules/ipc.cpp
+++ b/src/modules/ipc.cpp
@@ -61,29 +61,10 @@ namespace modules {
       hook->command = pid_token(hook->command);
     }
 
-    m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL});
-    m_formatter->add_optional("output", {});
+    m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL, TAG_OUTPUT});
 
-    if (m_formatter->has_format("output")) {
-      m_conf.warn_deprecated(name(), "output", "format");
+    m_label = load_optional_label(m_conf, name(), TAG_LABEL, "%output%");
 
-      if (m_formatter->get("output")->value.empty()) {
-        throw module_error(name() + ".output is empty or undefined");
-      }
-
-      m_format = "output";
-    } else {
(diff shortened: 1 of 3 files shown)


src/modules/ipc.cpp
```diff
@@ -184,7 +195,12 @@ namespace modules {
       set_hook(m_initial);
     } else {
       m_current_hook = -1;
+
       m_output.clear();
+      if (m_label) {
+        m_label->reset_tokens();
```
patrick96, line 201: Here you also need replace the token again, otherwise the user will see a literal `%output%` on their bar.

In particular, everytime `m_output` changes, you have to reset and replace the tokens and then call `broadcast()`. It's probably best to do that in a separate function.

src/modules/ipc.cpp
```diff
@@ -59,11 +60,20 @@ namespace modules {
     for (auto& hook : m_hooks) {
       hook->command = pid_token(hook->command);
     }
-    m_formatter->add(DEFAULT_FORMAT, TAG_OUTPUT, {TAG_OUTPUT});
+    m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL});
```
patrick96, line 63: This breaks any module that manually defined a `format`. For example the config

```dosini
[module/ipc]
type = custom/ipc
hook-0 = echo 0
hook-1 = echo 1
hook-2 = echo 2

format = |<output>|

initial = 1
```

This module showed the hook output in between pipe characters, but in this PR, it only displays `||`.

We still need to support eh `<output>` tag (just as in the text module) and emit a deprecation warning if it is used.

madhavpcm:
Output, and label are working for these two configs 
```[module/demo]
type = custom/ipc
label= "%output%"
hook-0 = echo foobar
hook-1 = date +%s
hook-2 = whoami
format= <label>
format-background = #ff2
format-0 = <label>
format-1 = <label>
format-2 = <label>

; Based on colors defined by user
format-0-background= ${color.shade1}
format-2-background= ${color.shade3}
initial = 1
click-left = "#demo.hook.0"
click-right = "#demo.hook.1"
double-click-left = "#demo.hook.2"

[module/demo1]
type = custom/ipc
hook-0 = echo 00
hook-1 = echo 10
format = |<output>|
hook-2 = echo 20

format-0 = |<output>|
format-1 = |<output>|
format-2 = |<output>|
format-0-background= ${color.shade1}
format-1-background= ${color.shade1}
format-2-background= ${color.shade3}
initial = 1
click-left = "#demo1.hook.0"
click-right = "#demo1.hook.1"
double-click-left = "#demo1.hook.2"

```. Format-x seems to work for both. Lemme know if there are any other test cases which I missed before I add details to contributing.md

**Comment by madhavpcm** [comment]
> Output, and label are working for these two configs
> 
> ```
> type = custom/ipc
> label= "%output%"
> hook-0 = echo foobar
> hook-1 = date +%s
> hook-2 = whoami
> format= <label>
> format-background = #ff2
> format-0 = <label>
> format-1 = <label>
> format-2 = <label>
> 
> ; Based on colors defined by user
> format-0-background= ${color.shade1}
> format-2-background= ${color.shade3}
> initial = 1
> click-left = "#demo.hook.0"
> click-right = "#demo.hook.1"
> double-click-left = "#demo.hook.2"
> 
> [module/demo1]
> type = custom/ipc
> hook-0 = echo 00
> hook-1 = echo 10
> format = |<output>|
> hook-2 = echo 20
> 
> format-0 = |<output>|
> format-1 = |<output>|
> format-2 = |<output>|
> format-0-background= ${color.shade1}
> format-1-background= ${color.shade1}
> format-2-background= ${color.shade3}
> initial = 1
> click-left = "#demo1.hook.0"
> click-right = "#demo1.hook.1"
> double-click-left = "#demo1.hook.2"
> 
> ```. Format-x seems to work for both. Lemme know if there are any other test cases which I missed before I add details to contributing.md
> ```

Nvm, I broke default format with this. Will push an update soon

File: src/modules/ipc.cpp [hunk 3]
```diff
@@ -184,7 +217,14 @@ namespace modules {
...
       m_current_hook = -1;
+
       m_output.clear();
+
+      if (m_label) {
+        m_label->reset_tokens();
+        m_label->replace_token("%output%", m_output);
+      }
+
       broadcast();
```
patrick96, line 228: Call `update_output` here. Also in `action_send` you also need to do the same

src/modules/ipc.cpp
```diff
@@ -113,21 +136,31 @@ namespace modules {
...
+    return m_format;
   }
   /**
    * Output content retrieved from hook commands
    */
   bool ipc_module::build(builder* builder, const string& tag) const {
-    if (tag == TAG_OUTPUT) {
-      builder->node(m_output);
-      return true;
+    if (m_label) {
```
patrick96, line 148: This check is not necessary, `builder->node` already checks this.

src/modules/ipc.cpp
```diff
@@ -59,11 +60,33 @@ namespace modules {
     for (auto& hook : m_hooks) {
       hook->command = pid_token(hook->command);
     }
-    m_formatter->add(DEFAULT_FORMAT, TAG_OUTPUT, {TAG_OUTPUT});
 
+    m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL});
+    m_formatter->add_optional("output", {});
```
patrick96, line 65: This adds a new format, the ipc module never had a format named `output`.

The deprecation here is slightly different to the text module. In the text module we deprecated the `output` format, here we need to deprecate the `<output>` tag on `format`.

src/modules/ipc.cpp
```diff
@@ -59,11 +60,33 @@ namespace modules {
...
+
+    if (m_formatter->has_format("output")) {
```
patrick96, line 67: I think what you are looking for is `m_formatter->has("<output>", DEFAULT_FORMAT)`.

src/modules/ipc.cpp
```diff
@@ -59,11 +60,33 @@ namespace modules {
...
+      m_format = DEFAULT_FORMAT;
+
+      if (m_formatter->has(TAG_LABEL)) {
+        m_label = load_optional_label(m_conf, name(), TAG_LABEL, "%output%");
+      }
+
+      m_formatter->add_optional(FORMAT_FAIL, {TAG_LABEL_FAIL});
+      if (m_formatter->has(TAG_LABEL_FAIL)) {
+        m_label_fail = load_optional_label(m_conf, name(), TAG_LABEL_FAIL, "%output%");
+      }
```
patrick96, line 85: What is `format-fail` and `label-fail` for?

patrick96 (changes requested):

src/modules/ipc.cpp
madhavpcm: I understood my mistake now. I took some methods used here to behave completely different. 
`cpp 
    if (m_formatter->has(TAG_OUTPUT)) {
      m_conf.warn_deprecated(name(), "<output>", "<label>");
      m_formatter->add(m_format, TAG_OUTPUT, {TAG_OUTPUT});

      for (size_t i = 0; i < m_hooks.size(); i++) {
        string format_i = "format-" + to_string(i);
        m_formatter->add_optional(format_i, {TAG_OUTPUT});
      }

    } else {
        m_label = load_optional_label(m_conf, name(), TAG_LABEL, "%output%");

        for (size_t i = 0; i < m_hooks.size(); i++) {
          string format_i = "format-" + to_string(i);
          m_formatter->add_optional(format_i, {TAG_LABEL});
        }
      }
```
```cpp
  bool ipc_module::build(builder* builder, const string& tag) const {
    if (tag == TAG_LABEL) {
      builder->node(m_label);
      return true;
    } else if (tag == TAG_OUTPUT) {
      builder->node(m_output);
      return true;
    }
    return false;
```
However with this particular approach.I Should I implement in such a way that all format-x and default format should follow  either label or output and not a combination of both ? Or Should I allow use of both output and label by the user. Seems to me that we should only allow tag_label or tag_output. 
Also another edge case was that If I do format= <output> <label>, Should I set whether the user has to follow <output> or <label> based on the default format ?

patrick96: There is no reason to restrict the user from only using `<output>` or `<label>` but never both, that's just more work to check for that.

In `format-x` we don't need to support `<output>` at all, because it's a new option and thus can't break existing configs.

src/modules/ipc.cpp
```diff
@@ -59,11 +60,28 @@ namespace modules {
...
     }
-    m_formatter->add(DEFAULT_FORMAT, TAG_OUTPUT, {TAG_OUTPUT});
 
-    for (size_t i = 0; i < m_hooks.size(); i++) {
-      string format_i = "format-" + to_string(i);
-      m_formatter->add_optional(format_i, {TAG_OUTPUT});
+    m_formatter->add(DEFAULT_FORMAT, TAG_LABEL, {TAG_LABEL});
+
+    m_format = DEFAULT_FORMAT;
+    if (m_formatter->has(TAG_OUTPUT, DEFAULT_FORMAT)) {
```
patrick96, line 67: You don't need to load separate formats based on whether the user uses `<output>`. You can simply add `<output>` to the whitelist when adding the default format (`{TAG_LABEL, TAG_OUTPUT}`).

That way, you don't need to duplicate the code in the if and else branches

patrick96 (approved):
Looks good now :smiley:

patrick96:
Can you still add a changelog entry? Then we can merge this.

madhavpcm:
Im slightly busy with work now, Ill add this tomorrow.

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

patrick96 (approved):

PR #2844, opened 2022-10-17
option to turn off struts

### LINKED ISSUE #2769: Option to turn off struts
### Discussed in https://github.com/polybar/polybar/discussions/2768

<div type='discussions-op-text'>

<sup>Originally posted by **patrick96** July 16, 2022</sup>
Polybar's struts can cover entire monitors if polybar is placed on a monitor that is in the middle of the X root screen. See also previous issues with tint2: https://gitlab.gnome.org/GNOME/mutter/-/issues/452 

There was also a report for this on [reddit](https://www.reddit.com/r/Polybar/comments/vz9271/polybar_restricts_my_windows_to_only_one_monitor/)

Ideally, WMs would compensate for that and only apply the parts of the struts which do not cover entire screens. However, many don't.

On the polybar side, the only conforming solution seems to be to just turn off struts.

For this, we should give the user an option in the bar section to enable/disable struts (enabled by default).
If struts are disabled, `global/wm.margin-{top,bottom}` being set, should produce a warning, because they are ignored.

</div>

VanillaViking:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

Setting option `enable-struts` to false will prevent polybar from configuring struts.
## Related Issues & Documents

Closes #2769

### DIFF
--- a/src/components/bar.cpp
+++ b/src/components/bar.cpp
@@ -218,6 +218,8 @@ bar::bar(connection& conn, signal_emitter& emitter, const config& config, const 
 
   m_opts.double_click_interval = m_conf.get(bs, "double-click-interval", m_opts.double_click_interval);
 
+  m_opts.struts = m_conf.get(bs, "enable-struts", m_opts.struts);
+
   if (only_initialize_values) {
     return;
   }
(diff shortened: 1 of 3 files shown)


src/components/bar.cpp
```diff
@@ -546,6 +548,16 @@ void bar::reconfigure_pos() {
...
+  if (!m_opts.struts) {
+    if (m_conf.has("global/wm", "margin-bottom")) {
+      m_log.warn("Struts are disabled, ignoring margin-bottom");
+    }
+    if (m_conf.has("global/wm", "margin-top")) {
+      m_log.warn("Struts are disabled, ignoring margin-top");
+    }
+    return;
+  }
+
```
patrick96, line 560: Like this, the warning gets printed every time the function is called (this happens when you toggle the bar). It would be better to move this to where the config options are loaded

patrick96 (approved):
Looks good. Thanks!

PR #2851, opened 2022-10-26
fix code in order to support cmake unity builds

tnixeu:

* [X] Refactor
* [ ] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

I tired building polybar with `CMAKE_UNITY_BUILD":"ON"` and the build failed.

The `#pragma once` in base.inl is needed because it is included multiple times when passed to the compiler.

I am not exactly sure why the changes in encoder.cpp are needed. However, both gcc and clang fail. The variable `header` could be renamed Instead of adding `polybar::` to `ipc::header`. However, this would change more lines.

Both changes should have no impact on the runtime ore a normal build.

### DIFF
--- a/src/ipc/encoder.cpp
+++ b/src/ipc/encoder.cpp
@@ -11,11 +11,11 @@ namespace ipc {
     size_t total_size = HEADER_SIZE + payload.size();
     std::vector<uint8_t> data(total_size);
 
-    auto* header = reinterpret_cast<ipc::header*>(data.data());
-    std::copy(ipc::MAGIC.begin(), ipc::MAGIC.end(), header->s.magic);
-    header->s.version = ipc::VERSION;
-    header->s.size = payload.size();
-    header->s.type = type;
+    auto* msg_header = reinterpret_cast<header*>(data.data());
+    std::copy(MAGIC.begin(), MAGIC.end(), msg_header->s.magic);
+    msg_header->s.version = VERSION;
+    msg_header->s.size = payload.size();
+    msg_header->s.type = type;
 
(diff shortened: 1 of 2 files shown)


src/ipc/encoder.cpp
```diff
@@ -11,9 +11,9 @@ namespace ipc {
     size_t total_size = HEADER_SIZE + payload.size();
     std::vector<uint8_t> data(total_size);
 
-    auto* header = reinterpret_cast<ipc::header*>(data.data());
-    std::copy(ipc::MAGIC.begin(), ipc::MAGIC.end(), header->s.magic);
-    header->s.version = ipc::VERSION;
+    auto* header = reinterpret_cast<polybar::ipc::header*>(data.data());
```
patrick96, line 14: I think renaming the variable `msg_header` and just using `header*` as the type makes more sense. Explicitly having the polybar namespace there will likely not survive any future refactoring.

patrick96:
Dang, I didn't know about unity builds, this makes compiling around twice as fast on my machine :smiley:.

*(It will probably affect the performance of ccache though)*

src/ipc/encoder.cpp
tnixeu: I have renamed it to `msg_header` and updated the commit.

tnixeu:
> Dang, I didn't know about unity builds, this makes compiling around twice as fast on my machine smiley.
> 
> _(It will probably affect the performance of ccache though)_

This surprised me as well, when I first learnt about it. One drawback is, that compiling small changes can take longer because more code is compiled again as well.

patrick96 (approved):
Thanks!

Issue #2551, opened 2021-11-11
Support for regex or globs in interface name (Network module)


mil-ad:
It'd be nice if we could use globs for interface names in the network module. My network link has different names on different machines but ideally I'd like to use the same config file. It'd be nice if I could use something like `interface = en*`

patrick96:
There is already a new `interface-type` setting coming in the next release, where you can select either `wired` or `wireless`. Would that address your use-case?

mil-ad:
so does that pick up the _first_ wired interface? What happens if there multiple links? I normally just have one wired interface but its name is different based on computer or the ethernet dongle I use.

patrick96:
It will choose the first one it finds. If you have multiple wired interfaces, your glob `en*` will probably match all of them as well.

mil-ad:
True. As long as this new `interface-type` does not require explicit interface name I think I'll be fine but globs are probably more flexible? They also match how systemd-network deals with the same issue. An example of `/etc/systemd/network/20-wired.network`:

```
[Match]
Name=en*

[Network]
DHCP=ipv4
```

patrick96:
Yeah, globs or regexes are definitely more flexible. If anyone wants to implement this, PRs are welcome. Make sure to also comment here, so that we know you are working on it.

Either globs or regex are fine. Using `glob` (`man 3 glob`) is more user-friendly but is tied to the filesystem (cannot write tests for it) and I'm not sure how safe it is to pass user-supplied data to it (can it crash the process?). Regex is more flexible and is directly supplied in the standard library.

mil-ad:
Great, I'll have a go at this weekend

patrick96:
Perfect :+1:

mil-ad:
Sorry I completely forgot about this!

madhavpcm:
Is this being implemented ? Im willing to work on this.

patrick96:
There hasn't been any PRs for this feature that I'm aware. @bazuin-32, @madhavpcm You can decide among yourselves who works on this.

rossbrierley:
Hi, I noticed this issue about a week ago, it seems like this issue is still unsolved. I would like to try and fix it!

PR #2853, opened 2022-10-28
Regex support for interfaces in network module

madhavpcm:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [x] Documentation Update

## Description

network module interface now supports regex
interface = wl.* will match any adapter named wlan0 wlan1 etc..

## Related Issues & Documents

#2551 

### DIFF
--- a/src/adapters/net.cpp
+++ b/src/adapters/net.cpp
@@ -122,6 +122,36 @@ namespace net {
     return "";
   }
 
+  std::string find_interface_regex(const string& regexpr) {
+    struct ifaddrs* ifaddrs;
+
+    if (regexpr.empty())
+      return "";
+
+    std::regex interface_regex(regexpr);
+    getifaddrs(&ifaddrs);
+
+    struct ifaddrs* candidate_if = nullptr;
+
+    for (struct ifaddrs* i = ifaddrs; i != nullptr; i = i->ifa_next) {
+      const std::string ifname{i->ifa_name};
+      if (std::regex_match(ifname, interface_regex)) {
+        if (candidate_if == nullptr) {
+          candidate_if = i;
+        } else if (((candidate_if->ifa_flags & IFF_RUNNING) == 0) && ((i->ifa_flags & IFF_RUNNING) > 0)) {
+          candidate_if = i;
+        }
+      }
+    }
+    if (candidate_if) {
+      const std::string name{candidate_if->ifa_name};
+      freeifaddrs(ifaddrs);
+      return name;
+    }
+    freeifaddrs(ifaddrs);
+    return "";
+  }
+
   std::string find_wireless_interface() {
     return find_interface(NetType::WIRELESS);
   }
--- a/src/modules/network.cpp
+++ b/src/modules/network.cpp
@@ -13,7 +13,28 @@ namespace modules {
   network_module::network_module(const bar_settings& bar, string name_)
       : timer_module<network_module>(bar, move(name_)) {
     // Load configuration values
+    // m_interface could be a regex as well
     m_interface = m_conf.get(name(), "interface", m_interface);
+    std::regex interface_regex(m_interface);
+
+    struct ifaddrs* ifaddrs;
+    getifaddrs(&ifaddrs);
+
+    // If it is a regex try matching with all interfaces provided by the system
+    // The first matched interface is chosen
+    bool found = false;
+    for (struct ifaddrs* i = ifaddrs; i != nullptr; i = i->ifa_next) {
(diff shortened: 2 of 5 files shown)


src/modules/network.cpp
```diff
@@ -13,7 +13,24 @@ namespace modules {
   network_module::network_module(const bar_settings& bar, string name_)
       : timer_module<network_module>(bar, move(name_)) {
     // Load configuration values
+    // m_interface could be a regex as well
     m_interface = m_conf.get(name(), "interface", m_interface);
+    std::regex interface_regex(m_interface);
+
+    struct ifaddrs* ifaddrs;
+    getifaddrs(&ifaddrs);
```
patrick96, line 21: This leaks memory. You can also detect that yourself if you configure polybar using the `SANITIZE` build type: `cmake -DCMAKE_BUILD_TYPE=Sanitize ..`

src/modules/network.cpp
```diff
@@ -13,7 +13,24 @@ namespace modules {
...
+    // If it is a regex try matching with all interfaces provided by the system
+    // The first matched interface is chosen
+    for (struct ifaddrs* i = ifaddrs; i != nullptr; i = i->ifa_next) {
+      const std::string ifname{i->ifa_name};
+      // If found set m_interface as the current
+      if (std::regex_match(ifname, interface_regex)) {
+        m_interface = ifname;
+        break;
+      }
+    }
```
patrick96, line 32: We have a similar functionality for finding a wired or wireless interface in `net::find_interface`. There, it prefers a running interface if multiple interfaces are matched. The regex matching should do the same, otherwise it gets confusing.

I think it would also be best to move this functionality into `net.cpp` next to the existing `find_interface`. And if you want and are familiar with `std::function`, you could probably make a generic `find_interface` function that takes a `std::function<bool(const ifaddrs&)>` predicate to do the interface matching in order to not duplicate too much code.

patrick96: We should probably also only run the regex matching if the given interface string is not empty (even though it won't match anything anyway, it makes the intent clearer)

patrick96: And there should be a notice log message stating that an interface was discovered through regex matching (just like when an interface was found using `interface-type`).

patrick96 (changes requested):

src/modules/network.cpp
madhavpcm: > I think it would also be best to move this functionality into `net.cpp` next to the existing `find_interface`. And if you want and are familiar with `std::function`, you could probably make a generic `find_interface` function that takes a `std::function<bool(const ifaddrs&)>` predicate to do the interface matching in order to not duplicate too much code.
Im not familiar with std::function. I have some confusion about combining the argments NetType for normal find_interface and string argument for find_interface_regex.

patrick96: Then don't worry about it and make two separate functions

src/adapters/net.cpp
```diff
@@ -122,6 +122,36 @@ namespace net {
     return "";
   }
 
+  std::string find_interface_regex(const string& regexpr) {
+    struct ifaddrs* ifaddrs;
+
+    if (regexpr.empty())
+      return "";
```
patrick96, line 129: missing braces

src/modules/network.cpp
```diff
@@ -13,7 +13,14 @@ namespace modules {
...
       : timer_module<network_module>(bar, move(name_)) {
     // Load configuration values
+    // m_interface could be a regex as well
     m_interface = m_conf.get(name(), "interface", m_interface);
+    string regexpr = m_interface;
+
+    m_interface = net::find_interface_regex(m_interface);
+
+    if (!m_interface.empty())
+      m_log.notice("%s: Matched regex %s with interface %s", name(), regexpr, m_interface);
```
patrick96, line 23: Missing braces

src/modules/network.cpp
```diff
@@ -13,7 +13,14 @@ namespace modules {
...
 
     if (m_interface.empty()) {
```
patrick96, line 25: If the regex doesn't match any interface, the module will shut down with the message `Missing parameter "module/wireless.interface-type"` because `m_interface` is empty again here.

If we tried to match using regex, we shouldn't try to use `interface-type`, so something like this:

```cpp
if (!m_interface.empty()) {
  // match regex
} else {
  // find interface type
}
```

patrick96 (changes requested):
Please also run clang-format on the files you changed.

patrick96:
Closing due to inactivity

PR #2859, opened 2022-11-05
feat(network): Add option for iec unit (1024 division)

Cationiz3r:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [x] Documentation Update

## Description

Added option `iec` (disabled by default) for the network module, which will divide net volume (eg: speed rate) by 1024 (instead of 1000) when enabled.

This can be added at the end of the **Basic settings** section:
```ini
; Use 1K = 1024, instead of 1K = 1000
; Default: false
iec = true
```

### DIFF
--- a/src/modules/network.cpp
+++ b/src/modules/network.cpp
@@ -57,6 +57,7 @@ namespace modules {
     m_accumulate = m_conf.get(name(), "accumulate-stats", m_accumulate);
     set_interval(1s);
     m_unknown_up = m_conf.get<bool>(name(), "unknown-as-up", false);
+    m_metric_units = m_conf.get<bool>(name(), "metric-units", m_metric_units);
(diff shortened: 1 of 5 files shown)


src/modules/network.cpp
```diff
@@ -57,6 +57,7 @@ namespace modules {
     m_accumulate = m_conf.get(name(), "accumulate-stats", m_accumulate);
     set_interval(1s);
     m_unknown_up = m_conf.get<bool>(name(), "unknown-as-up", false);
+    m_iec = m_conf.get<bool>(name(), "iec", m_iec);
```
patrick96, line 60: Maybe the config option should be called `si-units` or `metric-units` and `true` means `1KB = 1000B`, which is probably clearer than IEC to many people.

PR #2864, opened 2022-11-15
Remove space from CPU module

TH3-S4LM0N:
The `internal/cpu` module was formatted to have default output as: `CPU {cpu percent}`, but since the percentage itself contains a space at the start there are 2 spaces.

I get this is kinda like typo comment prs but it annoys me so.
If this is intentional then why?

### DIFF
--- a/doc/config.ini
+++ b/doc/config.ini
@@ -134,7 +134,7 @@ label = %percentage_used:2%%
 [module/cpu]
 type = internal/cpu
 interval = 2
-format-prefix = "CPU "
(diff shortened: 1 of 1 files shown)


patrick96:
There is only an extra space if the percentage is less than 10 (that's what `%percentage:2%` does):

```
CPU 10%
CPU  5%
```

Your PR would make this:

```
CPU10%
CPU 5%
```

We're doing this so that the total width (in characters) of the module does not change (unless you reach 100%).

PR #2997, opened 2023-07-28
feat(cpu): Add FreeBSD support

Tectu:
## What type of PR is this?

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update
* [ ] Other:

## Description
This adds support for the FreeBSD operating system to the CPU module.

## Related Issues & Documents
N/A

### DIFF
--- a/src/modules/cpu.cpp
+++ b/src/modules/cpu.cpp
@@ -2,8 +2,9 @@
 
 #include <fstream>
 #include <istream>
-#if __FreeBSD__
+#ifdef __FreeBSD__
   #include <sys/types.h>
+  #include <sys/resource.h>
   #include <sys/sysctl.h>
 #endif
 
@@ -133,7 +134,7 @@ namespace modules {
     m_cputimes.clear();
 
     try {
-#if __FreeBSD__
+#ifdef __FreeBSD__
       // Get number of CPUs
       // ToDo: No need to do this on every invocation.
       int ncpu = -1;
@@ -154,9 +155,8 @@ namespace modules {
       //  - interrupt
       //  - idle
       //
-      static constexpr std::size_t cpu_states_num = 5;
-      long cpu_stat[cpu_states_num];
-      sz = sizeof(*cpu_stat) * cpu_states_num;
+      long cpu_stat[CPUSTATES];
+      sz = sizeof(cpu_stat);
       if (sysctlbyname("kern.cp_time", cpu_stat, &sz, nullptr, 0) != 0) {
         m_log.err("Failed to query sysctl 'kern.cp_time' (errno: %s)", strerror(errno));
         return false;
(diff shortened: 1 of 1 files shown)


src/modules/cpu.cpp
```diff
@@ -129,6 +133,48 @@ namespace modules {
...
+      }
+
+      // Query 'kern.cp_time'. The returned values are (in order):
+      //  - user
+      //  - nice
+      //  - system
+      //  - interrupt
+      //  - idle
+      //
+      static constexpr std::size_t cpu_states_num = 5;
```
yuripv, line 157: sys/resource.h: CPUSTATES

and constants below from the same header

src/modules/cpu.cpp
```diff
@@ -129,6 +133,48 @@ namespace modules {
...
+      if (sysctlbyname("kern.cp_time", cpu_stat, &sz, nullptr, 0) != 0) {
+        m_log.err("Failed to query sysctl 'kern.cp_time' (errno: %s)", strerror(errno));
+        return false;
+      }
+
+      // Parse
+      static std::size_t field_offset = sizeof(*cpu_stat) + ncpu;
+      for (std::size_t i = 0; i < ncpu; i++) {
+        m_cputimes.emplace_back(new cpu_time);
+        m_cputimes.back()->user = cpu_stat[0];
```
yuripv, line 169: CP_USER

src/modules/cpu.cpp
```diff
@@ -129,6 +133,48 @@ namespace modules {
...
+        m_cputimes.back()->nice = cpu_stat[1];
```
yuripv, line 170: CP_NICE

src/modules/cpu.cpp
```diff
@@ -129,6 +133,48 @@ namespace modules {
...
+        m_cputimes.back()->system = cpu_stat[2];
```
yuripv, line 171: CP_SYS

src/modules/cpu.cpp
```diff
@@ -129,6 +133,48 @@ namespace modules {
...
+        m_cputimes.back()->steal = cpu_stat[3];   // Note: This is technically the reported "interrupt" time
```
yuripv, line 172: CP_INTR

src/modules/cpu.cpp
```diff
@@ -129,6 +133,48 @@ namespace modules {
...
+        m_cputimes.back()->idle = cpu_stat[4];
```
yuripv, line 173: CP_IDLE

src/modules/cpu.cpp
```diff
@@ -129,6 +133,48 @@ namespace modules {
     m_cputimes.clear();
 
     try {
+#if __FreeBSD__
```
yuripv, line 136: #ifdef would be more safe (and common)

src/modules/cpu.cpp
```diff
@@ -129,6 +133,48 @@ namespace modules {
...
+      // Query 'kern.cp_time'. The returned values are (in order):
+      //  - user
+      //  - nice
+      //  - system
+      //  - interrupt
+      //  - idle
+      //
+      static constexpr std::size_t cpu_states_num = 5;
+      long cpu_stat[cpu_states_num];
+      sz = sizeof(*cpu_stat) * cpu_states_num;
```
yuripv, line 159: Why not simply sizeof(cpu_stat)?

Tectu:
@yuripv Thanks for your review - couldn't agree more!

patrick96:
Thanks for the effort you put in here.

The current policy in polybar is to not add BSD-specific module code into our code base.
We have some BSD-specific code to make it compile without patches, but any actual BSD-specific implementation code should be maintained as patches in their respective packaging repo.

As discussed privately, this may change once we have a robust system for separating modules from their data-gathering code which would allow us to have different implementations for different systems that are accessed over the same API

PR #3072, opened 2024-01-12
Audio unmute on scroll

### LINKED ISSUE #3067: [Bug]: Pulseaudio volume increases on scroll when muted, but doesn't unmute
### Checklist

### Steps to reproduce

1. Run `polybar -c config.ini`
2. Note the current volume (example: 20%)
3. Click on the pulseaudio module to mute
4. Scroll on the pulseaudio module to change the volume
5. Click on the pulseaudio module again to unmute

### Minimal config

```dosini
[bar/top]
width=100%
height=32pt
background="#000000"
foreground="#ffffff"

font-0 = "Monospace"

modules-left = pulseaudio

[module/pulseaudio]
type = internal/pulseaudio
label-muted = muted
label-muted-foreground = #666
```

### Polybar log

```text
polybar|notice:  Parsing config file: config.ini
polybar|warn:  No monitor specified, using "eDP-1-1"
polybar|notice:  Loading module 'pulseaudio' of type 'internal/pulseaudio'
polybar|notice:  pulseaudio: using default sink alsa_output.pci-0000_00_1f.3.analog-stereo
polybar|notice:  Loaded 1 modules
polybar|notice:  Loaded font "Monospace" (name=Noto Sans Mono, offset=0, file=/usr/share/fonts/truetype/noto/NotoSansMono-Regular.ttf)
polybar|notice:  Received signal(2): Interrupt
```

### Expected behavior

I expect the pulseaudio to unmute on scroll if volume is currently muted, similar to what `amixer -D pulse sset Master 5%+ unmute` would do

### Actual behavior

The volume increases but stays unmuted. If this is intended maybe an option such as `unmute-on-scroll` could be added for this

bspwm 0.9.10-51-gaf3bd8b

### Linux Distribution

Ubuntu 23.04 x86_64

### Polybar version

```text
polybar 3.7.1

Build type: Release
Compiler: /usr/bin/c++
Compiler flags:  -O3 -DNDEBUG -Wall -Wextra -Wpedantic -Wdeprecated-copy-dtor -Wsuggest-override
Linker flags:  -Wall -Wextra -Wpedantic -Wdeprecated-copy-dtor -Wsuggest-override  -Wall -Wextra -Wpedantic -Wdeprecated-copy-dtor -Wsuggest-override
```

_No response_

roddyrap:

* [ ] Refactor
* [X] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

Added the option `unmute-on-scroll` to the `alsa` and `pulseaudio` modules. If this option is false (The default) then there is no difference in behaviour from the previous version. If the option is true, when the user uses the scrollwheel on the module, the sound will be unmuted.

## Related Issues & Documents

Closes #3067

It might be worth to mention this as an option in the wiki, so that folks could be aware of it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- An option `unmute-on-scroll` for `internal/pulseaudio` and `internal/alsa` to unmute audio when the user scrolls on the widget.
 
 ## [3.7.1] - 2023-11-27
 ### Build
--- a/include/modules/alsa.hpp
+++ b/include/modules/alsa.hpp
@@ -67,6 +67,7 @@ namespace modules {
     map<control, control_t> m_ctrl;
     int m_headphoneid{0};
     bool m_mapped{false};
+    bool m_unmute_on_scroll{false};
     int m_interval{5};
     atomic<bool> m_muted{false};
     atomic<bool> m_headphones{false};
--- a/src/modules/alsa.cpp
+++ b/src/modules/alsa.cpp
@@ -28,6 +28,7 @@ namespace modules {
     // Load configuration values
     m_mapped = m_conf.get(name(), "mapped", m_mapped);
     m_interval = m_conf.get(name(), "interval", m_interval);
+    m_unmute_on_scroll = m_conf.get(name(), "unmute-on-scroll", m_unmute_on_scroll);
 
     auto master_mixer_name = m_conf.get(name(), "master-mixer", "Master"s);
     auto speaker_mixer_name = m_conf.get(name(), "speaker-mixer", ""s);
@@ -261,6 +262,9 @@ namespace modules {
     }
     const auto& mixers = get_mixers();
     for (auto&& mixer : mixers) {
+      if (m_unmute_on_scroll) {
+        mixer->set_mute(true);
+      }
       m_mapped ? mixer->set_normalized_volume(math_util::cap<float>(mixer->get_normalized_volume() + interval, 0, 100))
                : mixer->set_volume(math_util::cap<float>(mixer->get_volume() + interval, 0, 100));
     }
--- a/src/modules/pulseaudio.cpp
+++ b/src/modules/pulseaudio.cpp
@@ -23,6 +23,7 @@ namespace modules {
 
     // Load configuration values
     m_interval = m_conf.get(name(), "interval", m_interval);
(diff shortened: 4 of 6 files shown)


include/modules/alsa.hpp
```diff
@@ -67,6 +67,7 @@ namespace modules {
     map<control, control_t> m_ctrl;
     int m_headphoneid{0};
     bool m_mapped{false};
+    bool m_unmuteOnScroll{false};
```
patrick96, line 70:
```suggestion
    bool m_unmute_on_scroll{false};
```

We're using snake_case for variable names, please also update in other module

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- An option `unmute-on-scroll` for `internal/pipewire` and `internal/alsa` to unmute audio when the user scrolls on the widget.
```
patrick96, line 13:
```suggestion
- An option `unmute-on-scroll` for `internal/pulseaudio` and `internal/alsa` to unmute audio when the user scrolls on the widget.
```

src/modules/alsa.cpp
```diff
@@ -261,6 +262,7 @@ namespace modules {
     }
     const auto& mixers = get_mixers();
     for (auto&& mixer : mixers) {
+      mixer->set_mute(m_unmuteOnScroll || !mixer->is_muted());
```
patrick96, line 265: Wouldn't this mute instead of unmute?

In any case, we should only call the `set_mute` method if `unmute-on-scroll` is enabled, to avoid the overhead of calling into the alsa library

src/modules/pulseaudio.cpp
```diff
@@ -156,10 +157,12 @@ namespace modules {
   }
 
   void pulseaudio_module::action_inc() {
+    m_pulseaudio->set_mute(!m_unmuteOnScroll && m_muted);
```
patrick96, line 160: Same here. Just call `set_mute(false)` if the user enabled `unmute-on-scroll`

patrick96 (changes requested):
Thanks!

Please also update the descriptions of the `inc` and `dec` [actions](https://polybar.readthedocs.io/en/stable/user/actions.html#internal-alsa) for the two modules to mention the `unmute-on-scroll`.

src/modules/alsa.cpp
roddyrap: I did check this before because it confused me as well, but no. Alsa's set_mute is inverted, meaning passing false makes it muted.
I verified (Before and after the incoming fixes) both modules on my computer, and muting and unmuting work correctly.

patrick96: Well that's incredibly bad design on our part, we will have to fix that at some point

PR #3078, opened 2024-01-17
fix: don't restart if zero active screens are reported

### LINKED ISSUE #2418: Polybar crashes when the system wakes up from sleep
## Checklist

* [x] I have read the appropriate section in the [contributing
  guidelines](https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md)
* [x] I believe this issue is a problem with polybar itself and not a misconfiguration on my part.
* [x] I have searched for other open and closed [issues](https://github.com/polybar/polybar/issues?q=is%3Aissue) that
  may have already reported this problem.
* [x] I have checked the [known issues](https://github.com/polybar/polybar/wiki/Known-Issues) page for this problem.

## Describe the bug

Polybar disappears whenever the system wakes up from sleep. I have to start it manually every time I sleep my laptop.
**Expected behavior:**

Polybar works as it is before system sleeps.

**Actual behavior:**

Polybar exits or crashes.

**Was it working before?**
Yes. It works seamlessly before. The behavior came up after I upgraded my Gentoo system. I think it might be caused by newer version of Xorg server or i3.
## To Reproduce
Sleep the system and then wake it up.

See the attachment file for polybar configs.

## Polybar Log

Below is only the error log. See the attachment file for full log.
```
polybar|notice:  randr_screen_change_notify (2160x1440)... reloading
polybar|warn:  module/i3: Attempting to reconnect socket (reason: Unexpected EOF while reading header)
polybar|notice:  Parsing config file: /home/zypang/.config/polybar/config
polybar|error: No monitors found
```

## Environment:
* WM: i3-4.18 xorg-server-1.20.11
* Distro: Gentoo
* Output of `polybar -vvv`:

```
polybar 3.5.5

Features: +alsa -curl +i3 -mpd +network(libnl) +pulseaudio +xkeyboard

Build type: Gentoo
Compiler: /usr/bin/x86_64-pc-linux-gnu-g++
Compiler flags: -march=native -O2 -pipe -Wall -Wextra -Wpedantic
Linker flags: -Wl,-O1 -Wl,--as-needed
```

## Additional context

[polybar-error.log](https://github.com/polybar/polybar/files/6390165/polybar-error.log)
[config.txt](https://github.com/polybar/polybar/files/6390173/config.txt)

farhaven:

* [ ] Refactor
* [ ] Feature
* [X] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

On laptops and similar devices, RandR sometimes returns zero active screens, for example when closing a laptop's lid before it suspends.

Don't restart in that case because the new polybar instance will see zero screens and quit. Instead, just ignore those kind of events.

## Related Issues & Documents

Fixes #2418. 

### DIFF
--- a/src/components/screen.cpp
+++ b/src/components/screen.cpp
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <csignal>
 #include <thread>
+#include <utility>
 
 #include "components/config.hpp"
 #include "components/logger.hpp"
@@ -101,13 +102,16 @@ void screen::handle(const evt::randr_screen_change_notify& evt) {
 
   m_connection.reset_screen();
   auto screen = m_connection.screen();
-  auto changed = false;
 
-  // We need to reload if the screen size changed as well
+  auto [changed, num_monitors] = have_monitors_changed();
+  if (num_monitors == 0) {
+    m_log.notice("randr_screen_change_notify got 0 connected screens, ignoring event");
+    return;
+  }
+
+  // We need to reload if the screen size changed as well, not just if screens were added/removed.
   if (screen->width_in_pixels != m_size.w || screen->height_in_pixels != m_size.h) {
     changed = true;
-  } else {
-    changed = have_monitors_changed();
   }
 
   if (changed) {
@@ -120,13 +124,14 @@ void screen::handle(const evt::randr_screen_change_notify& evt) {
 /**
  * Checks if the stored monitor list is different from a newly fetched one
  *
- * Fetches the monitor list and compares it with the one stored
+ * Fetches the monitor list and compares it with the one stored. Returns
+ * true if the list of monitors has changed. Also returns the number of monitors.
  */
-bool screen::have_monitors_changed() const {
+std::pair<bool, int> screen::have_monitors_changed() const {
   auto monitors = randr_util::get_monitors(m_connection, true, false);
 
   if (monitors.size() != m_monitors.size()) {
-    return true;
+    return std::make_pair(true, monitors.size());
   }
 
   for (auto m : m_monitors) {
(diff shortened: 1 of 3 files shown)


src/components/screen.cpp
```diff
@@ -107,7 +108,12 @@ void screen::handle(const evt::randr_screen_change_notify& evt) {
   if (screen->width_in_pixels != m_size.w || screen->height_in_pixels != m_size.h) {
     changed = true;
   } else {
-    changed = have_monitors_changed();
+    auto p = have_monitors_changed();
```
patrick96, line 111: We should use structured binding here for clarity. Something like this:

```
    auto [changed, num_monitors] = have_monitors_changed();
```

Not sure how that behaves though since `changed` is already defined, so maybe something like this:

```
    int num_monitors = 0;
    std::tie(changed, num_monitors) = have_monitors_changed();
```

Or this:

```
    auto [changed_, num_monitors] = have_monitors_changed();
    changed = changed_
```

src/components/screen.cpp
```diff
@@ -107,7 +108,12 @@ void screen::handle(const evt::randr_screen_change_notify& evt) {
...
+    changed = p.first;
+    if (p.second == 0) {
```
patrick96, line 113: Might be worth moving this one level up, outside of the `else` block so that we also don't reload if the screen size changed *and* there are no monitors.

patrick96 (changes requested):
Approach looks good.

Please also add an entry to the [changelog](https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md#changelog) under a `Changed` header

src/components/screen.cpp
farhaven: I did not know about destructed binding. TIL :)

farhaven:
Ping :)

Hibbins:
Can we get this in? @patrick96

h1divp:
Please.....

patrick96 (approved):
This one went under, sorry. Looks good now :)

PR #3080, opened 2024-01-18
Fix backlight initial value matching the actual brightness level

unclechu:
The initial value is `0`. If the actual brightness is also `0` you see `%percentage%` token as text, it’s not being replaced. Since `m_brightness` is an `int` the initial value could be just `-1` since the actual brightness is never goes below zero. Thus this bug will never show up again because actual brightness on first render won’t match the “old” `m_brightness` value.

See also for more detailed explanation of the bug: https://github.com/polybar/polybar/discussions/3079#discussioncomment-8169932

* [ ] Refactor
* [ ] Feature
* [x] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

### DIFF
--- a/include/modules/backlight.hpp
+++ b/include/modules/backlight.hpp
@@ -67,7 +67,7 @@ namespace modules {
     brightness_handle m_val;
     brightness_handle m_max;
 
-    int m_percentage = 0;
+    int m_percentage = -1;
 
     chrono::duration<double> m_interval{};
     chrono::steady_clock::time_point m_lastpoll;


patrick96:
I had this happen on my work laptop, props on finding out why this happens.

I think this was introduced in the 3.7 release. I'll port your commit onto the 3.7.2 hotfix branch so that we can get this into the 3.7.2 release.

patrick96:
This is being merged in #3081

PR #3104, opened 2024-03-24
fix(modules/battery): make rate positive if negative

classabbyamp:

* [ ] Refactor
* [ ] Feature
* [x] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

On the Lenovo X13s, the battery firmware returns a negative value for `power_now`, which is interpreted by the battery module as `0UL` because it converts the `string` to `unsigned long`. This should be read as a `long` instead, as the kernel specifies that `power_now` is an `int`: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/linux/power_supply.h?h=v6.0.11#n99

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -65,7 +65,7 @@ namespace modules {
     }
 
     m_rate_reader = make_unique<rate_reader>([this] {
-      unsigned long rate{std::strtoul(file_util::contents(m_frate).c_str(), nullptr, 10)};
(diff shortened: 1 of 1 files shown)


patrick96 (approved):
Change looks good. I'm gonna go ahead and remove the formatting commit because it pollutes the diff (not all files correspond to the formatting style yet and we don't have a stable style defined yet).

classabbyamp:
thanks!

PR #3134, opened 2024-07-16
Add systray to example config.ini

arash28134:

* [ ] Refactor
* [ ] Feature
* [x] Bug Fix
* [ ] Optimization
* [x] Documentation Update

## Description
In the example config (config.ini), systray module is not added to the example bar.
```
modules-left = xworkspaces xwindow
modules-right = filesystem pulseaudio xkeyboard memory cpu wlan eth date
```
```
[module/systray]
type = internal/tray

format-margin = 8pt
tray-spacing = 16pt
```

### DIFF
--- a/doc/config.ini
+++ b/doc/config.ini
@@ -51,7 +51,7 @@ separator-foreground = ${colors.disabled}
 font-0 = monospace;2
 
 modules-left = xworkspaces xwindow
-modules-right = filesystem pulseaudio xkeyboard memory cpu wlan eth date
+modules-right = systray filesystem pulseaudio xkeyboard memory cpu wlan eth date
 
 cursor-click = pointer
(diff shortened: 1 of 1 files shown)


patrick96:
The tray is not added to the module by default on purpose (this was also the case with the old tray, see [this commit](https://github.com/polybar/polybar/commit/97ce1f6e7a13ef887e53c57d0771289e69983c98)).

The main reason is that the tray module has some limitations that other modules don't have. Namely that only one can exist in your graphical session.
If people would use the default config on multiple monitors, the monitor the tray would end up on would be unpredictable and confusing to users that don't know this.

What we could maybe do is just add a comment to the module that it is not included by default and why. I'll open a PR for that.

patrick96:
Closing in favor of #3135

PR #3156, opened 2024-08-22
fix: check if git top-level directory matches CMake project source

### LINKED ISSUE #3152: Polybar picks up wrong version number when release tarball is built under foreign repo
### Steps to reproduce

1. Download a [polybar release](https://github.com/polybar/polybar/releases)
2. Extract it in a nested folder of *another* git repo (that repo has to have at least one tag on the current branch)
3. Build polybar regularly

### Expected behavior

Since polybar is not built from its git repo, it should extract the version information from the `version.txt` file

### Actual behavior

The `git describe ...` command used to determine the version (and whether we are in a git repo) picks up the version tag from the parent repo instead.

The top-level `CMakeLists.txt` should additionally check that the top-level git directory matches the top-level cmake directory. This way, we only use git to determine the version if the source code is in the top-level of the repo.

This still may get "wrong" results if someone is just putting the polybar sources at the top of some other repo. But that case is not distinguishable from polybar forks anyway and is acceptable.

**PRs are welcome**

tinopai:

* [ ] Refactor
* [ ] Feature
* [x] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

Checks if command `git rev-parse --show-toplevel` matches CMake variable `PROJECT_SOURCE_DIR` if it does, then it checks for the git tags for version information as it was doing before.
If directories do not match, it falls back to version.txt

**Considerations**:
- Code assumes that if `git rev-parse ...` does not return 0, `git describe ...` won't either.
- Message shown on mismatch (`CMake and git directory mismatch, falling back to built-in version information.`) could be clearer and also include that `git` may not be installed
- > This still may get "wrong" results if someone is just putting the polybar sources at the top of some other repo. But that case is not distinguishable from polybar forks anyway and is acceptable

## Related Issues & Documents

Fixes #3152 

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,19 +8,34 @@ project(polybar CXX)
 # a version string is used, so the file supports comments
 file(STRINGS version.txt version_txt REGEX "^[0-9]+\\.[0-9]+\\.[0-9]+.*$" LIMIT_COUNT 1)
 
-# If we are in a git repo we can get the version information from git describe
-execute_process(COMMAND git describe --tags --dirty=-dev
+
+execute_process(COMMAND git rev-parse --show-toplevel
   WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
-  RESULT_VARIABLE git_result
-  OUTPUT_VARIABLE git_describe
+  RESULT_VARIABLE git_top_dir_result
+  OUTPUT_VARIABLE git_top_dir
   OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET)
 
-if(git_result EQUAL "0")
-  set(APP_VERSION "${git_describe}")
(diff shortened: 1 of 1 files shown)


patrick96 (changes requested):
Thanks for the fix :)

Code logic looks good. The main thing I'm concerned about is symlinks. When the path to the repo goes through a symlink, this will not work.
`git rev-parse` seems to return the real path with all symlinks resolved while `PROJECT_SOURCE_DIR` seems to not resolve symlinks.

For the robustness, both values should probably be passed through [`get_filename_component`](https://cmake.org/cmake/help/v3.5/command/get_filename_component.html) with `REALPATH`. This also protects against any formatting differences between the two paths (e.g. trailing slashes).

patrick96 (approved):
Looks good now. Thanks!

PR #3157, opened 2024-08-22
 fix(modules/battery): fix battery state not reporting as full on trickle charge

agorgl:

* [ ] Refactor
* [ ] Feature
* [X] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
Many laptops allow for setting a charge threshold when using them mostly plugged in, in order to maintain the battery health in a good state. Currently polybar does not check for this and always sets the battery state to charging, even when the reported charging rate is zero (trickle charge).

This PR aims to fix this, by also checking the battery rate value.

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -305,7 +305,9 @@ namespace modules {
    */
   battery_module::state battery_module::current_state() {
     auto charge = read(*m_capacity_reader);
-    if (charge >= m_fullat) {
+    auto rate = read(*m_rate_reader);
+
+    if (charge >= m_fullat || (charge > 0 && rate == 0)) {
       return battery_module::state::FULL;
     } else if (!read(*m_state_reader)) {
       return charge <= m_lowat ? battery_module::state::LOW : battery_module::state::DISCHARGING;


patrick96:
So once the battery is over this threshold, it stops charging and switches over to this trickle charge?

What does the file at `/sys/class/power_supply/<supply_name>/status` (either in the battery or adapter folder, I don't know) report in that case?

A new "not charging" state may be more accurate in this case.

agorgl:
> So once the battery is over this threshold, it stops charging and switches over to this trickle charge?

Yes, exactly.

> What does the file at /sys/class/power_supply/<supply_name>/status (either in the battery or adapter folder, I don't know) report in that case?

The file at `/sys/class/power_supply/BAT0/status` reports `Charging`, both when normally charging and on trickle charge. There is no `status` file for the adapter (only an `online` one that indicates whether the power adapter is plugged in).

> A new "not charging" state may be more accurate in this case.

I don't think this is the correct view for this. The trickle charging functionality is like limiting the battery to a new capacity, e.g. instead of using the full 100% it just uses the 60% of the available battery capacity. The action performed by the device is similar to the one when the original limit (100%) is used, meaning the power adapter will trickle charge the battery to keep it around 100% charge.

agorgl:
Hello! Kind ping for the above, its been 1.5 years and still no trickle charge support :(

PR #3160, opened 2024-09-01
James/task/add timezone to date module

### LINKED ISSUE #596: Select time zone for date module

jamesikanos:

* [ ] Refactor
* [x] Feature
* [ ] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description
Added a `timezone` config option for the date module.

Follows the Linux timezone convention (ie: Europe/London).

If no timezone specified, just uses the local timezone.

## Related Issues & Documents
Closes #596 

I see that there have been several attempts for this issue.

Documentation change for the wiki mentions that the `timezone` config now exists for date module. Here is a sample config that I'm using:

```
[module/datepst]
type = internal/date
interval = 1

date = %H:%M %Z

timezone = "America/Los_Angeles"

label = %date%
label-foreground = ${colors.disabled}
```

### DIFF
--- a/include/modules/date.hpp
+++ b/include/modules/date.hpp
@@ -31,6 +31,8 @@ namespace modules {
     // @deprecated: Use <label>
     static constexpr auto TAG_DATE = "<date>";
 
+    static std::mutex tz_mutex;
+
     label_t m_label;
 
     string m_dateformat;
--- a/include/modules/date.hpp
+++ b/include/modules/date.hpp
@@ -41,6 +41,8 @@ namespace modules {
     string m_date;
     string m_time;
 
+    string m_timezone;
+
     // Single stringstream to be used to gather the results of std::put_time
     std::stringstream datetime_stream;
 
--- a/src/modules/date.cpp
+++ b/src/modules/date.cpp
@@ -6,6 +6,8 @@
 POLYBAR_NS
 
 namespace modules {
+  std::mutex date_module::tz_mutex;
+
   template class module<date_module>;
 
   date_module::date_module(const bar_settings& bar, string name_, const config& config)
@@ -44,6 +46,9 @@ namespace modules {
 
   bool date_module::update() {
     auto time = std::time(nullptr);
+
+    // Lock the mutex to prevent other threads from modifying TZ
+    std::lock_guard<std::mutex> lock(tz_mutex);
 
     const char* old_tz = getenv("TZ");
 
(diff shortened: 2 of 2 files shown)


patrick96:
First of all, thanks for putting time into this. And sorry that this may not go anywhere in the near future.

Temporarily setting the timezone in polybar is just extremely fragile and may cause unintended and non-deterministic side-effects.

Using `setenv` is fundamentally problematic in polybar because `setenv` isn't thread-safe and polybar's modules run concurrently in different threads.

This causes at least four problems (there may be more that I have not thought of):

1. Multiple date modules concurrently modifying the timezone
  * Your PR deals with this 
2. Other polybar code calling `setenv` concurrently.
  * #1684 tackles this by protecting all changes to the environment with a mutex
3. Programs spawned by polybar (script/ipc module, click commands) may inherit the temporarily modified timezone
  * This can cause really unexpected weirdness that's highly timing dependent
  * This in itself is a deal breaker, but it gets worse...
4. Any code anywhere calling `getenv` (even outside of our control) may be unsafe
  * Any code, including library code we don't control may call `getenv` while we are modifying the timezone for the date module.
  * Modifying the environment invalidates previous results from `getenv`
  * Concurrently calling `getenv` and `setenv` can cause a data race

So even if we solved all concurrency issues inside polybar itself, 3rd party libraries (including the standard library) may call `getenv` in places we are not aware of and trigger undefined behavior. [This post](https://www.reddit.com/r/programming/comments/17zto4i/setenv_is_not_thread_safe_and_c_doesnt_want_to/) on reddit has some good discussion under it.

To conclude, we should never use `setenv` in polybar ever. From the [glibc manual](https://www.gnu.org/software/libc/manual/html_node/Environment-Access.html#Environment-Access):

> Modifications of environment variables are not allowed in multi-threaded programs.

Now how can we actually get a timezone feature into polybar?

In C++20 there is native timezone support without env variables (see [`zoned_time`](https://en.cppreference.com/w/cpp/chrono/zoned_time)). There may be a way to use this in the date module.
Polybar currently still uses the C++17 standard and C++20 is still [experimental in GCC](https://gcc.gnu.org/projects/cxx-status.html#cxx2a). Once C++20 becomes the default mode in a GCC release, it might make sense to use C++20 and tackle this again.

jamesikanos:
I understand. Thank you for replying and giving the feedback. What you've written makes sense.

PR #3222, opened 2025-07-24
battery: fixes to consumption/remaining math

### LINKED ISSUE #3219: [Bug]: battery module handles negative kernel-reported current incorrectly
### Checklist

- [x] I have read the appropriate section in the [contributing guidelines](https://github.com/polybar/polybar/blob/master/CONTRIBUTING.md)
- [x] I believe this issue is a problem with polybar itself and not a misconfiguration on my part
- [x] I have searched for other open and closed [issues](https://github.com/polybar/polybar/issues?q=is%3Aissue) that may have already reported this problem
- [x] I have checked the [known issues](https://github.com/polybar/polybar/wiki/Known-Issues) page for this problem.
- [x] I have followed the [debugging guide](https://github.com/polybar/polybar/wiki/Debugging-your-Config) to narrow down the problem to a minimal config.

### Steps to reproduce

1. Find a system that uses negative current in `/sys/class/power_supply/BATx/current_now` to indicate discharging.
2. Run polybar.
3. Note that the battery indicator is not only a nonsensically large number, it also deletes all widgets to the right.

### Minimal config

```dosini
# Any config with the battery module on a kernel with this behavior should work.
# Here is the "discharging" portion of mine, since displaying "consumption" is critical
# to show the bug.
[module/battery]
type = internal/battery
battery = BAT1

format-discharging = <label-discharging> <ramp-capacity>
format-discharging-prefix = "BAT "
format-discharging-prefix-foreground = ${colors.primary}
label-discharging = -%consumption%W %time%
```

### Polybar log

```text
notice: Parsing config file: /home/cbiffle/.config/polybar/config.ini
notice: Listening for IPC messages (PID: 47568)
notice: Loading module 'xworkspaces' of type 'internal/xworkspaces'
notice: Loading module 'i3' of type 'internal/i3'
notice: Loading module 'cpu' of type 'internal/cpu'
notice: Loading module 'temperature' of type 'internal/temperature'
notice: Loading module 'memory' of type 'internal/memory'
notice: Loading module 'root-filesystem' of type 'internal/fs'
notice: Loading module 'home-filesystem' of type 'internal/fs'
notice: Loading module 'pulseaudio' of type 'internal/pulseaudio'
notice: pulseaudio: using default sink alsa_output.pci-0000_c1_00.6.analog-stereo
notice: Loading module 'wlan' of type 'internal/network'
notice: module/wlan: Discovered wireless interface wlan0
notice: Loading module 'eth' of type 'internal/network'
error: Disabling module "eth" (reason: No interface found for type 'wired')
notice: Loading module 'battery' of type 'internal/battery'
notice: Loading module 'date' of type 'internal/date'
notice: Loaded 11 modules
notice: Loaded font "Source Code Pro:size=6.5" (name=Source Code Pro, offset=2, file=/usr/share/fonts/adobe-source-code-pro/SourceCodePro-Regular.otf)
error: Failed to parse contents (reason: tag parser: No next element. THIS IS A BUG. (Context: '%{l}%{A5:#xworkspaces.prev:}%{A4:#xworkspaces.next:}%{u#cba6f7}%{+u}%{B#585b70} 1 %{B-}%{-u}%{A1:#xworkspaces.focus.1:} 2 %{A}%{A1:#xworkspaces.focus.2:} 3 %{A}%{A1:#xworkspaces.focus.3:} 5 %{A}%{u-}%{A}%{A}%{PR}%{r}%{F#cba6f7}CPU %{F-}%{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-} %{F#a6e3a1}▁%{F-}%{PR} %{F#000}║%{F-} %{F#cba6f7}T %{F-}36°C%{PR} %{F#000}║%{F-} %{F#cba6f7}RAM %{F-}25%%{PR} %{F#000}║%{F-} %{F#cba6f7}/ %{F-}52%%{PR} %{F#000}║%{F-} %{F#cba6f7}~ %{F-}59%%{PR} %{F#000}║%{F-} %{A1:#pulseaudio.toggle:}%{A4:#pulseaudio.dec:}%{A5:#pulseaudio.inc:}%{F#cba6f7}VOL %{F-} 80%%{A}%{A}%{A}%{PR} %{F#000}║%{F-} %{F#cba6f7}WLAN %{F-}Not For Sparkle Ponies2       0 K↑      1 K↓ ▆%{PR} %{F#000}║%{F-} %{F#cba6f7}BAT %{F-}-317320907980800)
```

### Expected behavior

Current would be indicated as discharging, time-remaining would work, and modules to the right of the battery would still display

### Actual behavior

Computation of discharge power hits a series of silent integer overflows (due to use of strtoul to parse a signed number and then some implicit conversions), resulting in a very large nonsense number. This then produces a result too large for a manually-sized string buffer used for conversion to decimal. The conversion routine handles the return value from snprintf incorrectly, and responds to this by enlarging the buffer _after_ the conversion, filling it with `\0`. The module label concatenation code then glibly inserts these `\0` into the assembled output, causing the tag parser to treat the string as ending prematurely (because some parts of the code use string::size and some parts look for an inline `\0`).

i3 version 4.24

### Linux Distribution

Arch

### Polybar version

```text
polybar 3.7.2

Build type: Release
Compiler: /usr/bin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection         -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -Wp,-D_GLIBCXX_ASSERTIONS -g -ffile-prefix-map=/build/polybar/src=/usr/src/debug/polybar -flto=auto -O3 -DNDEBUG -Wall -Wextra -Wpedantic -Wdeprecated-copy-dtor -Wsuggest-override
Linker flags: -Wl,-O1 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now          -Wl,-z,pack-relative-relocs -flto=auto -Wall -Wextra -Wpedantic -Wdeprecated-copy-dtor -Wsuggest-override  -Wall -Wextra -Wpedantic -Wdeprecated-copy-dtor -Wsuggest-override
```

This has some layers to it. Starting at the bottom...

**Negative discharge current?** Yup! I've specifically noticed this on a Framework 16 laptop. While you may not notice this on your machine, the [kernel ABI testing docs for the power subsystem](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-power) are quite clear: "Negative values are used for discharging batteries, positive values for charging batteries and for USB IBUS current."

**`strtoul` for reading current:** Because the current can be negative, this should probably be `strtol`, with some care around the implicit conversions in the arithmetic on the result. The range of a _signed_ `long` should be wide enough to describe discharge current in likely applications (if you're concerned about 32-bit, you could use `int64_t` explicitly). The code may also want to check the error result of `strtoul` since it uses in-band errors, though this is hard to do right -- since `strtoul` does not report an error if the input contains a minus sign, and instead returns the unsigned representation of the negative two's-complement number (for whatever historical reason).

More broadly, it might be worth checking all cases where you're using `strtoul` to parse kernel numeric results. It looks like at some point between the last release and now, _some_ accesses to the `m_frate` file have been updated to handle negative numbers correctly, but this was not done consistently e.g. [here](https://github.com/polybar/polybar/blob/master/src/modules/battery.cpp#L92).

**`snprintf` return value:** If `snprintf` fails, it returns the number of characters it _would have_ written, except for the trailing NUL terminator. This means [the code that reads](https://github.com/polybar/polybar/blob/master/src/modules/battery.cpp#L103)

```c++
      // convert to string with 2 decimmal places
      string rtn(16, '\0'); // 16 should be plenty big. Cant see it needing more than 6/7..
      auto written = std::snprintf(&rtn[0], rtn.size(), "%.2f", consumption);
      rtn.resize(written);
```

...will (1) attempt to render a number into a 16-character buffer, (2) fail if the number is very large due to overflow in strtoul, and then (3) resize the buffer after the fact to be long enough, but fill it with `\0`.

It looks like this specific module is generally trying to avoid the use of `iostream`, which would normally be the most convenient way to format a number into a dynamically sized string, so barring that, to avoid the "this should be plenty big" arbitrary buffer limit _and_ nul resize behavior, you could write something like

```c++
string rtn;
auto written = std::snprintf(&rtn[0], 0, "%.2f", consumption);
string rtn(written + 1, '\0'); // DO NOT FORGET THE +1 FOR THE NUL
std::snprintf(&rtn[0], rtn.size(), "%.2f", consumption);
```

which uses an initial snprintf pass to assess the required length of the string, and then formats it after allocating the buffer. (Alternatively, it'd be about one line with iostream, and I see other modules using iostream...)

This looks like the only place in the codebase where this specific pattern appears.

**Embedded NUL in strings:** The codebase seems to be using std::string pretty consistently, but regularly punts it back to C APIs using `c_str()`. This makes it sensitive to embedded zeros in strings, which is tricky, because those same C APIs (such as `snprintf`) have a habit of _producing_ zeros in strings. This is why the template output is seen as "truncated" by the tag parser, and why the exception message in my posted log is _itself also truncated_ (exception messages are printed with `e.what().c_str()`, masking any content after the first nul).

**Fix:** I have a fix for the battery parts (not the general NUL handling) that appears to work locally, but your contributor guidelines are very clear that you want to see and accept an issue report before receiving a PR. So. Here you go.

cbiffle:

* [ ] Refactor
* [ ] Feature
* [x] Bug Fix
* [ ] Optimization
* [ ] Documentation Update

## Description

_(At the time I posted this, the corresponding bug hadn't been accepted. I hope you'll forgive me for posting a PR anyway, I want to get this written down before I forget the context! I think the diff of this PR may make the bug easier to follow.)_

Measurements of current on Linux may be negative to indicate a discharging battery:

https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-power

Previously, the battery module was parsing them as unsigned longs, resulting in weird integer overflow behavior from negative rates. This commit changes the logic to parse rates as potentially signed and handle that properly.

As I noted in #3219, some changes at some point between the 3.7.2 release and today changed _one_ of the lines that parses the discharge rate to use signed longs. However, this line actually introduces a new opportunity for UB:

- It parses the number using strtol, which uses in-band error signaling and can produce LONG_MIN on error.
- It does not check the error.
- It passes the result to std::abs, which is _undefined_ when applied to LONG_MIN on two's-complement systems. In practice, on normal systems, abs(LONG_MIN) == LONG_MIN
- It then bitwise-casts that possibly-still-negative number to ULONG_MAX.
- This makes the following math unhappy.

This is a product of two's-complement integers having asymmetrical range (there are more negative numbers than positive). To avoid this category of error, I've switched all the relevant code to using floats, and specifically switched parsing to the C++17 stof function that throws on error instead of returning a nonsense value. Since the module is already using exceptions to indicate problems, this seemed like a cleaner solution.

Along the way, I noticed that a manually sized 16-byte buffer was being mishandled with very large discharge rate results, of the sort you get from integer overflows. The specific mishandling would wind up padding the buffer with NUL characters, out to the correct length of the number, but only formatting the first 16 characters of it. These NUL characters wound up getting inserted into the template result and fed to the tag parser, which interprets NUL as a terminator and rejects the string as malformed. Thiscommit changes that to a dynamically sized buffer using std::stringstream, preventing this class of error.

This module is trying to be compatible with battery drivers that present capacity and rate in units of charge/current (Ah, A) _or_ units of energy/power (Wh, W). Previously, it made this decision on a file-by-file basis, so it would accept e.g. a charge_now file for charge status (in Ah) but an energy_full file for the battery's max (in Wh). While unlikely, this would result in nonsense results. This commit alters the logic to choose one or the other and reject such configurations as incompatible. This change is not strictly related to the consumption/remaining math fix, but does make reasoning about it a _lot_ easier. I've added some comments about the units to assist future analysis of the code.

## Related Issues & Documents

Fixes #3219

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -1,5 +1,8 @@
 #include "modules/battery.hpp"
 
+#include <iostream>
+#include <iomanip>
+
 #include "drawtypes/animation.hpp"
 #include "drawtypes/label.hpp"
 #include "drawtypes/progressbar.hpp"
@@ -45,10 +48,25 @@ namespace modules {
     }
 
     // Make capacity reader
+    //
+    // The code from this point is trying to handle cases where the kernel
+    // presents current, or power, but not both. (If it's both, we can just
+    // ignore one.) This mostly doesn't affect the math, but it's important that
+    // we not _mix_ current-based and power-based measurements, or we wind up
+    // being off by a factor of the voltage (which in mV is order-of 10000).
+    //
+    // Somewhat arbitrarily, we'll see which of the _now files we can access,
+    // and then force all other files to be from the same unit family (current
+    // or power) rather than a mix.
     if ((m_fcapnow = file_util::pick({path_battery + "charge_now", path_battery + "energy_now"})).empty()) {
-      throw module_error("No suitable way to get current capacity value");
-    } else if ((m_fcapfull = file_util::pick({path_battery + "charge_full", path_battery + "energy_full"})).empty()) {
-      throw module_error("No suitable way to get max capacity value");
+      throw module_error("No suitable way to get charge/energy now value");
+    }
+
+    bool const units_are_current = string_util::contains(m_fcapnow, "charge_now");
+
+    m_fcapfull = path_battery + (units_are_current ? "charge_full" : "energy_full");
+    if (!file_util::exists(m_fcapfull)) {
+      throw module_error("No suitable way to get max charge/energy value");
     }
 
     m_capacity_reader = make_unique<capacity_reader>([=] {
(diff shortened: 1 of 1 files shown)


cbiffle:
My read of that coverage report is that the low coverage in this patch is not a _regression,_ which is to say, the code was not tested before, and continues to not be tested now. Let me know if I misinterpreted that.

cbiffle:
Hello! Anything else you'd like to see on this? I'd love to be able to switch back away from my fork of Polybar on Framework machines.

